PR #28, opened 2016-03-17
Fixed Swift 2.2 compile error

mpurland:
Compiler error: Generic parameter 'Key' could not be inferred

### DIFF
--- a/GRDB/Record/Record.swift
+++ b/GRDB/Record/Record.swift
@@ -154,7 +154,13 @@ public class Record : RowConvertible, TableMapping, Persistable {
     ///
     /// See `hasPersistentChangedValues` for more information.
     public var persistentChangedValues: [String: DatabaseValue?] {
-        return Dictionary(generatePersistentChangedValues())
+        var persistentChangedValues: [String: DatabaseValue?] = [:]
+        
+        for (key, value) in generatePersistentChangedValues() {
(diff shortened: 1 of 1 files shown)


groue:
Hello @mpurland 

GRDB targets the latest stable release of Xcode, version 7.2.1 at the time, which ships with Swift 2.1: I can't accept your pull request.

But the repository has a [Swift2.2](https://github.com/groue/GRDB.swift/tree/Swift2.2) branch.

Unfortunately, I did not update it lately. You can wait for me to bring it up to date, or do it yourself. What do you prefer?

mpurland:
The PR still works with Swift 2.1.

groue:
Yep, but Swift 2.2 requires more modifications. And eventually you'll find one that can not be made compatible with Swift 2.1. This is why I think that this PR doesn't belong to the master branch.

mpurland:
True, there may be changes necessary in the future. I can however make the PR to the Swift 2.2 branch if you like, but as of right now this is the only change that was needed to allow GRDB compile on Swift 2.2+.

groue:
All right. OK Matthew, I'll accept a PR on the Swift 2.2 branch. I'm downloading the beta 5 right now (yes I was late...).

mpurland:
Not to be difficult. I'll move the PR over the Swift 2.2 branch.

PR #62, opened 2016-06-01
Add GRDBCustomSQLite to permit easily building GRDB with a custom, embedded SQLite library

swiftlyfalling:
This PR lays the groundwork for building GRDB with a custom embedded version of SQLite. (Simply use the GRDBCustomSQLite variant instead of GRDB.)

This enables framework consumers to build with optional SQLite features and change other compile-time SQLite settings, as well as include the latest version of SQLite (instead of relying on the system version).

This is similar in spirit to the SQLCipher integration, but tweaks were made to no longer use SQLite compilation option flags to determine which library to import. (i.e. Previous uses of SQLITE_HAS_CODEC that were guarding "import GRDBCipher" have been changed to a neutral USING_SQLCIPHER or USING_BUILTIN_SQLITE as appropriate - the latter defined by both GRDBCipher and GRDBCustomSQLite.) This enables distinguishing the target framework from the actual options used to compile SQLite, and enables framework consumers to compile a custom SQLite with a different encryption extension (such as the official SQLite Encryption Extension.) The SQLITE_HAS_CODEC define is now only used to guard use of the SQLite encryption APIs.

GRDBCustomSQLite has OSX & iOS targets, and the standard test suite for both (matching GRDBCipher).

Documentation is available in the README.md in the /SQLiteCustom/ folder.


swiftlyfalling:
We need this to be able to take advantage of the new "[Pre-Update Hook](https://www.sqlite.org/c3ref/preupdate_count.html)" available in SQLite 3.13.0:

(Which will dramatically simplify our setup for syncing a local database with GRDB and the cloud - it provides the before & after! 😁)

groue:
That's gorgeous.

groue:
@swiftlyfalling You did an amazing job. Thanks *a lot*.

groue:
This PR has shipped in [v0.71.0](https://github.com/groue/GRDB.swift/blob/master/CHANGELOG.md#0710)

PR #64, opened 2016-06-02
Slow compilation fixes

swiftlyfalling:
Implemented a workaround for the cause of the two slowest-to-compile methods.


groue:
Hooray! Thanks @swiftlyfalling !

PR #66, opened 2016-06-02
Test coverage improvements

swiftlyfalling:
Additional test cases, improving code coverage. (See individual commits for details.)


Tests/Public/Core/DatabaseValue/DatabaseValueTests.swift
```diff
@@ -109,5 +109,36 @@ class DatabaseValueTests: GRDBTestCase {
...
+        
+        let fooData1 = "foo".dataUsingEncoding(NSUTF8StringEncoding)!.databaseValue
+        let fooData2 = "foo".dataUsingEncoding(NSUTF8StringEncoding)!.databaseValue
+        XCTAssertEqual(fooData1, fooData2)
+        XCTAssertEqual(fooData1.hashValue, fooData2.hashValue)
+    }
+    
+    func testNullDatabaseValueHash() {
+        // Null DatabaseValue has hash 0
+        XCTAssertEqual(DatabaseValue.Null.hashValue, 0)
```
groue: Please remove this test. Null having zero hash value is not part of the API contract. The contract is that DatabaseValue adopts Hashable, and this test does not test that.

Tests/Public/Core/DatabaseValueConvertible/NSDataTests.swift
```diff
@@ -14,4 +14,31 @@ class NSDataTests: GRDBTestCase {
         let databaseValue = NSData().databaseValue
         XCTAssertEqual(databaseValue, DatabaseValue.Null)
     }
+    
+    func testNSDataDatabaseValueRoundTrip() {
+        
+        func roundTrip(value: NSData) -> Bool
```
groue: I like that. I should use more helper local functions like that.

Tests/Public/Core/DatabaseQueue/DatabaseQueueTests.swift
```diff
@@ -28,4 +28,154 @@ class DatabaseQueueTests: GRDBTestCase {
             }
         }
     }
+    
+    func testReleaseMemory() {
+        // TODO: test DatabaseQueue.releaseMemory() ?
```
groue: What I care about is that sqlite3_db_release_memory has been called when releaseMemory returns.

We could take inspiration from https://github.com/groue/GRDB.swift/blob/v0.70.1/Tests/Private/DatabasePoolReleaseMemoryTests.swift#L16 and add to Configuration a closure to be called in Database.releaseMemory(). Those tests would fo in the Private directory, because this closure would not be exposed in the public API.

Until then, I think this testReleaseMemory() method should be removed. The TODO is already written at https://github.com/groue/GRDB.swift/blob/v0.70.1/GRDB/Core/DatabasePool.swift#L135

Tests/Public/Core/DatabaseQueue/DatabaseQueueTests.swift
```diff
@@ -28,4 +28,154 @@ class DatabaseQueueTests: GRDBTestCase {
...
+                XCTAssertTrue(compareOptional(inputConfiguration.trace, with: queueConfiguration.trace))
+                #if SQLITE_HAS_CODEC
+                    XCTAssertEqual(inputConfiguration.passphrase, queueConfiguration.passphrase)
+                #endif
+                XCTAssertTrue(compare(fileAttributes: inputConfiguration.fileAttributes, with: queueConfiguration.fileAttributes))
+                XCTAssertEqual(inputConfiguration.defaultTransactionKind, queueConfiguration.defaultTransactionKind)
+                XCTAssertTrue(compare(busyMode: inputConfiguration.busyMode, with: queueConfiguration.busyMode))
+                XCTAssertEqual(inputConfiguration.maximumReaderCount, queueConfiguration.maximumReaderCount)
+                
+                //threadingMode
```
groue: threadingMode is not public. And we only support the [multi-thread mode](https://www.sqlite.org/threadsafe.html). This line can be removed.

Tests/Public/Core/DatabaseQueue/DatabaseQueueTests.swift
```diff
@@ -28,4 +28,154 @@ class DatabaseQueueTests: GRDBTestCase {
...
+                //SQLiteConnectionDidOpen
```
groue: SQLiteConnectionDidOpen, SQLiteConnectionWillClose, and SQLiteConnectionDidClose are not public.

Tests/Public/Core/DatabaseQueue/DatabaseQueueTests.swift
```diff
@@ -28,4 +28,154 @@ class DatabaseQueueTests: GRDBTestCase {
...
+                //SQLiteConnectionWillClose
+                //SQLiteConnectionDidClose
+                //SQLiteOpenFlags
```
groue: Not public.

Tests/Public/Core/DatabaseQueue/DatabaseQueueTests.swift
```diff
@@ -28,4 +28,154 @@ class DatabaseQueueTests: GRDBTestCase {
...
+            do {
+                let inputConfiguration = dbConfiguration
+                let dbQueue = try makeDatabaseQueue()
+                var queueConfiguration = dbQueue.configuration
+                
+                // Test values of the Configurations for equality
+                // NOTE: some of the values can't be tested for equality (ex. closures)
+                
+                XCTAssertEqual(inputConfiguration.foreignKeysEnabled, queueConfiguration.foreignKeysEnabled)
+                XCTAssertEqual(inputConfiguration.readonly, queueConfiguration.readonly)
```
groue: Please remove this line. It is fragile.

What needs to be tested are the side-effects of providing a readonly configuration (inability to write), but not the conservation of this flag.

One could imagine that readonly would be set to true if the database file is locked. We already change input configuration sometimes: https://github.com/groue/GRDB.swift/blob/v0.70.1/GRDB/Core/DatabasePool.swift#L64

Tests/Public/Core/DatabaseQueue/DatabaseQueueTests.swift
```diff
@@ -28,4 +28,154 @@ class DatabaseQueueTests: GRDBTestCase {
...
+                // can only test whether both have trace set
+                XCTAssertTrue(compareOptional(inputConfiguration.trace, with: queueConfiguration.trace))
```
groue: > can only test whether both have trace set

Indeed. This is not a test. Please remove this line

Tests/Public/Core/DatabaseQueue/DatabaseQueueTests.swift
```diff
@@ -28,4 +28,154 @@ class DatabaseQueueTests: GRDBTestCase {
...
+                #if SQLITE_HAS_CODEC
+                    XCTAssertEqual(inputConfiguration.passphrase, queueConfiguration.passphrase)
+                #endif
+                XCTAssertTrue(compare(fileAttributes: inputConfiguration.fileAttributes, with: queueConfiguration.fileAttributes))
+                XCTAssertEqual(inputConfiguration.defaultTransactionKind, queueConfiguration.defaultTransactionKind)
+                XCTAssertTrue(compare(busyMode: inputConfiguration.busyMode, with: queueConfiguration.busyMode))
```
groue: The same. Generally speaking, I don't see much value in these comparisons tests. I much prefer tests.

groue:
@swiftlyfalling Many thanks, again.

I added a few comments. IMHO, the most problematic test is testConfigurationSetGet. I'm open to discussion.

Tests/Public/Core/DatabaseValue/DatabaseValueTests.swift
groue: If you really really want to raise the code coverage (a disputable goal), please move this test to the Tests/Private folder, the tests than can break without requiring a line in the changelog.

swiftlyfalling:
Agree with all your comments. Making fixes.

swiftlyfalling:
@groue Fixes made, per your comments.

groue:
That's perfect. Thanks for your dedication @swiftlyfalling, and for the visible energy you have put there, both in writing sensible tests, and by generally understanding the library inner guts - it doesn't look like I have much to tell you about the GRDB internals.

PR #69, opened 2016-06-03
Added missing call to uninstall the preupdate hook

swiftlyfalling:


groue:
Oops ;-)

swiftlyfalling:
Uninstalling is useful too. ;-)

PR #93, opened 2016-08-06
Table Creation DSL

### LINKED ISSUE #83: SQL Builder for CREATE TABLE
*Continuing the `CREATE TABLE` part of the discussion from issue #73...*

> Let's compare our two present options, last time we attempted at writing actual code:

I abandoned the idea of the single expression builder in favor of the closure-based one you proposed. Like you said, it is nice, but it has quite a few drawbacks. I also noticed the closure approach benefits from better auto-completion in Xcode.

> I was planning to give the primary key columns here, and you answered "but remember that you also need to support the optional conflict clause when you specify a primary key" - OK no primary key columns in this method. Conflict clauses on primary keys are rare, but useful: let's keep them in the scope of our DSL.

Happy to hear agreement on that.

> ```Swift
> struct SQLTableBuilder {
>     func column(_ name: String, _ type: SQLType, ...)
> }
>
>  enum SQLType: String {
>     case text = "TEXT"
>     case integer = "INTEGER"
>     case double = "DOUBLE"
>     case numeric = "NUMERIC"
>     case boolean = "BOOLEAN"
>     case blob = "BLOB"
>     case date = "DATE"
>     case datetime = "DATETIME"
> }
>
> db.create(table: "person") { t in
>     t.column("name", .text)
>     t.column("age", .integer)
> }
> ```

Minor point: if you want to support `VARCHAR(X)` and similar types, this enum could not use raw values but would have to have an associated value and a corresponding helper function to make the string:

```Swift
enum _SQLType {
    case text
    case integer
    case varchar(Int)
    ...
}

extension _SQLType {
    var sql: String {
        switch self {
        case text: return "TEXT"
        ...
        case let varchar(x):
            precondition(x <= 255)
            return "VARCHAR(\(x))"
        }
    }
}
```

Tangential question.. why use the `SQL...` prefix all over GRDB? This is all defined in a package which is known to be in the context of working with SQL. It's unlikely for `Column` to have any other meaning within GRDB so the `SQL` prefix doesn't seem necessary and only serves to make the types more verbose. If you really wanted to separate some parts of the lib you could either create a sub module in the module map to import `GRDB.SQL` (or just import `GRDB` and use the types like `SQL.Column`) or use a caseless enum as a namespace so you can have `SQL.Column` types as well etc... I don't feel too strongly about this, but I do think it helps keep code more readable by not having the types repeat terms.

> Not naming the parameters of column(_:_:) contradicts a few Swift 3 API guidelines, but the very first one: clarity at the point of use.

I think dropping the `name` label but keeping the `type` label would be even more clear:

```Swift
t.column("age", type: .integer)
```

By the way, I would make `SQLColumn` a `StringLiteralConvertible` and change the function definition to accept a column:

```Swift
struct SQLTableBuilder {
    func column(_ name: SQLColumn, type: SQLType, ...)
}
```

Any reason not to do this?

> Your initial idea about unicity was fluent: column(...).unique(onConflict: .rollback). Not bad, but we can't be fluent at some places, and use a parameter elsewhere, when dealing with the two very similar constructs that are NOT NULL and UNIQUE.
> ...
> Should we eventually pick a fluent interface, the nullability should be updated accordingly:

100% agree. The fluent version is more flexible in what it can express grammar-wise. You can easily overload the fluent methods to provide different versions with different defaults. And as you said, it makes sense to keep similar constructs together, `unique` and `notNull` are similar, but one of them cannot be represented by just a parameter, so let's make them all fluent methods.

> Easier said than done, though: just like the CHECK constraints you have already studied, ? placeholders aren't accepted in default clauses :-(. Well, I guess we'll have to perform our own escaping... Do you see another option?

Exactly. Unfortunately I don't see a way around performing our own escaping. I was hoping you would reveal some amazing alternative I had not thought of. One option, if you are worried about safety/injection, is to only allow literal convertible types in these types of expressions (prevent the user from accidentally using external input of some kind during table creation), though I can also see this being severely limiting (think of the SQL Browser app example).

Either way, the expression builder will have to be updated (or a new, separate one made) in order to support native expressions in `CHECK` and `DEFAULT`.

> The actual SQLite grammar accepts CURRENT_TIME, CURRENT_DATE and CURRENT_TIMESTAMP as default values. OK, seriously, do we really have to support them, when inserting a date created in Swift has the same effect? Not supporting them lets us use the simple DatabaseValueConvertible protocol, and throw any value in. Should we support the special literals, we'd have to write t.column("date", .datetime, default: .value("Anonymous")) just for t.column("date", .datetime, default: .current_time) :-/

One other problem with `CURRENT_TIME`, `CURRENT_DATE` and `CURRENT_TIMESTAMP`: If you use `DatabaseValueConvertible` it's possible you implemented your own format for storing dates (which you of course documented), how would those literals work in such cases? I think using `DatabaseValueConvertible` is the best way to go as you suggest, but that does not preclude us from *also* supporting literal SQL strings too, which could prove very useful.

```Swift
db.create(table: "person") { t in
    t.column("datetime", type: .datetime)
        .defaults(sql: "CURRENT_TIME") // or whatever SQL string you want
}

db.create(table: "person") { t in
    t.column("datetime", type: .datetime)
        .defaults(Date()) // the preferred way, any _SQLExpressible should work here
}

db.create(table: "person") { t in
    t.column("status", type: .integer)
        .defaults(42) // Swift integer literal, I guess this also counts as _SQLExpressible?
}
```

> raw SQL snippets?

I think it's valuable to allow for raw SQL snippets in certain parts like `check()`, `defaults()` and possibly others. The builder pattern and fluent methods make this easy to do without sacrificing anything. 
 
> decide on an eventual default id INTEGER PRIMARY KEY

I'm beginning to see some benefit to this, but still not convinced it's a good idea. By making this the default you then have to specify a way to say "this column, which is an integer, is NOT a primary key". 

If it's too magical (lets say you only do this on a column if it's int and the first column) then the declaration will look identical to other integer columns that are integer but not primary key. In the end, marking a column as primary key is simple and doesn't require a lot of extra typing, and it's explicit:

```Swift
try db.create(table: "foo") { t in
	t.column("id", type: .integer) // primary key?
	t.column("bar_id", type: .integer) // not primary key
	t.column("other_id", type: .integer) // not primary key
}
```

This also gets even more complicated if you want to have an integer column be the first column, but actually want a compound primary key. I think being explicit with primary key is still the most straightforward choice. These are all valid options for creating a table with a primary key:

```Swift
try db.create(table: "foo") { t in
	t.column("id", type: .integer)
	    .primaryKey()
}

try db.create(table: "foo") { t in
	t.column("id", type: .integer)
	    .primaryKey(true) // same as previous example, totally pointless but since the previous one used the default param value this is also technically possible. 
}

try db.create(table: "foo") { t in
	t.column("id", type: .integer)
	    .primaryKey(onConflict: .rollback)
}

try db.create(table: "foo") { t in
    t.column("id", type: .integer)
        .primaryKey(order: .desc, onConflict: .rollback)
}

try db.create(table: "foo") { t in
    t.column("id", type: .integer)
        .primaryKey(autoincrement: true, onConflict: .rollback)
}
```

I've [pushed some code to my fork](https://github.com/cfilipov/GRDB.swift/commit/53f069733bec53051e16e2bfe7da37c15b39e07f). This is very preliminary, I'm going to need guidance on your preferred code organization eventually (everything is in one source file and one test file right now) and other preferences you might have. I'm happy to see that your code snippets above actually look pretty close to what I have written so far.

> OK, a little pause. On the remaining list

> Columns:

* [x] `name SQLType PRIMARY KEY`
* [x] `name SQLType PRIMARY KEY ASC/DESC`
* [x] `name SQLType PRIMARY KEY AUTOINCREMENT`
* [x] `name SQLType CHECK ...` *
* [ ] `name SQLType COLLATE ...`
* [x] `name SQLType REFERENCES ... ON ...`

\* I have `CHECK` pretty much implemented but of course more work needs to be done to properly generate expression strings in this context. The SQL literal form of `CHECK` is working, however.

I haven't gotten to `COLLATE` yet but doesn't look very difficult though I haven't looked into how to also support your custom collations (I don't expect any problems).

> Table constraints:

* [x] `PRIMARY KEY (...)`
* [x] `PRIMARY KEY (...) ON CONFLICT ...`
* [x] `UNIQUE KEY (...)`
* [x] `UNIQUE KEY (...) ON CONFLICT ...`
* [ ] `CHECK`
* [x] `FOREIGN KEY ...`

What about the `SELECT` clause for table creation? I don't know how often it's used, but it looks easy to implement so I figure why not.

I haven't implemented `CHECK` as a table constraint yet but that's trivial (disregarding the `?` expression problem for a moment).

> Other:

* [ ] `DROP TABLE`
* [ ] `CREATE/DROP INDEX`
* [ ] `ALTER TABLE ADD COLUMN`
* [ ] `covering indexes`

I haven't even started on any of these. Most of the should actually be much simpler than `CREATE TABLE`. I'll start working on them soon.

groue:
This PR closes #83.

It contains:

- A DSL for creating and altering tables and indexes
- Support for the LENGTH SQLite built-in function (it helps documenting CHECK constraints)
- New SQLCollation enum replaces collation names everywhere in the API
- Promotion of _SQLExpressible to the fully public SQLExpressible protocol
- Extensive testing and documentation
- Unrelated: renaming of PrimaryKey into PrimaryKeyInfo, for consistency with ColumnInfo and IndexInfo.

Many thanks, @cfilipov, for your intense and great contributions. This would have simply not existed without you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,30 @@ Release Notes
 - Upgrade sqlcipher to v3.4.0 ([announcement](https://discuss.zetetic.net/t/sqlcipher-3-4-0-release/1273), [changelog](https://github.com/sqlcipher/sqlcipher/blob/master/CHANGELOG.md))
 
 - Row adopts DictionaryLiteralConvertible
-    
+
+- Database schema DSL (TODO: doc link):
+
+    ```swift
+    // CREATE TABLE pointOfInterests (
+    //   id INTEGER PRIMARY KEY,
+    //   title TEXT,
+    //   favorite BOOLEAN NOT NULL,
+    //   latitude DOUBLE NOT NULL,
+    //   longitude DOUBLE NOT NULL
+    // )
+    db.create(table: "pointOfInterests") { t in
+        t.column("id", .Integer).primaryKey()
+        t.column("title", .Text)
+        t.column("favorite", .Boolean).notNull()
+        t.column("longitude", .Double).notNull()
+        t.column("latitude", .Double).notNull()
+    }
+    ```
+
+**Breaking Changes**
+
+- Built-in SQLite collations used to be named by string: "NOCASE", etc. Now use the SQLCollation enum: `.Nocase`, etc.
+
 
 ## 0.77.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,13 +12,6 @@ Release Notes
 - Database schema DSL (TODO: doc link):
 
     ```swift
-    // CREATE TABLE pointOfInterests (
-    //   id INTEGER PRIMARY KEY,
-    //   title TEXT,
-    //   favorite BOOLEAN NOT NULL,
-    //   latitude DOUBLE NOT NULL,
-    //   longitude DOUBLE NOT NULL
-    // )
     db.create(table: "pointOfInterests") { t in
         t.column("id", .Integer).primaryKey()
         t.column("title", .Text)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Release Notes
     let row: Row = ["name": "foo", "date": NSDate()]
     ```
 
-- Create, and alter tables ([documentation](https://github.com/groue/GRDB.swift#database-schema)):
+- DSL for table creation and updates (closes [#83](https://github.com/groue/GRDB.swift/issues/83), [documentation](https://github.com/groue/GRDB.swift#database-schema)):
 
     ```swift
     db.create(table: "pointOfInterests") { t in
@@ -38,7 +38,12 @@ Release Notes
 
 - Built-in SQLite collations used to be named by string: "NOCASE", etc. Now use the SQLCollation enum: `.Nocase`, etc.
 
-- PrimaryKey has been renamed PrimaryKeyInfo.
+- PrimaryKey has been renamed PrimaryKeyInfo:
+
+    ```swift
+    let pk = db.primaryKey("persons")
+    pk.columns  // ["id"]
+    ```
 
 
 ## 0.77.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Release Notes
     }
     ```
 
+- Support for the `length` SQLite built-in function.
+
 **Breaking Changes**
 
 - Built-in SQLite collations used to be named by string: "NOCASE", etc. Now use the SQLCollation enum: `.Nocase`, etc.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@ Release Notes
 
 - Upgrade sqlcipher to v3.4.0 ([announcement](https://discuss.zetetic.net/t/sqlcipher-3-4-0-release/1273), [changelog](https://github.com/sqlcipher/sqlcipher/blob/master/CHANGELOG.md))
 
-- Row adopts DictionaryLiteralConvertible
+- Row adopts DictionaryLiteralConvertible:
 
-- Database schema DSL (TODO: doc link):
+    ```swift
+    let row: Row = ["name": "foo", "date": NSDate()]
+    ```
+
+- Create, and alter tables ([documentation](https://github.com/groue/GRDB.swift#database-schema)):
 
     ```swift
     db.create(table: "pointOfInterests") { t in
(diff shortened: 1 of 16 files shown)


groue:
@cfilipov Support for table creation DSL has landed in v0.78.0, now merged into the Swift3 branch ([documentation](https://github.com/groue/GRDB.swift/tree/Swift3#database-schema)).

Implementation was rather straightforward, since we two have been discussing the subject [in depth](https://github.com/groue/GRDB.swift/issues/83): https://github.com/groue/GRDB.swift/blob/Swift3/GRDB/QueryInterface/SQLTableBuilder.swift

You'll see that:

- SQLColumn has been renamed Column, as you have suggested :smile:
- The ASC/DESC in `INTEGER PRIMARY KEY ASC` is not supported: I simply did not understand the purpose/usefulness of these modifiers.
- Index on collated columns are not supported: `CREATE INDEX foo ON bar(bad COLLATE NOCASE)`
- Default values for columns are defined with a DatabaseValueConvertible value, not an expression. This is because SQLite, in practice, only accepts literals and expressions that evaluate to a constant.

There is always room for improvement: I hope that you'll provide feedback after you have played with this new API a little.

cfilipov:
@groue This looks wonderful! I've pulled the latest Swift3 branch and started playing with the lib. Switching to GRDB has been great, I'm really happy with the quality of this lib.

PR #103, opened 2016-08-18
Joins DSL

groue:
**This pull request is a RFC about an extension to GRDB that generates join queries.**

```swift
// Load all books along with their author
let books = Book.include(author).fetchAll(db)
```

The proposed API is midway between raw SQL and classic ORM associations. You won't find any "has-many" or "belongs-to" associations, but you won't have to care about table aliases. You won't get any magic, but some sugar, for sure.

The code is in the [Joins branch](https://github.com/groue/GRDB.swift/tree/Joins). I have written a [draft documentation](https://github.com/groue/GRDB.swift/tree/Joins#relations-and-joins) that is much too much low level, and requires SQL knowledge. That's the best I could write so far, as documenting joins is a real challenge. Yet it may be detailed enough to inspire you. And reveal shortcomings, missing stuff, or deep design issues that require a drastic rewrite.

I'd be happy if people who read this RFC would comment below. I'll be off for two weeks, so don't expect an quick answer. Yet I'm all ears, and @cfilipov, @hdlj, @mtissington, @pakko972, @palpatim, @pierlo, @schveiguy, and @swiftlyfalling are especially quite welcome :-)


mtissington:
I come from a background of writing SQL for many years. I dabbled with LINQ from Microsoft and a couple of other platforms too. In the end I see them add layers of complexity to SQL with the idea of then being able to write pseudo SQL in C++, Swift or other language.

With that said, my question that comes to mind is what is the use case for implementing Joins?
Why should I learn to do it this way when I can already do everything in SQL?

GRDB already brings a fabulous layer to SQLite.

It seems to me that in memory databases, for example, bring more value than an additional layer to support joins.

Maybe I'm missing something here?

palpatim:
Hi Gwendal,

I haven't looked over the code in detail, but I did give a quick look to the API.

For me, the benefit of an ORM is transparently handling reads/writes to a database via a straight-up OO interface. It's been years since I used ActiveRecord, and then only lightly, but my recollection is that it had a declarative interface that allowed you to map relationships in the model and then not have to think about them too hard at the call site. When I see examples in your API draft documentation like

```swift
let books = Book.include(author).fetchAll(db)
```

it feels like we're going the opposite direction, requiring a developer to think too hard about table structures at the call site. (Scoping rows especially seems burdensome.)

Recognizing that this is an early draft, is it part of your future direction to move to a model where a developer could largely ignore the underlying database, not just ignore SQL? The thought would be, once we create a model that exposes property-based accessors for relationships, we could simply do something like:

```swift
let book = Book.filter(Book.title == "The Fifth Season").fetchOne(db)
book.author.awards.append(Awards.Hugo)
```

Where each relationship-property would be lazily accessed as needed.

cfilipov:
I'd like to jump in and offer the opposite perspective to the previous comment. I **don't** want an ORM. I used GRDB specifically because it's a _very_ nice Swift layer on top of SQLite. It lets me write my queries and map my models without much unnecessary baggage. I feel like GRDB has a nice balance of convenience and flexibility. It's close enough to SQL where I need it, and adds nice conveniences where I would have had to make them anyway. What's more, I didn't see "ORM" mentioned anywhere in the project description. As far as I know, GRDB does not strive to be an ORM (although some features may be inspired by ORMs to some degree, which is OK when it makes sense), so I don't understand why the above comment is coming at this from that perspective.

Regarding @mtissington's comments:

> With that said, my question that comes to mind is what is the use case for implementing Joins?
Why should I learn to do it this way when I can already do everything in SQL?

Some people don't want to use strings for writing queries. I appreciate that GRDB comes with a thin layer on top of SQL for queries. It's not too much different from SQL so it's easy to learn if you already know SQL, and it lets me write queries without putting them entirely inside of strings. 

Let's take an example form the Readme file:

```Swift
let wines = Wine.fetchAll(db, "SELECT * FROM wines WHERE origin = ? ORDER BY price", arguments: ["Burgundy"])
```
vs
```Swift
let wines = Wine.filter(origin == "Burgundy").order(price).fetchAll(db)
```

While the later isn't as type safe as some of us would like (there's no guarantee that this query is correct with respect to the schema), it's a big step up from the former. The table name is safely handled by the table mapping protocol conformance on `Wine`, the order by or other clauses are written for me so I won't mis-type those.

If you do prefer to write your queries as SQL strings then you're not likely a user of the query interface, so this addition should have no impact since the string-based API is still there. What he's trying to do is make the query interface more complete.

cfilipov:
I am confused about `include`, why is this different from `join`?

> ```SQL
> -- All books along with their author, published in France since 2000
> SELECT books.*, authors.*
> FROM books
> LEFT JOIN authors ON authors.id = books.authorId
> LEFT JOIN publishers ON publishers.id = books.publisherId
> WHERE books.year >= 2000 AND publishers.country = 'FR'
> ```
> ...
> ```Swift
> class Book : Record { ... }
> let author = Relation(to: "authors", fromColumns: ["authorId"])
> let publisher = Relation(to: "publishers", fromColumns: ["publisherId"])
> ```
> The request:
> ```Swift
> let request = Book
>     .include(author)
>     .join(publisher)
>     .filter { source in
>         let bookYear = source["year"]
>         let publisherCountry = source.scoped(on: publisher)["country"]
>         return bookYear >= 2000 && publisherCountry == "FR"
>     }
> ```

In the example above, two relations (`authors` and `publishers`) are defined exactly the same way, they are both joined on `books`, but you have to use `include` for `authors` and `join` for `publishers`? I understand `publishers` is only used for the predicate, so this is likely why, but it's confusing. 

I'm still combing through the rest of the examples, but one thing is becoming clear to me: you are incurring a lot of complexity in the API by attempting to handle column aliasing and objects graphs automatically.

I think the proposed API is too low level and complex for someone who wants the SQL abstracted away (as noted by @palpatim). But it's also too unfamiliar for someone who does know SQL and just wants a light wrapper in Swift. Personally, I would rather do more work dealing with the object graph if that means having a closer mapping to the SELECT grammar.

Of course the current query interface in master doesn't _exactly_ resemble the SELECT grammar. But I feel the areas where you diverged were very reasonable and made it feel more idiomatic or pleasant to write queries in Swift (and still relatively close to SQL in many ways). In contrast, I feel this proposed API requires me to re-learn table joins, and it didn't really save me much in the end.

That said, I know this is a complex feature to tackle. I haven't spent enough time with this yet, so it might just be that this is as simple or as close as you can get without adding more complexity.

groue:
Thanks to all for your comments!

Let's answer, for the record, but be reassured: this PR is rejected, and I won't spend time trying to fix it.

@mtissington:

> I come from a background of writing SQL for many years. I dabbled with LINQ from Microsoft and a couple of other platforms too. In the end I see them add layers of complexity to SQL with the idea of then being able to write pseudo SQL in C++, Swift or other language.
> 
> With that said, my question that comes to mind is what is the use case for implementing Joins?
> Why should I learn to do it this way when I can already do everything in SQL?

That's two good questions. Especially the second one: until now it had prevented me from adding too much layers of bloat. A broad coverage of genuine SQLite features is OK, with a touch of convenience, but adding too many layers of abstraction is not. I aim at providing a good screwdriver that I trust and doesn't come in the way.

I had two related use-cases in mind:

- avoiding the N+1 problem (such as all authors with the number of books they have written, or all books with their authors). Today such a request forces the developer to write SQL, or to write inefficient code. Some developers prefer avoiding SQL, and other don't *know* SQL enough.

- navigation in a row whose columns come from several tables (what I call "row scoping").

> It seems to me that in memory databases, for example, bring more value than an additional layer to support joins.

You can use `DatabaseQueue()`: it returns a queue to an in-memory database ;-)

@palpatim:

> For me, the benefit of an ORM is transparently handling reads/writes to a database via a straight-up OO interface. It's been years since I used ActiveRecord, and then only lightly, but my recollection is that it had a declarative interface that allowed you to map relationships in the model and then not have to think about them too hard at the call site. When I see examples in your API draft documentation like
> 
> ```swift
> let books = Book.include(author).fetchAll(db)
> ```
> 
> it feels like we're going the opposite direction, requiring a developer to think too hard about table structures at the call site. (Scoping rows especially seems burdensome.)

I'm happy that row scoping exists (it [already does](https://github.com/groue/GRDB.swift#row-adapters)), but I take note that it's a feature that belongs to the advanced features of the low-level GRDB, and pollutes the topic of joins and relations between models. There's a level mismatch, here.

> Recognizing that this is an early draft, is it part of your future direction to move to a model where a developer could largely ignore the underlying database, not just ignore SQL? The thought would be, once we create a model that exposes property-based accessors for relationships, we could simply do something like:
> 
> ```swift
> let book = Book.filter(Book.title == "The Fifth Season").fetchOne(db)
> book.author.awards.append(Awards.Hugo)
> ```
> 
> Where each relationship-property would be lazily accessed as needed.

I see where you look at, and I don't aim at that at all. Not that I'm fundamentally against: it's very convenient. I'm a happy user of Active Record too. ORMs really shine on the server. But in applications, they bite back and hard: properties that lazily access the database imply managed objects, and with managed objects come the concurrency difficulties of Core Data and Realm. I happily leave managed objects to those frameworks, and prefer having GRDB explore the consequences of protocols and immutability (see https://medium.com/@gwendal.roue/how-to-build-an-ios-application-with-sqlite-and-grdb-swift-d023a06c29b3).

@cfilipov

> What's more, I didn't see "ORM" mentioned anywhere in the project description. As far as I know, GRDB does not strive to be an ORM (although some features may be inspired by ORMs to some degree, which is OK when it makes sense)

Yes. Since nobody wants to deal with raw rows all day long, it's easy to map query results to objects. But relationships between models are not addressed at all - we thus do not have, properly speaking, an "Object Relational Mapping". This RFC/PR was an attempt at addressing relations, but:

> I'm still combing through the rest of the examples, but one thing is becoming clear to me: you are incurring a lot of complexity in the API by attempting to handle column aliasing and objects graphs automatically.
> 
> I think the proposed API is too low level and complex for someone who wants the SQL abstracted away (as noted by @palpatim). But it's also too unfamiliar for someone who does know SQL and just wants a light wrapper in Swift. Personally, I would rather do more work dealing with the object graph if that means having a closer mapping to the SELECT grammar.
> 
> Of course the current query interface in master doesn't exactly resemble the SELECT grammar. But I feel the areas where you diverged were very reasonable and made it feel more idiomatic or pleasant to write queries in Swift (and still relatively close to SQL in many ways). In contrast, I feel this proposed API requires me to re-learn table joins, and it didn't really save me much in the end.
> 
> That said, I know this is a complex feature to tackle. I haven't spent enough time with this yet, so it might just be that this is as simple or as close as you can get without adding more complexity.

I think those paragraphs sum up quite well all of your comments, @mtissington, @palpatim and @cfilipov. The proposed API just doesn't fit any expectation or need well enough. Your precious feedback leads to the only reasonable conclusion: this attempt is a dead-end :-) Let's move on!

cfilipov:
@groue I disagree about your conclusion that it was a dead-end. I think you had many interesting ideas in there. I still hope to see a join feature in your query interface one day.

groue:
@cfilipov - whatever turns out to be an eventual support for joins, it won't be based on this PR, for all the reasons you and others have quite well explained. I just stumbled upon [Diesel](http://diesel.rs) which shares some design goals with GRDB, and has [interesting ideas about joins](http://docs.diesel.rs/diesel/associations/index.html). Anyway, some time has to pass. And I'll remember that difficulties writing documentation are a good sign of a lacking service :sweat_smile:

PR #114, opened 2016-09-10
Automatic renaming Fix-Its for many of the Swift 2.x -> Swift 3 GRDB API changes

swiftlyfalling:
This enables Xcode to automatically detect and suggest the new names in many cases, easing project migration from GRDB/Swift 2.x to GRDB/Swift3.

Not sure if this is the sort of thing you want to include in GRDB main, but I wrote it to ease the migration of several projects here, and figured it was worth offering up so others may benefit. 😊

NOTE: This covers many, but not all, of the changes.


groue:
This is so useful and delicate, and I wasn't aware of the technique: I'm pretty glad to accept this PR!

> NOTE: This covers many, but not all, of the changes

You were right looking for inspiration in the CHANGELOG: it contains all the breaking changes I could spot. You've already done the biggest part of the job. GRBD users will love that!

PR #130, opened 2016-09-26
[Swift 2.3] Add `@noescape` to appropriate closures, matching the Swift3 branch's behavior

swiftlyfalling:
This PR adds some `@noescape` markers for closures in the Swift 2.3 branch. (Matching the default behavior of the Swift3 branch.)

This should result in:

- additional compiler optimizations / small performance gains
- no need to explicitly capture `self` (weakly or otherwise) in synchronous methods like `.writeInTransaction`
- explicit matching of the Swift3 branch's closure escaping behavior for those methods, easing forward/back migration between the Swift 2.3/Swift 3 branches of GRDB

This also should be a fully backwards-compatible change.

### DIFF
--- a/GRDB/Core/DatabasePool.swift
+++ b/GRDB/Core/DatabasePool.swift
@@ -282,7 +282,7 @@ extension DatabasePool : DatabaseReader {
     ///
     /// - parameter block: A block that accesses the database.
     /// - throws: The error thrown by the block.
-    public func read<T>(block: (db: Database) throws -> T) rethrows -> T {
+    public func read<T>(@noescape block: (db: Database) throws -> T) rethrows -> T {
         // The block isolation comes from the DEFERRED transaction.
         // See DatabasePoolTests.testReadMethodIsolationOfBlock().
         return try readerPool.get { reader in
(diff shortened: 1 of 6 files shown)


groue:
Hello @swiftlyfalling! Thanks for this new contribution!

Xcode 8 doesn't generate any warning related to @noescape closures in the [Swift2.3 branch](https://github.com/groue/GRDB.swift/tree/Swift2.3). Does this pull request target Xcode 8.1 beta? Or anything else?

swiftlyfalling:
@groue: This can be safely used with Xcode 8 and merged right away.

This PR actually has to do with fixing the escaping behavior of the closures to match up between the branches:
-Swift 1/2 defaults to "escaping" closures (unless marked `@noescape`).
-Swift 3 defaults to "non-escaping" closures (unless marked by `@escaping`).

Because of the prior default, the Swift 2.3 (and earlier) branches had some closures that were unnecessarily defaulting to "escaping", when they could have been "non-escaping".

In the Swift 3 branch, these unmarked closures are now "non-escaping" closures. Since they always could have been, this PR explicitly fixes it in the Swift 2.3 branch.

groue:
A big thank you, as always!

groue:
(I'll ship version 0.81.2 shortly)

groue:
There we go, [v0.81.2](https://github.com/groue/GRDB.swift/tree/v0.81.2) is out!

PR #174, opened 2017-02-22
Fix Record's persistenceConflictPolicy

hartbit:
`Record` did not implement `persistenceConflictPolicy` and got the protocol's extension default method. Fix so that subclasses can now implement their own.


groue:
Great, @hartbit! Thank you very much!

Issue #171, opened 2017-02-22
Result codes


hartbit:
One of the pieces which I would love to see in GRDB is better support for result/error codes. One important piece of that would be to start supporting extended error codes, which can be done easily enough (I have a PR ready in case). The second would be improved errors.

Quick and dirty ideas:

```
enum DatabaseError : Error {
    case busy
    // ...
    case constraint(ConstraintError)
}

enum ConstraintError : Error {
    case foreignKey
    case check
    // ...
}

do {
    try db.execute("SOMETHING")
} catch .constraint(.foreignKey) {
    // do something about it
}
```

I this has a few issues: we've lost the error message, etc.. but perhaps it can help us find a good design.

groue:
@hartbit There are two subjects in this issue:

1. Enable extended error codes (through the [sqlite3_extended_result_codes](https://sqlite.org/c3ref/extended_result_codes.html) C function, currently not exposed by GRDB)
2. Exposing SQLite error codes as Swift constants.

Enabling extended error codes is a valid feature request: thanks for the suggestion!

I'm balanced about exposing error codes through a specific GRDB API.

Those codes are already available through [the C SQLite module](https://github.com/groue/GRDB.swift#raw-sqlite-pointers):

```swift
import GRDB
import SQLiteMacOSX

// Without extended error codes
do {
    try db.execute("SOMETHING")
} catch let error as DatabaseError where error.code == SQLITE_CONSTRAINT {
    // do something about it
}

// With extended error codes
do {
    try db.execute("SOMETHING")
} catch let error as DatabaseError where (error.code & SQLITE_CONSTRAINT) != 0 {
    // do something about it
}
```

(However, for some reason I don't understand yet, extended error codes are not usable this way (I get an "unresolved identifier" compilation error). I'd like this oddity to be resolved.)

The reason why I'm not keen defining an error hierarchy similar to the one you provide above (enum DatabaseError, enum ConstraintError, etc.) is that I'd prefer error codes to keep their original names (SQLITE_CONSTRAINT, etc.). For me, it's part of the "don't surprise experts, don't confuse beginners" rule. Plus it fosters googleability, and supports SQLite documentation quotes:

```swift
catch let error as DatabaseError where error.code == SQLITE_CONSTRAINT {
    // https://www.sqlite.org/rescode.html#constraint says:
    // > The SQLITE_CONSTRAINT error code means that an SQL constraint
    // > violation occurred while trying to process an SQL statement. 
    ...
}
```

[...]

PR #180, opened 2017-03-01
Activate extended result codes, and expose result codes as Swift constants

groue:
This PR addresses #171 by @hartbit.

- it activates [extended result codes](https://www.sqlite.org/rescode.html#primary_result_codes_versus_extended_result_codes), so that SQLite reports more precise errors.
- it introduces a new type `ResultCode` which has constants for all SQLite-declared result codes, so that the end user can access those result codes by name without importing the low-level C SQLite module.
- it modifies `DatabaseError` so that its former `code: Int32` property is replaced with two new properties: `resultCode` and `extendedResultCode`, which both let the user handle errors at the desired level of precision.

[Documentation](https://github.com/groue/GRDB.swift/tree/Issue171#databaseerror)

Remaining TODOs:

- [x] add a fixit for migrating away from obsolete `code` property
- [x] check the availability of `sqlite3_extended_result_codes`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Release Notes
     }
     ```
 
+
 **New: Request**
 
 - The Request protocol for [custom requests](https://github.com/groue/GRDB.swift#custom-requests) learned how to count:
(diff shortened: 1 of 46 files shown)


groue:
@hartbit, this has shipped in v0.102.0.

PR #181, opened 2017-03-01
Let all requests count, introduce Request.fetchCount()

groue:
This PR addresses a [suggestion](https://github.com/groue/GRDB.swift/issues/176#issuecomment-282764867) from @hartbit, and adds a `fetchCount` method to the `Request` protocol which fuels [query interface requests](https://github.com/groue/GRDB.swift#requests) as well as [custom requests](https://github.com/groue/GRDB.swift#custom-requests):

```swift
let request: Request = ...
let count = try request.fetchCount(db) // Int
```

The `fetchCount` method has a default implementation which performs a naive counting in the form of a `SELECT COUNT(*) FROM (...)` query. For example:

```swift
let request = SQLRequest("SELECT ...")
// SELECT COUNT(*) FROM (SELECT ...)
let count = try request.fetchCount(db)
```

The `fetchCount` method is a *customization point*, though, that adopting types can implement in order to provide an efficient counting SQL query. This is the case, for example, of the requests from the query interface ([examples](https://github.com/groue/GRDB.swift#fetching-aggregated-values)).

[Documentation](https://github.com/groue/GRDB.swift/tree/RequestFetchCount#custom-requests)

### DIFF
--- a/GRDB/Core/Request.swift
+++ b/GRDB/Core/Request.swift
@@ -13,7 +13,7 @@ public protocol Request {
     /// executed, and an eventual row adapter.
     func prepare(_ db: Database) throws -> (SelectStatement, RowAdapter?)
     
-    /// The number of rows matched by the request.
+    /// The number of rows fetched by the request.
     ///
     /// Default implementation builds a naive SQL query based on the statement
     /// returned by the `prepare` method: `SELECT COUNT(*) FROM (...)`.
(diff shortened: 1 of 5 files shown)


groue:
@hartbit, this has shipped in v0.102.0.

hartbit:
🙏

Issue #72, opened 2016-06-15
SPM support


zmeyc:
(copied from Twitter discussion) It would be nice to have SPM support in GRDB so it can be used with server-side apps.

I'm working on Telegram Bots SDK: https://github.com/zmeyc/telegram-bot-swift and wanted to add an example of using a DB to show how to maintain sessions etc.

>  It’d also help telling which kind of bot you’d like to build on top of your SDK.

A template project to build other bots on. Probably a "shopping list" bot or something similar - for example, if you want to create a list of items to buy and share with other people, just add them to a group chat with the bot and everyone will be able to modify the list.

I've already ported another lib, SQLite.swift to Swift3/SPM, but it lacks some of the functionality I had to add: Records, migrations, upsert etc which GRDB supports out of the box.

> SPM requires a lot of work, and last time I tried (early after the release of OpenSource Swift), I could never link to sqlite.

Currently it links without problems to system sqlite, but it's located in different dirs on OS X (if installed using brew) and Linux and requires using two different packages to workaround this. At first, I opted to compile it from source on OS X to install it into /usr/local instead:
https://github.com/zmeyc/CSQLite

But, in the end I bundled sqlite with the Swift lib itself which requires no additional build steps from user and doesn't require adding flags to 'swift build':
https://github.com/zmeyc/SQLite.swift/blob/master/Package.swift
https://github.com/zmeyc/SQLite.swift/tree/master/Sources/CSQLite

Another problem with SPM is that it works only with 'master' branch now. So, if it's going to be supported, probably Swift3 branch has to be renamed to master, and current 'master' renamed to Stable and probably made the default on github. I believe other package managers don't have this limitation. This needs to be carefully tested.

Folder structure also needs to be adjusted a bit, but it's not going to be too hard.

groue:
Hello @zmeyc, thanks for your interest, and the context you have provided.

I prefer answering no.

First of all, GRDB focuses on applications, and applications do not need Swift Package Manager. At least, not yet. Future will tell.

Next, shipping GRDB as a Swift package would raise quality issues, because packaging implies server support, when the library is not ready for the server. Precisely speaking:

1. The behavior of GRDB in case of multiple writers on a single database is [not documented enough](https://github.com/groue/GRDB.swift#concurrency), without any clear roadmap to avoid concurrency errors.
    
    [Database changes observation](https://github.com/groue/GRDB.swift#database-changes-observation) is only reliable when there is a single writer to the database.
    
[...]

PR #202, opened 2017-04-10
SPM support

zmeyc:


Package.swift
```diff
@@ -0,0 +1,9 @@
+import PackageDescription
+
+let package = Package(
+    name: "GRDB.swift",
```
groue: Is the name of the package also used when one imports it from Swift with `import` ? Wouldn't be `GRDB` a better name, then? (Honest question, I really don't know)

Package.swift
```diff
@@ -0,0 +1,9 @@
...
+    dependencies: [
+        .Package(url: "https://github.com/groue/SQLiteSDK.git", majorVersion: 0)
```
groue: Is the convention of prefixing C packages with the letter C dead? The [SPM documentation](https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#require-system-libraries) still uses it: Clibgit, CJPEG...

Package.swift
```diff
@@ -0,0 +1,9 @@
...
+    ],
+    exclude: ["GRDB.xcworkspace", "Playgrounds", "SQLCipher", "SQLiteCustom", "Support", "DemoApps", "Tests", "Documentation"]
```
groue: This list looks like a maintenance burden. Not a big problem, but if gathering sources in a `Sources` directory would avoid it, then maybe we should follow the recommended convention.

Package.swift
zmeyc: I don't know if it affects anything at all, it doesn't seem to affect filenames in build dir nor import name. Can be changed to GRDB probably.
Module name is derived from directory name (it's currently 'GRDB', so it's 'import GRDB').

Package.swift
zmeyc: Yes, I used SQLiteSDK because there were 'import SQLiteSDK' in most of the files and this allowed not patching source code at all. I had to restore these imports as they were removed in latest master. It's probably better to rename the repo to "CSQLite" and adjust imports accordingly.

Package.swift
zmeyc: Yes, we can move GRDB dir into Sources/

There are three possible layouts:

1) Dirs for each module in root (currently used). The downside is having to exclude everything which SPM doesn't know how to handle.

2) For single module repos, Sources/ can contain source files **without** directories.

3) If at least one directory is present in Sources/, dirs are treated as distinct modules and no other source files are allowed at top level in Sources/
We can use this layout. So, it will become:
Sources/GRDB/...other files...

groue: OK, thanks for this detailed info. I don't feel like holding this PR just because we can't decide how to layout the files and directories: the status quo is good enough, even if we have to maintain an exclude list.

Package.swift
groue: OK, you made it clear that it's not quite clear ;-) In doubt let's rename to `GRDB`. This `.swift` suffix can only add confusion or nasty surprises ;-)

Package.swift
groue: All right! groue/SQLiteSDK has just been renamed https://github.com/groue/CSQLite :-)

groue:
Answering your [comment](https://github.com/groue/GRDB.swift/issues/72#issuecomment-292955764), @zmeyc:

> Regarding test_SPM_install: I'm not sure how it should work. To test that it still builds with SPM, it should be enough to call
> 
> ```
> swift package clean
> swift build
> ```

This checks that the package can build, and that's already great! OK, I'll add those lines to the Makefile myself. And this is well enough for a start. I'll want later to check that GRBD package can actually be integrated in a minimal binary package - but this PR doesn't need that.

> Unit tests require a flat layout:
> Tests/BlablaTests/File.swift
> Every dir in Tests/ becomes a module. I don't know where the code shared between tests should go, probably we'll have to ask SPM devs.

All right. And unit tests will need a deep overhaul in the future Linux PR... So let's not lose time with unit tests right now. Just make me a little favor and tell me that you were *actually* able to use SQLite with GRDB through SPM :-)

If I sum up:

- please rename the package to GRDB
- please rename the C package to CSQLite
- wait until I can update the Makefile, and build a minimal binary that loads GRDB through SPM: this will convince myself that this PR deserves to be merged for good, and that I'll be able to support it in the future 😄

groue:
(BTW @zmeyc, my personal machine is currently being repaired - I may be quick to talk but slow to actually perform my part of the job - please be patient, and assured I have the deepest interest in this PR)

zmeyc:
@groue One reason we might want to consider another name is because some other frameworks (notably Kitura and Perfect) have their own versions of CSQLite which will clash with GRDB's one. :(
I already had such problems with SwiftyJSON (everyone has their own fork). I don't know how to properly resolve these clashes or is there any automated mechanism in the plans.

zmeyc:
@groue But I hope that unlike SwiftyJSON, CSQLite is not imported and used in these frameworks by default. So maybe it's not a problem.

groue:
When this problem happens, we'll deal with it. Unless you feel like a trendsetter (which I'm not on the SPM topic), sticking with existing conventions is better than crafting an nth standard. Besides, server stacks already have their database packages. And last, GRDB is not ready at all (really not) for the server. I even think that a hypothetical server-enabled GRDB will live with another name in another repo. Did I miss something?

zmeyc:
@groue I fully agree.

ekimia:
@groue for what it's worth there is the application of GRDB for framework usage. SPM support fits in nicely with that use case.

groue:
@zmeyc I took the liberty to complete the renaming tasks on your pr_spm1 branch.

`swift build` performs as expected (without any error).

I still can't check integration since GRDB is not yet published as an SPM package, and top-of-tree development doesn't look available yet with Xcode 8.3.1. I'll do that.

Thanks for this PR, and your patience.

zmeyc:
@groue Thanks for finishing the PR, I suddenly felt ill yesterday. Hopefully I'll recover in a few days and will continus with Linux PR.
I've done a quick check:
```
 swift package init --type=executable
...
    dependencies: [
        .Package(url: "https://github.com/groue/GRDB.swift.git", majorVersion: 0)
    ]
...
```
and it seems everything is built and linked correctly.

Top of trunk development can be done like this:
```
swift package edit GRDB --revision master
```
After this SPM will use whatever is present in Packages/GRDB directory.
It can even be a symlink to another directory.

```
swift package generate-xcodeproj
```
also seems to work.

Btw, last tag seems to be misspelled (no 'v' prefix).

groue:
> I've done a quick check [...] and it seems everything is built and linked correctly.

Yes, thanks to you. There is a small executable in Tests/SPM which simply checks access to an in-memory database, and to the C API. Your tip for the top of trunk development makes it suitable for an automated test (see https://github.com/groue/GRDB.swift/commit/5804c044d513a9b1d93dd4f04b4641a6b2deeb34).

> Btw, last tag seems to be misspelled (no 'v' prefix).

It was because of an old tag without `v` that prevented SPM from performing well. Now everything is back in place, with consistent v-prefixed tags!

> Hopefully I'll recover in a few days and will continue with Linux PR.

Take care, and recover first and foremost :-)

PR #205, opened 2017-04-14
Linux support

zmeyc:
@groue 
Lots of crude workarounds for compilations problems on Linux. It's especially bad when working around NSScanner API mismatches. I propose using this as a starting point to get library into compilable state then getting rid of these #if os(Linux) one by one.

Compiles with Swift 3.1 Snapshot-2017-04-13. My app works with this patch, but I haven't runned library tests as they still need to be converted to SPM-friendly layout.

It won't compile with Swift 3.1 Release with one obscure compilation error in NSFileManager.

Setting up the environment on Ubuntu 16.04:
```
cd ~
wget https://swift.org/builds/development/ubuntu1604/swift-DEVELOPMENT-SNAPSHOT-2017-04-13-a/swift-DEVELOPMENT-SNAPSHOT-2017-04-13-a-ubuntu16.04.tar.gz
tar xzf swift-DEVELOPMENT-SNAPSHOT-2017-04-13-a-ubuntu16.04.tar.gz
```
In ~/.profile:
```
PATH="$HOME/swift-DEVELOPMENT-SNAPSHOT-2017-04-13-a-ubuntu16.04/usr/bin/:$PATH"
```
Then:
```
source ~/.profile
```


groue:
**All right, let's commit to Linux!**

That's a pretty big commitment, that I don't take lightly. As soon as Linux is added to the list of supported platforms, it will remain so, with all the maintenance consequences.

I'll list below the remaining tasks that I can see until I feel comfortable with this PR. If you don't mind, let's discuss below about the ones you feel like doing. 

- [ ] **Tests**
    
    That's meaty. `make test` does a lot a things. Particularly, most test are shared between platforms, and GRDB variants (System SQLite, custom SQLite, SQLCipher with and without encryption). And there are performance tests. Swift Package Manager `swift test` is not supported yet. I don't know which is the best path towards Linux tests, but maybe we can start by supporting `swift test` on an Apple OS first, and then have those tests run on Linux. The two blog posts by [Ole Begemann](https://oleb.net) look very interesting:
    
    - [Keeping XCTest in sync on Linux
    ](https://oleb.net/blog/2017/03/keeping-xctest-in-sync/)
    - [How to test a Swift package on Linux using Docker](https://oleb.net/blog/2017/03/testing-swift-packages-on-linux/)
    
    Maybe the topic of tests should be tracked in another issue.
    
- [ ] **Logs**
    
    Logging is different from query tracing, provided with `Configuration.trace`. Logging is about letting SQLite and GRDB speak when they feel like doing it.
    
    Using `print` on Linux and `NSLog` on Apple platforms makes the compiler happy, but we can do better.
    
    A little reflection on cross-platform logging is necessary, with [SQLITE_CONFIG_LOG](https://www.sqlite.org/c3ref/c_config_covering_index_scan.html#sqliteconfiglog) in mind (an SQLite feature not supported by GRDB yet).
    
    If a "logger" can be implemented with a simple `(String) -> Void` function, then we could auto-configure GRDB so that it uses `NSLog` when available, and prints on stderr otherwise (unless there is a more sensible default).

- [ ] **SQLite versions**
    
    You [wrote](https://github.com/groue/CSQLite/pull/1#discussion_r110670633):
    
    > It's actually needed only on Linux (for apt-get). It prints a hint if sqlite is not installed.
    >
    > 3.8.2 on Ubuntu 14.04
    >
    > 3.11.0 on 16.04
    
    GRDB uses availability checks to adapt its APIs to the SQLite versions that ship with iOS and macOS. When it compiles with the latest SQLCipher or SQLite (whose version is known), availability checks are useless and unused.
    
    Linux brings in a new challenge: support for *unknown* SQLite version (if we keep on using `.Apt("libsqlite3-dev")` in the [CSQLite package definition](https://github.com/groue/CSQLite/blob/master/Package.swift)).
    
    Does this mean runtime checks?
    
- [ ] **`String(format:)` vs. `NSString(format:)`**
    
    Can't we unify those, today?
    
- [ ] **Other `#if os(Linux)`**
    
    Those depend a lot on the Swift version, don't they?
    
- [ ] **Maintaining the Linux version on a Mac**
    
    Something has to be done to help mac users easily test the Linux version.
    
If you have more ideas, I'll add them above.

@zmeyc: I don't use Linux myself so I won't be able to complete all this alone. But be sure I'll support you if you are willing to give a big push.

zmeyc:
> I don't know which is the best path towards Linux tests, but maybe we can start by supporting swift test on an Apple OS first, and then have those tests run on Linux.

Sounds like a good idea to me. I also like the idea of testing with Docker image and generating LinuxMain with Sourcery as suggested in the articles.

> Maybe the topic of tests should be tracked in another issue.

Agreed.

> Logging is different from query tracing, provided with Configuration.trace. Logging is about letting SQLite and GRDB speak when they feel like doing it.
> 
> Using print on Linux and NSLog on Apple platforms makes the compiler happy, but we can do better.
> 
> A little reflection on cross-platform logging is necessary, with SQLITE_CONFIG_LOG in mind (an SQLite feature not supported by GRDB yet).
> 
> If a "logger" can be implemented with a simple (String) -> Void function, then we could auto-configure GRDB so that it uses NSLog when available, and prints on stderr otherwise (unless there is a more sensible default).

This should probably also be a separate thread, but trivial implementation can indeed be something like:

```
    public var logger: (_ text: String) -> () = {
       #if os(Linux)
       print($0)
       #else
       NSLog($0)
       #endif
    }
```
Where should it go to?

> GRDB uses availability checks to adapt its APIs to the SQLite versions that ship with iOS and macOS. When it compiles with the latest SQLCipher or SQLite (whose version is known), availability checks are useless and unused.
> 
> Linux brings in a new challenge: support for unknown SQLite version (if we keep on using .Apt("libsqlite3-dev") in the CSQLite package definition).
> 
> Does this mean runtime checks? Is there any opportunity to turn them into static checks?

We don't know which version is installed because system package updates can alter it. Also user can put CSQLite into editable state and replace it with his own repo with custom-built sqlite. So I don't think we can reliably check sqlite version statically.


> String(format:) vs. NSString(format:)
> Can't we unify those, today?

These can be unified, I tried not to touch OS X code in this patch. Ideally, these incompatibilities should be patched (or logged in Jira) first.

> Other #if os(Linux)
> 
> Those depend a lot on the Swift version, don't they?

I don't think it's possible to test for a specific snapshot version (or I don't know how). Checking for version number like 3.1 is not fine grained enough, so maintaining backwards compatibility is a problem. 

> Support latest Swift 3.1 Release
> 
> Eventually 😉

Do you mean the older one it doesn't build with? I don't think it's worth the effort, it's very unfinished on Linux anyway.

> Maintaining the Linux version on a Mac
> 
> Something has to be done to help mac users easily test the Linux version. Running tests locally? Configure the Github repository for Travis? I dunno. But without this, there won't be any long term Linux support.

This should be done, yes. Docker is a good idea imo.

I will be happy to contribute, but can't give time guarantees. I propose splitting it into small tasks. For example, starting with tests on OS X, then setting up docker so it's easier to check if Linux build is broken, then moving to Linux-specific issues.

zmeyc:
@groue SPM now supports custom target layouts: https://github.com/apple/swift-evolution/blob/master/proposals/0162-package-manager-custom-target-layouts.md

groue:
Hello @zmeyc,

My mac has finally been repaired, I'm back on the field. I've been playing with SPM a little bit, to get a better grasp of the tool.

The #206 PR has "solved" the logging point, although logging is currently not available for SPM (see #206 for details). Still, GRDB has a sane logging entry point now: it will accommodate Linux fine whenever we manage to make it available.

I've also introduced conflicts in your PR, I'm sorry. That's because in a failed attempt at building GRDB as several modules (SPM targets), I could eventually reduce some couplings in GRDB internals. GRDB is still a monolithic library, though.

zmeyc:
Hello @groue! Great news. :) I've sent PRs to fix SPM logging:
https://github.com/groue/CSQLite/pull/2
https://github.com/groue/GRDB.swift/pull/207
I'll fix Linux PR asap.

groue:
@zmeyc Very clever 👍

groue:
@zmeyc: `swift test` is not available, ready to support Linux.

zmeyc:
@groue Cool! That was fast :) Going to try Linux now.

groue:
> > I don't know which is the best path towards Linux tests, but maybe we can start by supporting swift test on an Apple OS first, and then have those tests run on Linux.
> 
> Sounds like a good idea to me. I also like the idea of testing with Docker image and generating LinuxMain with Sourcery as suggested in the articles.

I've made tests available for Swift Package Manager in #208, ready for Linux support.

> > GRDB uses availability checks to adapt its APIs to the SQLite versions that ship with iOS and macOS. When it compiles with the latest SQLCipher or SQLite (whose version is known), availability checks are useless and unused.
> > 
> > Linux brings in a new challenge: support for unknown SQLite version (if we keep on using .Apt("libsqlite3-dev") in the CSQLite package definition).
> > 
> > Does this mean runtime checks? Is there any opportunity to turn them into static checks?
> 
> We don't know which version is installed because system package updates can alter it. Also user can put CSQLite into editable state and replace it with his own repo with custom-built sqlite. So I don't think we can reliably check sqlite version statically.

All right, this means that the Linux PR will have to use runtime checks with the [`sqlite3_libversion_number`](https://www.sqlite.org/c3ref/libversion.html) function ([example](https://github.com/groue/GRDB.swift/blob/v0.106.2/GRDB/Core/Database.swift#L527)).

I thus expect the PR to modify *all* availability checks with code snippets like the following:

```swift
#if os(Linux)
    // SQLite version is unknown at compile time: use runtime checks
    func eventuallyAvailableFunction() {
        GRDBPrecondition(sqlite3_libversion_number() >= 3007014, "SQLite 3.7.14 required")
        ...
    }
#elseif USING_CUSTOMSQLITE || USING_SQLCIPHER
    // GRDBCustomSQLite and GRDBCipher always use the latest SQLite and
    // SQLCipher: no check is necessary (but beware: SQLCipher is always lagging
    // behind SQLite and may lack the feature).
    func eventuallyAvailableFunction() {
        ...
    }
#else
    // iOS, macOS, watchOS, ...
    @available(...)
    func eventuallyAvailableFunction() {
        ...
    }
#endif
```

> > String(format:) vs. NSString(format:)
> > Can't we unify those, today?
> 
> These can be unified, I tried not to touch OS X code in this patch. Ideally, these incompatibilities should be patched in Foundation (or logged in Jira) first.
> 
> > Other #if os(Linux)
> > 
> > Those depend a lot on the Swift version, don't they?
> 
> I don't think it's possible to test for a specific snapshot version (or I don't know how). Checking for version number like 3.1 is not fine grained enough, so maintaining backwards compatibility is a problem. 
> 
> > Support latest Swift 3.1 Release
> > 
> > Eventually 😉
> 
> Do you mean the older one it doesn't build with? I don't think it's worth the effort, it's very unfinished on Linux anyway.

I just wish that, if possible, that we avoid useless `#if os(Linux)`, and that we support stable releases. If we must remain on the bleeding edges of nightly builds, no problem: talk to me about that. It'll help documenting what, *exactly*, the Linux version of GRDB is.

> > Maintaining the Linux version on a Mac
> > 
> > Something has to be done to help mac users easily test the Linux version. Running tests locally? Configure the Github repository for Travis? I dunno. But without this, there won't be any long term Linux support.
> 
> This should be done, yes. Docker is a good idea imo.
> 
> I will be happy to contribute, but can't give time guarantees. I propose splitting it into small tasks. For example, starting with tests on OS X, then setting up docker so it's easier to check if Linux build is broken, then moving to Linux-specific issues.

I think GRDB is ready for you now. If you need my help, please tell me!

zmeyc:
@groue I fixed the conflicts with master branch. Your recent changes seem to have fixed the compilation issue on Swift 3.1 release, it builds cleanly now. :) Will send Docker PR soon.

groue:
Yes @zmeyc, good job: `swift built` passes on Linux :-)

Makefile
```diff
@@ -125,6 +126,10 @@ test_SPM:
 	$(SWIFT) package clean
 	$(SWIFT) test
 
+test_docker:
+	$(DOCKER) build --tag grdb .
```
groue: Please guard $(DOCKER) with `ifdef`, because if Docker is not installed then $(DOCKER) is empty and the Makefile executes weird commands. See example with $(POD) in the Makefile.

Maybe there is a better way to write Makefiles - but until we figure it out, let's use this convention for tools that may be missing.

groue:
OK, so now we're left with three main tasks:

1. Remove `#if os(Linux)` that we can avoid
2. Tests (your better guarantee against future regressions)
3. Revise availability checks for Linux (your better guarantee against deployment surprises)

Both tests and availability checks are also our better guarantee against issues opened by other GRDB users. *Answering issues [takes time](https://github.com/groue/GRDB.swift/pulse/monthly).* I thus think that it's just better to prevent bugs rather than fixing known issues after somebody has fallen into them.

zmeyc:
@groue In tests I get some compilation errors in exception handling.
`catch let error as? NSError` covers all exceptions on OS X, but produces error (missing 'try') on Linux because `Error` case is not handled. I made a quick test to find if these can be replaced with simple `catch {}`:

```
import Foundation

func f() throws {
  throw NSError(domain: "foo", code: 120, userInfo: ["bla": 123])
}

do {
    try f()
//} catch let error as NSError {
//    print("\(error.domain), \(error.code), \(error.userInfo)")
} catch {
    print("\(error._domain), \(error._code), \((error as! NSError).userInfo)")
}
```

_domain and _code work on both Linux & OS X. UserInfo is only accessible by casting to NSError though, but it's not used in tests anyway.

groue:
I admit I'm surprised how much the two flavors of Foundation are different. Thanks for doing this.

GRDB/Core/Support/Foundation/UUID.swift
```diff
@@ -17,7 +21,17 @@ extension NSUUID : DatabaseValueConvertible {
         guard let data = NSData.fromDatabaseValue(databaseValue), data.length == 16 else {
             return nil
         }
+        #if os(Linux)
```
groue: Maybe the `constructing an object of class type 'Self' with a metatype value must use a 'required' initializer` error can be addressed with the `cast()` function defined in [Utils.swift](https://github.com/groue/GRDB.swift/blob/v0.106.1/GRDB/Core/Utils.swift#L43-L46). See https://github.com/groue/GRDB.swift/blob/v0.106.1/GRDB/Core/Support/Foundation/NSData.swift#L17 for an example

zmeyc: Thanks, I'll try it.

zmeyc:
```
    func testCollation() throws {
...
            let length1 = string1.utf8.count
            let length2 = string2.utf8.count
            if length1 == length2 {
                return (string1 as NSString).compare(string2)
            } else if length1 < length2 {
                return .orderedAscending
            } else {
                return .orderedDescending
            }
...
```
In testCollation() is it safe to remove this cast to NSString or is it relying on side effects of NSString?

groue:
`compare` used to be missing in String, and only available on NSString. It's safe to use `String.compare` instead.

zmeyc:
@groue Added automated Tests/LinuxMain.swift generation using Sourcery.

`make sourcery` will install Sourcery and generate LinuxMain (can be done on OS X only).
Can be called repeatedly - subsequent calls won't rebuild sourcery.

`make sourcery_clean` when sourcery is no longer needed.

Next step is working on failing tests.

zmeyc:
@groue Docker cache was invalidated by COPY resulting in apt-get called each time, fixed. Still a bit slow, it's probably worth trying building from local tree. I also disabled parallel tests because the output was unreadable.

groue:
> @groue Added automated Tests/LinuxMain.swift generation using Sourcery.

I understand that you intend `make sourcery` to be run whenever the list of tests changes. Is this exact? If so, OK: LinuxMain.swift will ship in the repo, and make GRDB testable on Linux without the need for Sourcery. LGTM.

Four comments on the current state of the PR:

- Makefile: `make sourcery` is not a very good name. So is the "Code Generation" chapter of the Makefile: sourcery and code generations are not goals: Linux builds and tests are. Also I don't understand the necessity of `sourcery_clean` in the Makefile.

- Package.swift: I don't understand why sourcery has to be added as a dependency of GRDB, when all we need is a binary that we could build in a temporary directory. After all, we have that `make sourcery` rule, isn't it? It is that rule that needs Sourcery. Not GRDB.

- Makefile: The various `test -d` checks look smelly to me: Makefiles use rule dependencies in order to cascade intermediate build products into the ones that depend on them. I thus expect, for example, Linux testing rules to have a dependency on `Tests/GRDBTests/LinuxMain.swift`, and a rule for `Tests/GRDBTests/LinuxMain.swift` to depend on the phony rule that we'll run when more tests are added and need to regenerate that file. That phony rule would depend on the Sourcery binary, which is built by its own rule. etc, in a traditional Makefile way. I expect that it is possible to write it that way. I may overlook unexpected difficulties.

- Dockerfile: `swift build && swift test`. Does `swift test` depend on `swift build`? `&&` implies such a dependency. If not, please split this line. And also please add `swift build -c release`.

> Docker cache was invalidated by COPY resulting in apt-get called each time, fixed.

Didn't notice that :-)

> Still a bit slow, it's probably worth trying building from local tree. I also disabled parallel tests because the output was unreadable.

You're right: we currently need good diagnostics above all!

zmeyc:
> I understand that you intend make sourcery to be run whenever the list of tests changes. Is this exact?

Yes. It's even possible to automate this: LinuxMain.swift can depend on Tests/GRDBTests/* and run sourcery if there are any changes and sourcery is installed. But this need to be triggered by user-callable Makefile rule anyway (by "make test" or by a dedicated rule?)

Maybe it's better to keep it as standalone rule (as 'make sourcery' now), just using a better name.

> Makefile: make sourcery is not a very good name. So is the "Code Generation" chapter of the Makefile: sourcery and code generations are not goals: Linux builds and tests are.

I agree. "make update_linux_tests"?

> Also I don't understand the necessity of sourcery_clean in the Makefile.

'make sourcery' leaves artefacts behind (a package in editable state), sourcery_clean just rolls it back. I thought of rolling it back in 'make sourcery' itself, but it often needs to be called repeatedly and fetching / recompiling everything again is slow.

> Package.swift: I don't understand why sourcery has to be added as a dependency of GRDB, when all we need is a binary that we could build in a temporary directory. After all, we have that make sourcery rule, isn't it? It is that rule that needs Sourcery. Not GRDB.

It's one of official installation methods. I don't like the ENV hack too, I can roll it back and use 'git clone / swift build' in Makefile instead. Which method do you prefer: compiling it from source or downloading/unpacking a binary package?

> Makefile: The various test -d checks look smelly to me: Makefiles use rule dependencies in order to cascade intermediate build products into the ones that depend on them. I thus expect, for example, Linux testing rules to have a dependency on Tests/GRDBTests/LinuxMain.swift, and the rule for Tests/GRDBTests/LinuxMain.swift to depend on the phony rule that we'll run when more tests are added and we need to regenerate that file. That phony rule would depend on the Sourcery binary, which is built by its own rule. etc, in a traditional Makefile way. I expect that it is possible to write it that way. I may overlook unexpected difficulties - yet the goal is to write a traditional Makefile without flaky rules.

Ok, I'll try to rewrite it this way.

> Dockerfile: swift build && swift test. Does swift test depend on swift build? && implies such a dependency. If not, please split this line. And also please add swift build -c release.

It doesn't. Ok.

groue:
> > I understand that you intend make sourcery to be run whenever the list of tests changes. Is this exact?
> 
> Yes. It's even possible to automate this: LinuxMain.swift can depend on Tests/GRDBTests/* and run sourcery if there are any changes and sourcery is installed. But this need to be triggered by user-callable Makefile rule anyway (by "make test" or by a dedicated rule?)
> 
> Maybe it's better to keep it as standalone rule (as 'make sourcery' now), just using a better name.

I agree with you.

**The Makefile is, today, a tool that checks if GRDB is ready for a new release.** `make test` tells if it's OK on iOS, macOS, watchOS, if CocoaPods integration is OK, GRDBCipher, recently SPM, and soon Linux, etc. It only runs on macOS, and that is OK for today.

In this context, the Linux tests are totally dependent on macOS. And thus we could have LinuxMain.swift in .gitignore, and generate it each time the Makefile runs Linux tests.

But there is no reason to prevent Linux users from running `swift test`. Since Linux can't run sourcery (as you said above), we have to ship LinuxMain.swift in the repo.

We'll have the Makefile generate LinuxMain.swift before running tests on Linux via Docker. Can we have `Tests/GRDBTests/LinuxMain.swift` rule depend on Tests/GRDBTests/*, since LinuxMain.swift would then depend on itself? I don't quite know yet :-) Maybe the answer to this question will be the reason why the `Tests/GRDBTests/LinuxMain.swift` will be phony or not :-)

> > Makefile: make sourcery is not a very good name. So is the "Code Generation" chapter of the Makefile: sourcery and code generations are not goals: Linux builds and tests are.
> 
> I agree. "make update_linux_tests"?

OK!

> > Also I don't understand the necessity of sourcery_clean in the Makefile.
> 
> 'make sourcery' leaves artefacts behind (a package in editable state), sourcery_clean just rolls it back. I thought of rolling it back in 'make sourcery' itself, but it often needs to be called repeatedly and fetching / recompiling everything again is slow.
> 
> > Package.swift: I don't understand why sourcery has to be added as a dependency of GRDB, when all we need is a binary that we could build in a temporary directory. After all, we have that make sourcery rule, isn't it? It is that rule that needs Sourcery. Not GRDB.
> 
> It's one of official installation methods. I don't like the ENV hack too, I can roll it back and use 'git clone / swift build' in Makefile instead. Which method do you prefer: compiling it from source or downloading/unpacking a binary package?

The two last questions go together. The "official installation method" sucks.

`git clone / swift build` adds another network dependency on tests.

Personnally, I can live with the responsability of installing Sourcery before running tests. The Makefile doesn't install CocoaPods or jazzy, for example. We can always improve on this situation later.

Today Sourcery 0.5.9 is available with Homebrew, and 0.6.0 is available in the [download page](https://github.com/krzysztofzablocki/Sourcery/releases). That is OK for me. And we'll be able to remove `sourcery_clean`.

> > Makefile: The various test -d checks look smelly to me: Makefiles use rule dependencies in order to cascade intermediate build products into the ones that depend on them. I thus expect, for example, Linux testing rules to have a dependency on Tests/GRDBTests/LinuxMain.swift, and the rule for Tests/GRDBTests/LinuxMain.swift to depend on the phony rule that we'll run when more tests are added and we need to regenerate that file. That phony rule would depend on the Sourcery binary, which is built by its own rule. etc, in a traditional Makefile way. I expect that it is possible to write it that way. I may overlook unexpected difficulties - yet the goal is to write a traditional Makefile without flaky rules.
> 
> Ok, I'll try to rewrite it this way.

Makefiles are amazing.

> > Dockerfile: swift build && swift test. Does swift test depend on swift build? && implies such a dependency. If not, please split this line. And also please add swift build -c release.
> 
> It doesn't. Ok.

Thanks @zmeyc! I hope I don't sound too picky.

zmeyc:
@groue Not at all. :)

Brew's sourcery seems to broken (I guess it relies on Swift runtime which is not binary compatible yet):
```
bash-3.2$ sourcery
dyld: Symbol not found: __TWPVVs6Mirror14LegacyChildrens21RandomAccessIndexables
  Referenced from: /usr/local/bin/sourcery
  Expected in: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/libswiftCore.dylib
 in /usr/local/bin/sourcery
Abort trap: 6
```

zmeyc:
@groue Now test_docker depends on LinuxMain, LinuxMain depends on tests and sourcery presence; sourcery itself is installed from binary distribution. Package.swift changes reverted. So, it's now sufficient to call `make test_docker`.

zmeyc:
@groue I added both debug & release test targets to Dockerfile (I think `swift test` builds the library, so separate `swift build` step is not needed). Also, I'm not sure but I think there can be only one RUN command, so I used &&. When I used multiple RUNs, first one was ignored.

zmeyc:
This test fails:
```
Test Case 'DatabaseCollationTests.testDefaultCollations' started at 00:33:47.246
/package/Tests/GRDBTests/DatabaseCollationTests.swift:25: error: DatabaseCollationTests.testDefaultCollations : XCTAssertEqual failed: ("[1, 3, 2, 4, 6, 5, 8, 7]") is not equal to ("[1, 3, 2, 6, 7, 4, 5, 8]") - 
```
And indeed:
```
let s1 = "Z"
let s2 = "a"
print(s1 > s2)
```
`true` on Linux.
`false` on OS X.

It's being worked on:
https://bugs.swift.org/browse/SR-530
http://stackoverflow.com/questions/42317699/comparing-strings-in-swift-3-gives-false-results-linux

Should I comment out `unicodeCompare` on Linux with a link to SR-530 until it's fixed? As I understand this is going to break databases which use this collation. But probably better than silently using incorrect sort order.

groue:
> Dockerfile [...] I think there can be only one CMD command

OK

> This test fails: `DatabaseCollationTests.testDefaultCollations`

That's unexpected :-)

> As I understand this is going to break databases which use this collation.

Well, if the user uses this collation so that SQLite provides the same ordering as String, everything behaves as expected. Still, you're right that there are creeping bugs here: we have to update the documentation and say that this collation is platform-dependent, and unstable on Linux as of April 2017.

> Should I comment out unicodeCompare on Linux with a link to SR-530 until it's fixed?

I think that the problem is the test, not SR-530. Quoting the [documentation](https://github.com/groue/GRDB.swift#string-comparison): "`unicodeCompare` uses the built-in <= and == Swift operators." The test fails this, and should be updated to reflect the intent:

```swift
func testDefaultCollations() throws {
    let dbQueue = try makeDatabaseQueue()
        try db.execute("CREATE TABLE strings (id INTEGER PRIMARY KEY, name TEXT)")
        
        let strings = ["1", "2", "10", "a", "à", "A", "Z", "z"]
        for string in strings {
            try db.execute("INSERT INTO strings (name) VALUES (?)", arguments: [string])
        }
        
        func assert(collation: DatabaseCollation, ordersLike comparison: (String, String) -> ComparisonResult) throws {
            let dbStrings = try String.fetchAll(db, "SELECT name FROM strings ORDER BY name COLLATE \(collation.name)")
            let swiftStrings = strings.sorted { comparison($0, $1) == .orderedAscending }
            for (dbString, swiftString) in zip(dbStrings, swiftStrings) {
                XCTAssert(comparison(dbString, swiftString) == .orderedSame)
            }
        }
        
        try assert(collation: .unicodeCompare, ordersLike: {
            if $0 < $1 { return .orderedAscending }
            else if $0 == $1 { return .orderedSame }
            else { return .orderedDescending }
        })
        try assert(collation: .caseInsensitiveCompare, ordersLike: {
            $0.caseInsensitiveCompare($1)
        })
        try assert(collation: .localizedCaseInsensitiveCompare, ordersLike: {
            $0.localizedCaseInsensitiveCompare($1)
        })
        try assert(collation: .localizedCompare, ordersLike: {
            $0.localizedCompare($1)
        })
        try assert(collation: .localizedStandardCompare, ordersLike: {
            $0.localizedStandardCompare($1)
        })
    }
}
```

zmeyc:
@groue  I'm stuck with this test:
```
/package/Tests/GRDBTests/DatabaseFunctionTests.swift:330: error: DatabaseFunctionTests.testFunctionThrowingCustomError : XCTAssertTrue failed -
/package/Tests/GRDBTests/DatabaseFunctionTests.swift:331: error: DatabaseFunctionTests.testFunctionThrowingCustomError : XCTAssertTrue failed -
```
The code that fails:
```
            throw NSError(domain: "CustomErrorDomain", code: 123, userInfo: [NSLocalizedDescriptionKey: "custom error message"])
...
            } catch let error as DatabaseError {
...
                XCTAssertTrue(error.message!.contains("CustomErrorDomain"))
                XCTAssertTrue(error.message!.contains("123"))
```

On OS X `error.message` is
```
Error Domain=CustomErrorDomain Code=123 "custom error message" UserInfo={NSLocalizedDescription=custom error message}
```
On Linux:
```
custom error message
```
I can't find in code where `message` is assigned. It's declared in GRDB/Core/DatabaseError.swift as `String?` variable.

groue:
@zmeyc Were you able to restore `DatabaseFunctionTests.testFunctionThrowingCustomError`?

> I can't find in code where message is assigned

It's there: https://github.com/groue/GRDB.swift/blob/v0.106.1/GRDB/Core/Database.swift#L799

The error thrown from Swift when SQLite evaluates an SQL function is injected into SQLite, which later outputs it as an error thrown from the execution of the statement:

```swift
db.add(function: DatabaseFunction("throw") { _ in
    // injects error into SQLite
    throw ...
})
do {
    try db.execute("SELECT throw()") 
} catch {
    // the error throw from the function
}
```

The SQLite roundtrip loses information if the error is not a DatabaseError. That's because SQLite only accepts, from custom functions, an error code and an error message. NSErrors and other non-DatabaseError don't have codes that are relevant for SQLite: we thus use 1 (the general SQLITE_ERROR). Errors don't have standard string representation, so we use `"\(error)"`, hoping we lose as little information as possible. This code and message are turned into a DatabaseError for final consumption.

> The code that fails:
> 
> ```
>     throw NSError(domain: "CustomErrorDomain", code: 123, userInfo: [NSLocalizedDescriptionKey: "custom error message"])
>     ...
> } catch let error as DatabaseError {
>     ...
>     XCTAssertTrue(error.message!.contains("CustomErrorDomain"))
>     XCTAssertTrue(error.message!.contains("123"))
> ```
> 
> On OS X error.message is `Error Domain=CustomErrorDomain Code=123 "custom error message" UserInfo={NSLocalizedDescription=custom error message}`
> On Linux: `custom error message`

OK, that's because errors don't have the same string representation on Darwin and Linux: `"\(error)"` doesn't produce the same string. It lose error domain and code on Linux 👎 

OK. We can change the test so that those two lines are disabled on Linux, or change the string generation so that it includes domain and code for Linux. Maybe the second option is the best: if errors are eventually fixed on Linux we'll get errors that are too verbose - which is always better than lazy errors. What do you think?

zmeyc:
@groue Thanks for explanation! I found this by stepping in debugger so I deleted the original question. I think second option is better (i.e. to mimic error.description behavior in OS X). I'll just reconstruct the string from _domain, _code, _userInfo and localizedDescription. Even if it gets fixed, Linux code won't use .description and errors won't become verbose. We can remove this workaround later.

groue:
> availability checks [...] the Linux PR will have to use runtime checks with the sqlite3_libversion_number function.

Let's not rush on this. I'll try to look harder for static checks opportunities.

zmeyc:
@groue Encountered another problem: userInfo becomes nil when NSError is catched as Error, had to account for this in workaround. Correspondingly, Error.localizedDescription doesn't work so I had to use "\(error)" instead. Workaround breaks another test, but it was easy to fix.

zmeyc:
@groue Another problem: in `testFetchCursorStepFailure (DatabaseValueConvertibleFetchTests.swift)`, 
`_ = try cursor.next()` ultimately calls `Statement.next()`, which calls `sqlite3_step`.
On OS X it returns 1, then on subsequent call 21 (SQLITE_MISUSE).
On Linux it returns 1, then again 1 and test fails because it expects 21.
SQL string is `SELECT throw()`, throw is custom function.
What differs: on OS X custom function is called once (resulting in throw and 1), then 21 is returned.
```
function called!
sqlite3_result_error()
next() code=1
next() code=21
```
On Linux it's called again on subsequent step:
```
function called!
sqlite3_result_error()
next() code=1
function called!
sqlite3_result_error()
next() code=1
```

Code path seems to be the same, it calls `sqlite3_result_error(context, message, -1)` in Database.add's `catch NSError {...}` handler. But why on Linux it doesn't return 21 on next call, could it be different sqlite versions?

groue:
@zmeyc 3decb8fe6dfbd611d23862f34e9a042be02fd3da has fixed sqlite3_step() behavior on SQLite 3.11.0. I've refactored the way custom functions inject their errors into SQLite, aiming at simplifying tests. But they are still broken on Linux: I'll take care of this later this day.

groue:
OK: tests about NSError to DatabaseError conversions now pass.

zmeyc:
@groue I've fixed casts between NSDate/Date which triggered asserts in FoundationDateTests.testDate. But now it fails on milliseconds mismatch.
OS X (initial / after fetching from DB):
```
nanoseconds1: 123456789
nanoseconds2: 123000025
```
Linux:
```
nanoseconds1: 123456789
nanoseconds2: 16843009
```
I haven't determined yet at which step they're lost, could be Calendar's bug.
Related SR:
https://bugs.swift.org/browse/SR-3158
https://github.com/rbukovansky/Nanosecond/blob/master/Sources/main.swift

groue:
> But now it fails on milliseconds mismatch
> https://bugs.swift.org/browse/SR-3158

Wow. What do you think? We won't rewrite NSCalendar: let those tests fail on Linux. I'll update the doc.

zmeyc:
@groue I'll test it a bit more, maybe it can be worked around.

groue:
>> https://bugs.swift.org/browse/SR-3158: let those tests fail on Linux
> maybe it can be worked around

@zmeyc, this is such an horrible bug... i'll try to list a few options:

1. Fix DateFormatter
    
    Difficult
    
2. Store and load garbage milliseconds, with documentation (as in c51acfb4b18d9d966e5622ac1013a0af5eb34d5f)
    
    :-( This makes GRDB unreliable
    
3. Preserve the format, but store '000' instead of garbage milliseconds, and load 0 milliseconds.
    
    It is bad to store values with a false precision in the database. And it will be difficult to restore millisecond precision later, because this may introduce bugs in applications that explicitly or implicitly rely on the second precision.
    
4. Remove milliseconds from the database format
    
    This has the same backward-compatibility problems as option 3.
    
5. Let user specify how Dates are stored.
    
    This just pushes the responsibility of backward compatibility on the user's back.
    
6. Any more idea?

zmeyc:
@groue We could also help patching Foundation or pinging the author of that code.

Milliseconds are lost when constructing date:
```
let date = calendar.date(from: dateComponents)!
117195072.0, expected: 117184272.123457
```
https://github.com/apple/swift-corelibs-foundation
Calendar forwards calls to NSCalendar, NSCalendar constructs date by passing an array with single letter component type flags and additional info to _CFCalendarComposeAbsoluteTimeV:
```
    open func date(from comps: DateComponents) -> Date? {
        var (vector, compDesc) = _convert(comps)              
...
        return _CFCalendarComposeAbsoluteTimeV(...)
   }
```
_convert encodes nanoseconds as '#':
```
        _convert(comps.nanosecond, type: "#", vector: &vector, compDesc: &compDesc)
```
But in underlying code I don't see any checks for '#', it expects 'S" for milliseconds.

zmeyc:
It seems to be not too hard to fix. Internally it uses libicu's ucal_*
NSCalendar's letters are mapped to UCAL_* in
```
static UCalendarDateFields __CFCalendarGetICUFieldCodeFromChar(char ch) {
...
    case 'S': return UCAL_MILLISECOND;
```
So we can try converting nanoseconds to milliseconds and passing them as 'S' in NSCalendar.swift.

Also, in `_CFCalendarComposeAbsoluteTimeV` msecs should probably be initialized to zero like rest of the fields.

zmeyc:
@groue Apple's implementation of CFCalendar doesn't support '#', so it should be fixed at Foundation level probably.
https://opensource.apple.com/source/CF/CF-855.17/CFCalendar.c

groue:
@zmeyc: I've updated FoundationDateTests so that it avoids depending on broken DateComponents.nanosecond, and fixed Date.fromDatabaseDateComponents(). All FoundationDateTests now tests pass.

groue:
Particularly, FoundationDateTests.testDate() checks that Date is correctly serialized (no loss of accuracy).

zmeyc:
```
+                return cast(NSDate(timeIntervalSince1970: date.timeIntervalSince1970))
```
@groue Internally NSDate uses timeIntervalSinceReferenceDate, should I update casts to use it?

groue:
> Internally NSDate uses timeIntervalSinceReferenceDate, should I update casts to use it?

Yes, why not?

zmeyc:
@groue I changed testDateSerialization to look like this:
```
    func testDateSerialization() throws {
        let datePairs: [(TimeInterval, String)] = [
            (117195072.1234, "1973-09-18 10:11:12.123"),
            (117195072.9876, "1973-09-18 10:11:12.987")
        ]
        for pair in datePairs {
            let dbQueue = try makeDatabaseQueue()
                let date = Date(timeIntervalSince1970: pair.0)
                try db.execute("INSERT INTO dates (creationDate) VALUES (?)", arguments: [date])
                let string = try String.fetchOne(db, "SELECT creationDate FROM dates")!
                print("\(string) == \(pair.1)")
                XCTAssertEqual(string, pair.1)
            }
        }
    }
```
And it fails with second date:
```
1973-09-18 10:11:12.123 == 1973-09-18 10:11:12.123
1973-09-18 10:11:12.123 == 1973-09-18 10:11:12.987
```
In fromDatabaseComponents I replaced round() with:
```
let seconds = date.timeIntervalSince1970.rounded(.down) + Double(nanoseconds) / 1e9
```
but still the same result. Debugging this right now.

groue:
.9876 rounds to 988, this may be why?

zmeyc:
> .9876 rounds to 988, this may be why?

You're right, but looks like it's fetching the previous value again. I changed .1234 to another value and it appears twice.

groue:
@zmeyc: I know why it fails :-) It's because the database is reused, and you don't load the date you expect

groue:
Database cleanup happens in teardown(). If you want a unique db, use `makeDatabaseQueue(filename: ...)`. Or insert the date in a rollbacked transaction?

groue:
@zmeyc Do you know why NSDecimalNumber and NSNumber tests crash without any error message? NSNumber.databaseValue behaves very very oddly!

zmeyc:
@groue No idea. :( I haven't looked into it yet.

groue:
We've already done a lot 😅

zmeyc:
@groue Not related to crash, but another issue to be aware of:
https://bugs.swift.org/browse/SR-90
http://stackoverflow.com/questions/39254804/behavior-of-nsnumbervalue-in-xcode-compared-to-open-source-swift-distribution

It's hard to stop when we're almost done. :)

groue:
[SE-0170](https://github.com/apple/swift-evolution/blob/master/proposals/0170-nsnumber_bridge.md) is on the way, which directly addresses [SR-90](https://bugs.swift.org/browse/SR-90). Not approved yet, though :-)

zmeyc:
@groue I've done a bit more debugging. Had to add `fflush(stdout)` to see print's output.

These crashes occur on `String(cString: NSNumber.objCType)` calls.
NSDecimalNumber's test calls .databaseValue which resolves to `NSNumber.databaseValue` and crashes on this line:
```
        switch String(cString: objCType) {
```

objCType returns C strings on OS X, but a single character (not a pointer!) on Linux. Also, they work inconsistently ("c" on OS X, 'B' on Linux). For example:
```
let x = true as NSNumber
print(x); fflush(stdout)
print(String(describing: x.objCType)); fflush(stdout)
print(String(cString: x.objCType)); fflush(stdout)
```

OS X:
```
1
0x00007fff75c43a67
c
```

Linux:
```
1
0x0000000000000042  // ASCII 66: 'B'
0  swift           0x0000000003632048
1  swift           0x0000000003632786
2  libpthread.so.0 0x00007f6f0145b390
3  libc.so.6       0x00007f6effe16b96 strlen + 38
```

'B' and 'c' are defined in NSObjCRuntime.swift:
```
internal enum _NSSimpleObjCType : UnicodeScalar {
...
    case Char = "c"
    case Bool = "B"
```

I looked how SwiftyJSON worked around these API inconsistencies. Older versions used to have `String(cString:)`:
```
private let trueObjCType = String(cString: trueNumber.objCType)
private let falseObjCType = String(cString: falseNumber.objCType)
...
extension NSNumber { ...
        let objCType = String(cString: self.objCType)
        if (... && objCType == trueObjCType)
```

Newer ones switched to `String(describing:)`:
```
private let trueObjCType = String(describing: trueNumber.objCType)
private let falseObjCType = String(describing: falseNumber.objCType)
...
extension NSNumber { ...
            let objCType = String(describing: self.objCType)
            if (... && objCType == trueObjCType)
```

Of course I had a question if these pointers can be reliably compared on OS X. Seems they produce the same pointer value for same type:

```
let x = true as NSNumber
let y = NSNumber(booleanLiteral: false)
print(x); fflush(stdout)
print(String(describing: x.objCType)); fflush(stdout)
print(String(cString: x.objCType)); fflush(stdout)
print(y); fflush(stdout)
print(String(describing: y.objCType)); fflush(stdout)
print(String(cString: y.objCType)); fflush(stdout)
```

OS X:
```
1
0x00007fff75c43a67
c
0
0x00007fff75c43a67
c
```
Linux (without cString: constructors):
```
1
0x0000000000000042
0
0x0000000000000042
```

Also, I observed that using NSNumber(**integerLiteral**: 1) produces "q" (LongLong) on OS X and 'q' on Linux correspondingly, but this is likely to change when SE-0170 will be implemented. As the compatibility will be lost anyway, I guess single-char approach as we currently see on Linux will be adopted on OS X too.

zmeyc:
@groue I've fixed objCType casts and returned to fixing NSDecimalNumber tests.
First crash can be reproduced with the simplified code below:
```
let n = NSDecimalNumber(string: "9223372036854775808")
print("n=\(n)"); fflush(stdout)
let v = n.int64Value // crash
print("v=\(v)"); fflush(stdout)
```

```
n=9223372036854775808
0  swift            0x0000000003632048
1  swift            0x0000000003632786
2  libpthread.so.0  0x00007f571bda0390
3  libFoundation.so 0x00007f57157ad4e5 _TFC10Foundation15NSDecimalNumberg10int64ValueVs5Int64 + 101
4  libFoundation.so 0x00007f571c1c81f3 _TFC10Foundation15NSDecimalNumberg10int64ValueVs5Int64 + 111258995
```

zmeyc:
Interestingly, that with smaller number it doesn't crash but loses precision. :)
```
let n = NSDecimalNumber(string: "922337203685477580")
print("n=\(n)"); fflush(stdout)
let v = n.int64Value
print("v=\(v)"); fflush(stdout)
```
OS X:
```
n=922337203685477580
v=922337203685477580
```
Linux:
```
n=922337203685477580
v=922337203685477632
```

groue:
@zmeyc Reading your last comments, it looks like NSNumber is a recipe for the cardinal sin of a persistence library: silent data loss 😅

zmeyc:
@groue Docker ate all my disk space (42+ gb), you'll probably hit this bug too. ~/Library/Containers grows when images are created. This thread describes some workarounds: https://github.com/docker/for-mac/issues/371
Easiest is resetting to factory defaults in Docker client.

groue:
Thanks for tip, @zmeyc.

What do you think about the NSNumber/NSDecimalNumber situation? Should we remove support for them until Foundation is fixed?

zmeyc:
@groue Sorry for my recent lack of activity, been very busy with work. I plan to return to Linux port in a week or two. Regarding removing NSNumber/NSDecimalNumber - how much rewriting would it require in the lib? Does GRDB rely on them internally or are they just convenience adapters?

groue:
> @groue Sorry for my recent lack of activity, been very busy with work. I plan to return to Linux port in a week or two.

No offense @zmeyc! The PR is still there, and you'll be just as welcome as before when you're ready!

> Regarding removing NSNumber/NSDecimalNumber - how much rewriting would it require in the lib? Does GRDB rely on them internally or are they just convenience adapters?

GRDB does not rely on NSNumber. It extends NSNumber with DatabaseValueConvertible conformance as a convenience, should anybody want to store NSNumbers. I don't see much trouble removing this conformance for Linux: NSNumber is not widely used, and [NSDecimalNumber is not a good fit for SQLite](https://github.com/groue/GRDB.swift#nsnumber-and-nsdecimalnumber).

groue:
Closing due to inactivity

PR #209, opened 2017-04-23
Add test_docker Makefile target

zmeyc:
Run `make test_docker` to try building GRDB on official Swift Docker image.
The code in 'pr_linux' (https://github.com/groue/GRDB.swift/pull/205) branch should build cleanly, tests aren't working yet and are commented out in Dockerfile.
I've followed this article: https://oleb.net/blog/2017/03/testing-swift-packages-on-linux/

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git/
+.build/
+!.build/checkouts/
+!.build/repositories/
+!.build/workspace-state.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM swift:3.1
+
+WORKDIR /package
+
+COPY . ./
+
+# Not neccessary because dependencies are copied from host
+# but let's ensure they're in clean state
+RUN swift package fetch
+RUN swift package clean
+
+# https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/
+RUN apt-get update && \
+    apt-get install -y libsqlite3-dev && \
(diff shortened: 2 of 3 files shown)


groue:
> ```
> Sending build context to Docker daemon 5.612 GB
> ```
>
> What is this enormous amount of data sent to docker daemon?
>
> According to https://docs.docker.com/engine/reference/builder/, it is all about the "PATH", whatever that means.
>
> I'm not sure that PATH is correct.

My mistake, this is normal: my local copy of GRDB contains the whole build of Realm, used for [performance tests](https://github.com/groue/GRDB.swift/wiki/Performance). This beast weights 5.1 GB.

zmeyc:
> Did I miss something? You write:
> 
> The code in 'pr_linux' (#205) branch should build cleanly, tests aren't working yet and are commented out in Dockerfile.
> I'm not sure to understand what is the goal of this PR. Shouldn't those changes go into #205, so that swift build has any chance to run successfully?

I split it out because it's unrelated to Linux, it just setups Docker. Code in master won't build in container, but I'll merge it to pr_linux.

groue:
> I split it out because it's unrelated to Linux, it just setups Docker. Code in master won't build in container, but I'll merge it to pr_linux.

I understand. Since it setups a command that fails, I prefer to invalidate this PR. Those changes should end up directly in #205 where they belong.

zmeyc:
@groue Merged to pr_linux.

Issue #211, opened 2017-05-12
Support SQLCipher through Cocoapods


hartbit:
I'm currently using the SQLCipher variant of GRDB, which forces me to forgo managing it with Cocoapods. Now that I want to start using `RxGRDB`, it gets worse: I can't get that through Cocoapods as it would pull the version of GRDB I don't want.

How difficult is it to support SQLCipher through Cocoapods? I know that FMDB does it in their [Podspec](https://github.com/ccgus/fmdb/blob/master/FMDB.podspec) but I don't know how applicable it is to GRDB.

groue:
Hello @hartbit!

> How difficult is it to support SQLCipher through Cocoapods?

I don't know 😉. And I don't know either how to have RxGRDB use GRDBCipher.

I guess we'd need subspecs in both GRDB.swift.podspec and RxGRDB.podspec, but I'm not yet familiar with "advanced" CocoaPods setup.

We also have https://github.com/stephencelis/SQLite.swift/blob/master/SQLite.swift.podspec#L60-L69 as a reference.

groue:
Thanks to @sroddy it is now possible to install GRDBCipher with CocoaPods: I'm closing this issue. However RxGRDB is not ready yet, though: the story continues on https://github.com/RxSwiftCommunity/RxGRDB/issues/6.

PR #212, opened 2017-05-14
SQLCipher support with Cocoapods

sroddy:
Hi,
I've refactored the actual structure of the Podspec in order to support SQLCipher as a different subspec.
It also needed a minor fix in a couple of header files in order to include the right sqlite3.h file.
See #211 for reference.

It should be noted that there might be other Cocoapods library that link to the official SDK's SQLite package so that it might be needed to adjust the project file in order to make sure that the SQLCipher Framework is linked before the sqlite3 library.
One example is the very famous `FirebaseAnalytics` pod.
https://discuss.zetetic.net/t/important-advisory-sqlcipher-with-xcode-8-and-new-sdks/1688

I've seen that GRDB already has a check in place to fail as soon as possible during execution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4579,13 +4579,14 @@ try controller.performFetch()
 
 **GRDB can encrypt your database with [SQLCipher](http://sqlcipher.net) v3.4.0.**
 
-Using CocoaPods:
+You can use [CocoaPods](http://cocoapods.org/) (version 1.1 or higher), and specify in your `Podfile`:
 
 ```ruby
-pod 'GRDB.swift/SQLCipher'
+use_frameworks!
+pod 'GRDBCipher'
 ```
 
-Using a manual installation of GRDB:
+Alternatively, perform a manual installation of GRDB and SQLCipher:
 
 1. Clone the GRDB.swift git repository, checkout the latest tagged version, and download SQLCipher sources:
     
@@ -4601,12 +4602,12 @@ Using a manual installation of GRDB:
 
 4. Add the `GRDBCipher.framework` from the targetted platform to the **Embedded Binaries** section of the **General**  tab of your target.
 
-5. Import `GRDBCipher` module in your project's files
-
 
 **You create and open an encrypted database** by providing a passphrase to your [database connection](#database-connections):
 
 ```swift
+import GRDBCipher
+
 var configuration = Configuration()
 configuration.passphrase = "secret"
 let dbQueue = try DatabaseQueue(path: "...", configuration: configuration)
--- a/README.md
+++ b/README.md
@@ -4579,7 +4579,12 @@ try controller.performFetch()
 
 **GRDB can encrypt your database with [SQLCipher](http://sqlcipher.net) v3.4.0.**
 
-This requires a manual installation of GRDB:
+Using CocoaPods
+```ruby
+pod 'GRDB.swift/SQLCipher'
+```
+
+Using a manual installation of GRDB:
 
 1. Clone the GRDB.swift git repository, checkout the latest tagged version, and download SQLCipher sources:
     
@@ -4587,7 +4592,7 @@ This requires a manual installation of GRDB:
     cd [GRDB.swift directory]
     git checkout v0.107.0
     git submodule update --init SQLCipher/src
-    ````
+    ```
     
 2. Embed the `GRDB.xcodeproj` project in your own project.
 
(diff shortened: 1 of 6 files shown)


groue:
Hello @sroddy! Many thanks, @hartbit will be happy :-) I'll check and merge this PR very soon.

hartbit:
Oh yes :D Very happy.

README.md
```diff
@@ -4579,27 +4579,33 @@ try controller.performFetch()
 
 **GRDB can encrypt your database with [SQLCipher](http://sqlcipher.net) v3.4.0.**
 
-This requires a manual installation of GRDB:
+Using CocoaPods
```
groue: Please add a new line and a colon.

README.md
```diff
@@ -4579,27 +4579,33 @@ try controller.performFetch()
...
+Using a manual installation of GRDB:
 
 1. Clone the GRDB.swift git repository, checkout the latest tagged version, and download SQLCipher sources:
     
     ```sh
     cd [GRDB.swift directory]
     git checkout v0.107.0
     git submodule update --init SQLCipher/src
-    ````
+    ```
```
groue: Thanks 😅

README.md
```diff
@@ -4579,27 +4579,33 @@ try controller.performFetch()
...
     
 2. Embed the `GRDB.xcodeproj` project in your own project.
 
 3. Add the `GRDBCipherOSX` or `GRDBCipheriOS` target in the **Target Dependencies** section of the **Build Phases** tab of your application target.
 
 4. Add the `GRDBCipher.framework` from the targetted platform to the **Embedded Binaries** section of the **General**  tab of your target.
 
+5. Import `GRDBCipher` module in your project's files
```
groue: This is not currently true with this PR: currently one has to `import GRDB`. Is it possible to have the `GRDB.swift/SQLCipher` pod define the `GRDBCipher` module? Maybe with a specific `module_name` for the `SQLCipher` subspec?

Support/GRDB.h
```diff
@@ -7,4 +7,9 @@ FOUNDATION_EXPORT double GRDB_VersionNumber;
 FOUNDATION_EXPORT const unsigned char GRDB_VersionString[];
 
 #import <GRDB/GRDB-Bridging.h>
-#import <GRDB/sqlite3.h>
+
+#if SQLITE_HAS_CODEC
+    #import <SQLCipher/sqlite3.h>
```
groue: I'd like to see a `#ifdef COCOAPODS` guard the inclusion of `<SQLCipher/sqlite3.h>`.

Support/grdb_config.c
```diff
@@ -1,5 +1,10 @@
 #include <GRDB/grdb_config.h>
-#include <GRDB/sqlite3.h>
+#if SQLITE_HAS_CODEC
+    #include <SQLCipher/sqlite3.h>
```
groue: I'd like to see a `#ifdef COCOAPODS` guard the inclusion of `<SQLCipher/sqlite3.h>`.

README.md
sroddy: That step is in the manual installation steps.
I can do that but I'm not sure I would name the module differently... after all I don't think you can really use the both of them together.
But of course I can add a module_name if you prefer... let me know!

sroddy:
Hi @groue, I've updated the PR according to your feedback. Let me know what you think about the module name, and thanks a lot for the amazing library!

README.md
groue: "That step is in the manual installation steps", because you moved it there.

Currently GRDB exposes three different modules: GRDB, GRDBCipher, and GRDBCustom (for [custom SQLite builds](https://github.com/groue/GRDB.swift/blob/master/Documentation/CustomSQLiteBuilds.md)). I'd rather be *consistent*: either expose a single module name `GRDB` for all frameworks and installation methods (but this may create a new set of problems), or use the `GRDBCipher` module name for GRDB+SQCipher whatever the installation method.

So yes, please, if CocoaPods supports it, I'd like a specific `module_name`. And you will be able to restore the `import GRDBCipher` line in its previous place in the README.

groue:
Thanks @sroddy for the changes. As I said above, I'd like a custom module name if CocoaPods supports it, yes. Later I'll look after #211, and specifically, a SQLCipher-enabled version of [RxGRDB](http://github.com/RxSwiftCommunity/RxGRDB). Since you taught me how to write CocoaPods subspecs, it should be easy :-)

sroddy:
@groue unfortunately this doesn't seem to be the case :(
https://github.com/CocoaPods/CocoaPods/issues/3979

groue:
According to https://github.com/CocoaPods/CocoaPods/issues/3979#issuecomment-128594234, we can have a dedicated podsepc for GRDBCipher, instead of a specific subspec for GRDB. Can you please investigate?

groue:
> We can have a dedicated podspec for GRDBCipher, instead of a specific subspec for GRDB. Can you please investigate?

I'm on it.

groue:
👍 Many thanks to you, @sroddy!

groue:
Something is weird: I can't load GRDBCipher any longer. See https://github.com/CocoaPods/CocoaPods/issues/6720

PR #221, opened 2017-05-27
Swift 4 / GRDB 2.0

groue:
This PR is a work-in-progress towards Swift 4 support, and the release of GRDB 2.0.

Highlights:

- **[SE-0148](https://github.com/apple/swift-evolution/blob/master/proposals/0148-generic-subscripts.md)**: Use row subscript to extract column values:
    
    ```diff
     let nameColumn = Column("name")
    -let name: String = row.value(atIndex: 0)
    -let name: String = row.value(named: "name")
    -let name: String = row.value(nameColumn)
    +let name: String = row[0]
    +let name: String = row["name"]
    +let name: String = row[nameColumn]
    ```
- **[SE-0166](https://github.com/apple/swift-evolution/blob/master/proposals/0166-swift-archival-serialization.md)**: Support for the `Codable` protocol.
    
    Record types that adopt the `Codable`, `Decodable`, or `Encodable` protocols of the Swift Standard Library derive their database support from the `init(from: Decoder)` and `encode(to: Encoder)` methods, which happen to be automatically generated. This means that you no longer have to write boilerplate code :tada:!
    
    ```swift
    enum Color: String, Codable {
        case red, green, blue
    }
    
    struct Player : RowConvertible, Persistable, Codable {
        static let databaseTableName = "players"
        let name: String
        let score: Int
        let color: Color
    }
    
    // Automatically derived from Codable:
    //
    // extension Player {
    //     init(row: Row) {
    //         name = row["name"]
    //         score = row["score"]
    //         color = row["color"]
    //     }
    //     
    //     func encode(to container: inout PersistenceContainer) {
    //         container["name"] = name
    //         container["score"] = score
    //         container["color"] = color
    //     }
    // }
    
        let arthur = Player(name: "Arthur", score: 100, color: .red)
        try arthur.insert(db)
        
        let players = try Player.fetchAll(db) // [Players]
    }
    ```
    
    Many thanks to @itaiferber for his support.

- **Record protocols have more precise semantics**: RowConvertible *reads database rows*, TableMapping *builds SQL requests*, and Persistable *writes*.
    
    In GRDB 2.0, all methods of TableMapping that would write have been moved to Persistable. This means that being able to write `Player.fetchAll(db)` does not imply that `Player.deleteAll(db)` is available.
    
- **Records can now specify the columns they feed from.**
    
    In previous version of GRDB, `SELECT *` was the norm, with a [special exception](https://github.com/groue/GRDB.swift/tree/v1.3.0#exposing-the-rowid-column) for the `rowID` column.
    
    GRDB 2.0 introduces `databaseSelection`, which allows any type to define its preferred set of columns:
    
    ```swift
    struct Player : RowConvertible, TableMapping {
        let id: Int64
        let name: String
        
        enum Columns {
            static let id = Column("id")
            static let name = Column("name")
        }
        
        init(row: Row) {
            id = row[Columns.id]
            name = row[Columns.name]
        }
        
        static let databaseTableName = "players"
        static let databaseSelection: [SQLSelectable] = [Columns.id, Columns.name]
    }
    
    // SELECT id, name FROM players
    let players = Player.fetchAll(db)
    ```

Travis support comes from #222 by @swiftlyfalling.


hartbit:
Very happy you got this to work! So looking forward to it!

iNoles:
I don't know you heard or seen this before

According to Xcode 9 Beta 1's release note, 
`SQLite APIs can now be imported into Swift under the module name SQLite3`

PR #225, opened 2017-06-02
GRDB 1.0

groue:
This PR aims at briging GRDB to v1.0.

**The 1.0 version requires strong backward compatibility guarantees: the goal is  API stabilization.**

Not all APIs will be locked, though: some GRDB APIs will be made *experimental*.

- [X] **Identify and drop legacy features**
    
    Legacy features are APIs dragged from previous versions of GRDB, and either have a newer and better alternative, or are of insufficient quality.
    
    - [X] Drop DatabaseCoder. `DatabaseCoder` is a helper type that helps NSCoding types to be stored and loaded from blob columns. Not only does it provide little help, it is also threatened by [SE-0166](https://github.com/apple/swift-evolution/blob/master/proposals/0166-swift-archival-serialization.md), and it has already been removed from the main documentation
    - [X] Question poorly supported/implemented/used SQL function and operators in the [query interface](https://github.com/groue/GRDB.swift#expressions)
        - [X] Drop QueryInterfaceRequest.exists()
        - [X] Drop QueryInterfaceRequest.contains(_:)
    - [X] Question `Type.fetchAll(db, request)` now that `request.asRequest(of: Type.self).fetchAll(db)` exists. -> KEEP
    - [X] Restore default transaction kind to DEFERRED (the SQLite default).
    
    
- [X] **Identify the stable and experimental parts of the library**
    
    The most used core features of GRDB will be made stable: [database connections](https://github.com/groue/GRDB.swift#database-connections), [fetching methods](https://github.com/groue/GRDB.swift#fetching-methods), [persistence methods](https://github.com/groue/GRDB.swift#persistence-methods), [record protocols](https://github.com/groue/GRDB.swift#records), the [Record class](https://github.com/groue/GRDB.swift#record-class), and the [request protocols](https://github.com/groue/GRDB.swift#custom-requests).
    
    Features that can not change without breaking databases created with former versions of GRDB will be made stable too. This includes conversions of Swift types to and from [database values](https://github.com/groue/GRDB.swift#values), [migrations](https://github.com/groue/GRDB.swift#migrations), and [built-in collations](https://github.com/groue/GRDB.swift#string-comparison).
    
    However, some features are advanced, or young, or lack user feedback, and may be deemed experimental. Experimental features will not be protected by semantic versioning:
    
    - [X] Question [DatabaseReader and DatabaseWriter](https://github.com/groue/GRDB.swift#databasewriter-and-databasereader-protocols) -> STABLE
    - [X] Question [transaction observers](https://github.com/groue/GRDB.swift#database-changes-observation) -> STABLE
    - [X] Question [row adapters](https://github.com/groue/GRDB.swift#row-adapters) -> EXPERIMENTAL
    - [X] Question [query interface extensibility](https://github.com/groue/GRDB.swift/blob/master/Documentation/ExtendingGRDB.md) -> SQLExpressionFunction, SQLFunctionName, SQLExpressionUnary, SQLExpressionBinary, SQLExpressionFunction are EXPERIMENTAL
    - [X] Question [custom FTS5 tokenizers](https://github.com/groue/GRDB.swift/blob/master/Documentation/FTS5Tokenizers.md) -> STABLE

- [X] **Fix known design issues**
    
    - [X] Migrations aren't easily testable enough: `DatabaseMigrator.migrate(_:upTo:)`
    - [X] Rename TypedRequest's associated type from `Fetched` to `RowConverter`, because if TypedRequest needs a type that can decode rows, the type of the decoded rows does not need to be identical to the decoder type.
    - [X] Improve support for record types that define columns
        
        GRDB is very "stringly-typed". This is a *feature* that will remain. Yet some users (ping @hartbit) like to put the strong typing system of Swift to some use, and their programming style could be granted with improved support:
        
        ```swift
        struct Person : RowConvertible, MutablePersistable {
            static let databaseTableName = "persons"
            
            var id: Int64?
            let name: String
            let email: String?
            
            // The hard-coded list of columns that deserve some attention
            enum Columns {
                static let id = Column("id")
                static let name = Column("name")
                static let email = Column("email")
            }
            
            init(row: Row) {
                // Isn't there room for improvement here?
                id = row.value(Columns.id)
                name = row.value(Columns.name)
                email = row.value(Columns.email)
            }
            
            var persistentDictionary: [String: DatabaseValueConvertible?] {
                // Isn't there room for improvement here?
                return [
                    Columns.id.name: id,
                    Columns.name.name: name,
                    Columns.email.name: email,                    
                ]
            }
        }
        
        // Isn't there room for improvement here?
        let persons = try Persons.filter(Person.emailColumn != nil).fetchAll(db)
        ```
        
        - [x] [Replace "persistentDictionary" with PersistenceContainer](https://github.com/groue/GRDB.swift/pull/226)

The [lack of support for joined queries](https://github.com/groue/GRDB.swift/issues/176) will not be adressed by this PR. When joins eventually ship, maybe with breaking changes, this will make a great v2.0 release.

GRDB users and contributors, feel free to express your pet-peeves and wishes!


hartbit:
> Improve support for record types that define columns
Isn't this kind fo fixed with Swift 4's Encodable?

groue:
> > Improve support for record types that define columns
>
> Isn't this kind fo fixed with Swift 4's Encodable?

No @hartbit, for several reasons:

- [SE-0166](https://github.com/apple/swift-evolution/blob/master/proposals/0166-swift-archival-serialization.md) doesn't let Codable types change the way they encode themselves depending on the coder:

    ```swift
    // This type requires mandatory customized database coding/decoding because
    // SQL rows are *flat*, when Codable stores the location property in a
    // hierarchical way (good for JSON and PList):
    struct POI: RowConvertible, Persistable, Codable {
        let name: String
        let location: Location
        
        init(row: Row) {
            name = row["name"]
            location = Location(
                latitude: row["latitude"]
                longitude: row["longitude"]
            )
        }
        
        var persistenDictionary: [String: DatabaseValueConvertible?] {
            return [
                "name": name,
                "latitude": location.latitude,
                "longitude": location.longitude,
            ]
        }
    }
    
    struct Location: Codable {
        let latitude: Double
        let longitude: Double
    }
    ```

- [SE-0166](https://github.com/apple/swift-evolution/blob/master/proposals/0166-swift-archival-serialization.md) generates code where the coding keys are **private**, which means that custom row decoding/encoding can't use them:

    ```swift
    struct POI: RowConvertible, Persistable, Codable {
        let name: String
        let location: Location
    
        init(row: Row) {
            // Compiler errors: CodingKeys is unavailable due to private access level
            name = row[CodingKeys.name]
            location = Location(
                latitude: row[Location.CodingKeys.latitude]
                longitude: row[Location.CodingKeys.longitude]
            )
        }
    
        var persistenDictionary: [String: DatabaseValueConvertible?] {
            // Compiler errors: CodingKeys is unavailable due to private access level
            return [
                CodingKeys.name.stringValue: name,
                Location.CodingKeys.latitude.stringValue: location.latitude,
                Location.CodingKeys.longitude.stringValue: location.longitude,
            ]
        }
    }
    ```
    
    For the same reason, requests won't be able to use coding keys either:
    
    ```swift
    // Nope
    POI.filter(.name == "Paris")
    ```

- Codable records will be *much slower* to decode. In order to support [SE-0166](https://github.com/apple/swift-evolution/blob/master/proposals/0166-swift-archival-serialization.md), GRDB has to perform many runtime checks, and can't profit from raw direct access to database rows (one of the secrets of its [speed](https://github.com/groue/GRDB.swift/wiki/Performance)):

    ```swift
    struct POI: RowConvertible, Persistable, Decodable {
        let name: String
        let location: Location
        
        private enum CodingKeys : CodingKey {
            case name
            case latitude
            case longitude
        }
        
        init(row: Row) {
            // Fast sqlite3 access
            name = row["name"]
            location = Location(
                latitude: row["latitude"]
                longitude: row["longitude"]
            )
        }
        
        public init(from decoder: Decoder) throws {
            // Slow runtime checks and value extraction:
            let container = try decoder.container(keyedBy: CodingKeys.self)
            name = try container.decode(String.self, forKey: .name)
            location = try Location(
                latitude: container.decode(Double.self, forKey: .latitude)
                longitude: container.decode(Double.self, forKey: .longitude))
        }
    }
    ```
    
    This means that some user will have to do without SE-0166 services.

hartbit:
> SE-0166 generates code where the coding keys are private, which means that custom row decoding/encoding can't use them:

Wow. Was not aware of that. Such a pity.

> Codable records will be much slower to decode

All very good reasons. Thanks for the explanation.

swiftlyfalling:
I'd like to put my vote in for `FetchedRecordsController` and Transaction Observers. Both have been incredibly useful.

groue:
All right, @swiftlyfalling! I have a tiny issue with the current state of transaction observers, though: once they declare interest in some kind of database events, they get notified of all of them. When the observer is only interested in the first one, notification of subsequent events is somewhat useless.

swiftlyfalling:
@groue: That's been the main way I've used them (notifying of all events that meet criteria, not just the first one). Couldn't someone waiting for the first event just call `remove(transactionObserver:)` after they receive it?

If you think this is a common use-case, maybe a block-based convenience wrapper could handle that?

So you could use something like the following if you wanted a notification for the first matching event that was committed:
```swift
// func notifyAfter(firstMatchingEvent: (DatabaseEventKind) -> Bool, block: (Database, DatabaseEvent) -> Void)

dbQueue.notifyAfter(firstMatchingEvent: { $0.tableName == "persons" }) { db, event in
    // block is called with a copy of the event from databaseDidChange
    // but only if committed (i.e. from the underlying databaseDidCommit)
}
```

Behind the scenes, it would take care of registering a TransactionObserver and removing it after the first matching *committed* event came through. Just an initial thought.

groue:
@swiftlyfalling I didn't make myself clear. I wasn't talking about transaction observers that want to unregister after an event. Those can indeed, as you say, call `remove(transactionObserver:)` whenever they want.

No, my trouble is for transaction observers that set a flag in their `databaseDidChange(with:)` method, waiting for `databaseDidCommit(_:)` to process this flag. For example:

```swift
class TableObserver : TransactionObserver {
    let tableName: String
    private var modified: Bool
    
    func observes(eventsOfKind eventKind: DatabaseEventKind) -> Bool {
        return eventKind.tableName == tableName
    }
    
    func databaseDidChange(with event: DatabaseEvent) {
        modified = true
    }
    
    func databaseWillCommit() throws { }
    
    func databaseDidCommit(_ db: Database) {
        if modified {
            print("Table \(tableName) was changed in \(db)")
        }
        modified = false
    }
    
    func databaseDidRollback(_ db: Database) {
        modified = false
    }
}
```

Many transaction observers (starting from FetchedRecordsController and RxGRDB) follow this pattern. They are not interested in the detailed events that are provided to `databaseDidChange(with:)`. They just care that *some* change has been committed. When a transaction contains many inserts, updates, or deletes, the `databaseDidChange(with:)` method is called many many times, repeatedly setting the flag to true, before `databaseDidCommit` is eventually called, and the flag handled.

All those useless calls slightly bother me. But I should do some benchmarks before I can positively claim they are a real problem.

In a related fashion, GRDB prevents the [truncate optimization](https://www.sqlite.org/lang_delete.html#truncateopt) so that transaction observers can observe individual deletions. This optimization enhances the `DELETE FROM table` request by dropping the table content as a whole, instead of deleting each individual row. This is another reason to claim that TransactionObserver has not quite reached its final state.

Both those "issues" with TransactionObserver could be fixed before GRDB 1.0, or not. If not, TransactionObserver could enter GRDB 1.0 in the *experimental* state (which does not mean it would be removed - just that it could be enhanced without bumping the major version of the library).

I only know three users of the raw TransactionObserver: me, you (so that you can use [SQLITE_ENABLE_PREUPDATE_HOOK](https://github.com/groue/GRDB.swift#support-for-sqlite-pre-update-hooks)) and @pierlo. Maybe you both would be dissapointed if TransactionObserver were not made stable for GRDB 1.0. If so, well I'll have to address my above questions, or lock TransactionObserver until GRDB 2.0.

swiftlyfalling:
@groue: Aaaah - gotcha.

Interesting. I would be very curious to see what sort of real-world performance impact all those calls have. I'm not certain it's a huge impact, but it's certainly worth investigating.

Re: disabling the truncate optimization: Slightly restructuring things could allow TransactionObservers to opt-out of the truncate optimization for specific tables (if they need per-item deletes from those tables)... or GRDB could just require that TransactionObservers properly handle the case of full table delete (which I suppose would require modifications to `DatabaseEvent` to support a lack of a rowID for that case?). Will ponder this some more.

groue:
OK guys, I think GRDB 1.0 is ready to ship!

Check out the [changelog](https://github.com/groue/GRDB.swift/blob/GRDB1/CHANGELOG.md).

The list of [experimental features](https://github.com/groue/GRDB.swift/tree/GRDB1#what-are-experimental-features), those that won't follow semantic versioning, is limited to extra-advanced features:

- query interface extensions (when one defines new types of SQL expressions for example)
- custom row adapters (when one needs more than the [built-in adapters](https://github.com/groue/GRDB.swift/tree/GRDB1#row-adapters))
- [adapted requests](https://github.com/groue/GRDB.swift/tree/GRDB1#building-custom-requests), because I'm not confident yet about adapters stacks (but chances are nobody will use those features, but the future [associations API](https://github.com/groue/GRDB.swift/issues/176)).

All other APIs are stable now :smile:

hartbit:
Congratulations! 👏

groue:
Oh, and GRDB 1.0 will require Swift 3.1 and Xcode 8.3.3. A version 0.111.0, with exactly the same APIs, will be the last version with guaranteed Swift 3.0 and Xcode 8.1 support.

swiftlyfalling:
Woohoo! 👍

The [After-Commit Hook](https://github.com/groue/GRDB.swift/tree/GRDB1#after-commit-hook) and [Observation Extents](https://github.com/groue/GRDB.swift/tree/GRDB1#observation-extent) are **_fantastic_**! 🎉 Very elegant.

groue:
Many thanks @hartbit,  @pierlo, @swiftlyfalling, for your challenging questions, inspiration, skills, and support! 🤗

PR #230, opened 2017-06-06
Fix compilation with Xcode 9 in compatibilty mode

hartbit:
I think this is a swift compiler regression. I'm fairly sure that GRDB should continue to build without errors in `-swift-version 3` mode. But as GRDB is on the source compatibility suite, not sure if we need to let them know.

### DIFF
--- a/GRDB/Core/Database.swift
+++ b/GRDB/Core/Database.swift
@@ -596,46 +596,73 @@ public final class Database {
     }
     
     private static func close(connection sqliteConnection: SQLiteConnection) {
-        // sqlite3_close_v2 was added in SQLite 3.7.14 http://www.sqlite.org/changes.html#version_3_7_14
-        // It is available from iOS 8.2 and OS X 10.10 https://github.com/yapstudios/YapDatabase/wiki/SQLite-version-(bundled-with-OS)
-        if sqlite3_libversion_number() >= 3007014 {
-            // https://www.sqlite.org/c3ref/close.html
-            // > If sqlite3_close_v2() is called with unfinalized prepared
-            // > statements and/or unfinished sqlite3_backups, then the database
-            // > connection becomes an unusable "zombie" which will automatically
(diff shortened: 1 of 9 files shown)


groue:
Thanks @hartbit!

> I think this is a swift compiler regression.

I do too.

This PR is likely to break Travis builds, which build GRDB against Xcode 8.3 and Xcode 8.1. If it does, maybe try this:

```swift
#if swift(>=3.2)
    getBytes(buffer.baseAddress!)
#else
    getBytes(buffer.baseAddress)
#endif
```

swiftlyfalling:
@groue, @hartbit: Yeah, you may need to do the `#if swift(>=3.2)` bit. We had to make similar little tweaks to [ProcedureKit](https://github.com/ProcedureKit/ProcedureKit). I think the toolchain that shipped with Xcode 9 beta 1 may differ slightly from the Swift 4.0 Development toolchains.

hartbit:
@groue I talked to a Foundation engineer in WWDC labs yesterday and that specific API change is intentional for security reasons. So I closed the bug report.

groue:
Travis is unhappy, but this is unrelated to your change:

```
xcodebuild: error: Unable to find a destination matching the provided destination specifier:
		{ platform:iOS Simulator, OS:10.3, name:iPhone 7 }
```

Let me have a look.

groue:
@hartbit, will you please let me push to your `xcode9-fix` branch? I think the setting is in the PR page itself.

hartbit:
@groue Of course! Got ahead.

groue:
Thanks @hartbit!

I did fix a couple other errors and warnings. GRDB still builds from Xcode 8.1 to Xcode 8.3.3, and Xcode9-beta.

I guess Travis will still complain: I'll deal with it after it has completed its jobs.

groue:
Thanks a lot, @hartbit & @swiftlyfalling!

Issue #236, opened 2017-06-30
Support for custom aggregate functions


hashemi:
GRDB currently supports defining custom SQL functions. However, it doesn't support custom aggregate functions. I looked into it and here are my notes.

In the C interface, a custom SQLite aggregate function is defined as a pair of functions: `step` and `final`. Step is called for each row and does not return a value but instead keeps track of the data somehow. Final is called at the end and returns the actual result of the aggregate.

For example, in an `average` aggregate function, when `step` is called and is passed a value from a row, it adds the value to a sum and increments a count variable. When `final` is called, it divides the sum by the count and returns an average as an SQL value.

In PHP, the user implements a couple of functions for step and finalize. The first argument the functions is passed by reference with each call and is used to keep track of state between calls to `step` and the last call to `final`:

```php
function max_len_step(&$context, $string) {
    if (strlen($string) > $context) {
        $context = strlen($string);
    }
}

function max_len_finalize(&$context) {
    return $context;
}
```

In Python, it's done by writing a class that has `step` and `finalize`. `step` takes a single argument (database value) and does not return. `finalize` does not take any arguments and returns the final result. State is tracked in instance variables of the class:

```python
import sqlite3

class MySum:
    def __init__(self):
        self.count = 0

    def step(self, value):
        self.count += value

    def finalize(self):
        return self.count
```

I think we could use a technique similar to either in Swift:

1. We can have a struct or class that takes a pair of functions or closures. Each function accepts a `context` of a generic type as an `inout`. The step function is called with this `context` as well as a database value per row. The `final` function is called with the `context` only and is expected to return a database value. The GRDB API would expect an *instance* of this struct (or class) type.

Here's an example of what the interface looks like:

```swift
typealias DatabaseValue = Int

// Interface provided by GRDB
struct Aggregate1<Context> {
    let step: (inout Context?, DatabaseValue) -> ()
    let final: (inout Context?) -> (DatabaseValue)
}

// Simulates how the functions are called by GRDB/SQLite
extension Aggregate1 {
    func aggregate(_ rows: [DatabaseValue]) -> DatabaseValue {
        var context: Context? = nil
        for row in rows {
            step(&context, row)
        }
        return final(&context)
    }
}

// What the user implements
func step1(context: inout Int?, value: DatabaseValue) {
    if context == nil {
        context = value
    } else {
        context = context! + value
    }
}

func final1(context: inout Int?) -> DatabaseValue {
    return context! as! DatabaseValue
}

[...]

PR #237, opened 2017-07-01
Support for custom aggregate functions

groue:
As requested by @hashemi in #236, here is the support for custom aggregate functions:

```swift
struct MySum : DatabaseAggregate {
    var sum: Int = 0
    
    mutating func step(_ dbValues: [DatabaseValue]) {
        if let int = Int.fromDatabaseValue(dbValues[0]) {
            sum += int
        }
    }
    
    func finalize() -> DatabaseValueConvertible? {
        return sum
    }
}

let dbQueue = DatabaseQueue()
let fn = DatabaseFunction("mysum", argumentCount: 1, aggregate: MySum.self)
dbQueue.add(function: fn)
    try db.execute("CREATE TABLE test(i)")
    try db.execute("INSERT INTO test(i) VALUES (1)")
    try db.execute("INSERT INTO test(i) VALUES (2)")
    try Int.fetchOne(db, "SELECT mysum(i) FROM test")! // 3
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,36 @@ Release Notes
 
 ## Next Version
 
+**New**
+
+- `DatabaseAggregate` is the protocol for custom aggregate functions (fixes [#236](https://github.com/groue/GRDB.swift/issues/236), [documentation](https://github.com/groue/GRDB.swift#custom-aggregates)):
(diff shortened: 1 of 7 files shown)


groue:
Shipped in v1.1

PR #246, opened 2017-08-05
Fix typo

vs-savchenko:


groue:
Thanks but I think you forgot both the subject and the body of your message, and just sent an attachment. Try again when you have time!

swiftlyfalling:
@vs-savchenko: For `poiCount` & `poiTitles`, the "poi" is, I believe, short for "**p**oint **o**f **i**nterest", not an accidentally truncated version of "point".

groue:
We're not in the mind of @vs-savchenko so we may never know what he/she meant (and that's why I closed the issue as invalid). I could rename PointOfInterest to Place, if it could improve legibility. What I care about in this example is coordinates: I want to make it clear that not only structs made of ints and strings are supported. CLLocationCoordinate2D was the perfect well-known and ready-made compound type I needed, hence "point of interest".

Issue #283, opened 2017-12-06
[logging] misuse at line 147939 of [2b0954060f] when opening a database on iOS 11


gereons:
## What did you do?

    var config = Configuration()
    config.readonly = true
    let db = try DatabaseQueue(path: dbFile, configuration: config)

## What did you expect to happen?

No log output

## What happened instead?

Each time I run this on iOS 11, I get a log line like 
`2017-12-06 10:35:24.210182+0100 MyApp[33044:6574873] [logging] misuse at line 147939 of [2b0954060f]`

Running the same code, using the exact same database file (it's generated offline and installed as part of the app bundle) on iOS 10 is silent, as expected.

This happens both on the Simulator and on real devices.

## Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 2.4.0
**Installation method:** CocoaPods
**Xcode version:** 9.1
**Swift version:** 4.0
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 10.13

groue:
Hello @gereons. I don't know who is logging this line, and what it means. Did you find any evidence that this line is the sign of an actual problem that needs fixing inside GRDB?

gereons:
It's definitely from within SQLite or GRDB, in `Database.setupGlobalErrorLog()` where the `Impl` object is initialized.

So far I have found no indication that anything else is broken. Other mentions of this error (e.g. https://stackoverflow.com/questions/33239845/sqlite-error-sqlite-misuse-error-ios-add-table) indicate that it could be improper API usage that is now detected by the iOS 11 version of SQLite.

It may well be harmless and unavoidable, and that may warrant a mention in the README.

groue:
> It's definitely from within SQLite or GRDB, in Database.setupGlobalErrorLog() where the Impl object is initialized.

You say so. Were you able to use the debugger and trace the GRDB line that triggers the log? Which line is it? Share full information, please.

>  Other mentions of this error

I don't see how this stack overflow question is related to the`[logging] misuse at line 147939 of [2b0954060f]` line we're talking about : this SO question is about an SQLITE_MISUSE error that the user is getting. Since GRDB does not swallow any SQLITE_MISUSE error, and throws all of them back to the host application, this question would be relevant to this issue if and only if your application were indeed catching a SQLITE_MISUSE error. This is not at all the situation described in this issue.

groue:
> It may well be harmless and unavoidable, and that may warrant a mention in the README.

[...]

PR #284, opened 2017-12-07
Don't call sqlite3_config() unless explicitly required to do so

groue:
**Context**

GRDB calls the SQLite [sqlite3_config](https://sqlite.org/c3ref/config.html) function in order to configure the global [error log](https://github.com/groue/GRDB.swift/blob/master/README.md#error-log).

Any call to sqlite3_config must happen *before* any connection to any database. In other words, changing the SQLite configuration after a connection has been opened is a *misuse* of SQLite .

**Bug**

GRDB was calling sqlite3_config() once, before opening its first database connection. But issue #283 has revealed that this GRDB practice could conflict with other pieces of code that also open database connections. When GRDB opens its first connection *after* another part of the program has already opened a connection, this sqlite3_config() invocation is a misuse.

**Fix**

This PR fixes this, by only calling sqlite3_config() when the GRDB user configures the error log:

```swift
// Invokes sqlite3_config
Database.logError = { (resultCode, message) in
    NSLog("%@", "SQLite error \(resultCode): \(message)")
}
```

This also mean that the user is now responsible for setting the `Database.logError` *before* any other piece of code would open a connection.

### DIFF
--- a/GRDB/Core/Database.swift
+++ b/GRDB/Core/Database.swift
@@ -46,7 +46,20 @@ public final class Database {
     /// > advantage of the error logging facility of SQLite in their products,
     /// > as it is very low CPU and memory cost but can be a huge aid
     /// > for debugging.
-    public static var logError: LogErrorFunction? = nil
+    public static var logError: LogErrorFunction? = nil {
+        didSet {
+            if let logError = logError {
+                registerErrorLogCallback { (_, code, message) in
+                    guard let logError = Database.logError else { return }
+                    guard let message = message.map({ String(cString: $0) }) else { return }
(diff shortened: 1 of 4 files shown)


groue:
@gereons, I expect this PR to fix #283. Will you eventually give it a try and confirm that the warning does no longer appear?

gereons:
Will do, sure!

gereons:
LGTM. Log message is gone, and my app continues to work as intended, I haven't noticed anything unusual.

groue:
OK @gereons, thanks for your confirmation! I'll see why some test fail on travis, and ship a new version soon after.

PR #313, opened 2018-02-27
GRDB 3.0

groue:
Work has started on GRDB 3.0.

Improvements & breaking changes:

- [X] #314 The Record protocols have been renamed: `RowConvertible` to `FetchableRecord`, `Persistable` to `PersistableRecord`, and `TableMapping` to `TableRecord`
- [x] #319 Associations & Joins
- [X] #321 Improve database testability
- [X] ~#311~ ~#328~ #348 Support for RxGRDB
- [X] 3160b65 Upgrade custom SQLite builds to to 3.22.0
- [x] #332 Implicit transaction in DatabasePool.write and DatabaseQueue.write
- [x] #336 Enhancements to logical operators
- [x] #337 Foster auto-incremented primary keys
- [x] #340 ColumnExpression Protocol
- [x] #342 Reversing unordered requests has no effect
- [x] #347 Common API for requests and associations derivation
- [x] #334 vsscanf() based date string parser by @sobri909 
- [x] #354 Preliminary Linux support for the main framework by @fpillet 
- [x] #355 Automatic table name generation
- [x] #365 Delayed Request Ordering

Minor or under-the-hood improvements:

- [x] #312 The inner "Cursor library" had an `IteratorCursor` type that would turn any iterator into a cursor. This type has no purpose: one should use `AnyCursor(iterator:)` instead.
- [X] #329 Request derivation protocols
- [x] #330 Expressions are no longer PATs
- [x] #331 Improve Row descriptions
- [x] #335 Row scopes collection, breadth-first scope search

Companion libraries

- [x] https://github.com/RxSwiftCommunity/RxGRDB/pull/32
- [ ] [GRDBObjc](https://github.com/groue/GRDBObjc)

Documentation

- [x] README.md
- [x] CHANGELOG.md
- [x] Inline documentation
- [x] [Migration guide from GRDB 2](https://github.com/groue/GRDB.swift/blob/GRDB3/Documentation/GRDB2MigrationGuide.md)
- [ ] [Medium articles](https://medium.com/@gwendal.roue/)
- [x] [GRDBDemo](https://github.com/groue/GRDBDemo)
- [ ] [WWDCCompanion](http://github.com/groue/WWDCCompanion)


fpillet:
I'm actually porting a codebase from Realm to using an SQLite database with GRDB -- working right on the GRDB3 branch although I'm doing simple stuff at this stage (writes, reads, fetchall queries)

sobri909:
I could be up for a bit of testing!

What sort of timeline are you looking at before shipping a stable? I'm probably about 2-4 weeks away from shipping a [refactor of LocoKit's stores/managers](https://github.com/sobri909/LocoKit/issues/11), so it might line up nicely.

**Aside:** I'm mulling over some questions for you about `FetchedRecordsController`, but I've got a bit more digging and experimenting to do before asking, so I'll come back to that later :)

hartbit:
I've been using this branch while developing a new app and making extensive use of the association APIs. While I haven't had much time to think about or provide feedback on the design of the APIs,but everything has been working fine for me.

groue:
Thanks folks!

@fpillet: happy everything works :-) Maybe the migration guide will have you want to push it further :-)

@sobri909: the stable version should ship in June. In the next week, you'll be able to use something very close. Yet, I have a lot of demo apps, documentation, and articles about GRDB to update before I can ship the stable one. Also, writing doc often has me tweak APIs until the doc feels right: there will be little changes during this phase. About FetchedRecordsController: just open an issue if you want to discuss. This beast remains the best one at smooth table view animations, but it hasn't received much love since RxGRDB has been introduced.

@hartbit: cheers! We still have time for the feedback, if you have any.

freak4pc:
I could possibly check on our codebase - but I have some custom stuff around migration etc so not sure it would be super straightforward to test the branch.

groue:
@freak4pc: no worries, I'm sure you'll come back soon 😉

groue:
Hello folks,

The pre-release of GRDB 3 is there for you. Follow the [GRDB 2 Migration Guide](https://github.com/groue/GRDB.swift/blob/GRDB3/Documentation/GRDB2MigrationGuide.md), that should help you jump in the new version, or install directly the `GRDB3` branches of GRDB and RxGRDB. Associations are covered in their dedicated [Associations Guide](https://github.com/groue/GRDB.swift/blob/GRDB3/Documentation/AssociationsBasics.md). If you don't say anything, I'll suppose everything is OK!

fpillet:
I went through the Associations guide and it's wonderful. Certainly something I'm going to put to good use soon! Congrats 👍 for all this hard work Gwendal, it's incredible.

freak4pc:
@groue Plan on trying GRDB 3 today, I expect it would be a bit tricky due to some of the custom code I added around manual associations, but maybe I can finally get rid of some of that :) 

While browsing through the guide it seems the associated object must always be an Optional, is that a limitation or just a selection of a specific use case?

groue:
@fpillet: Thanks Florent! I'm pretty excited that you and other users can check if this hard work can be put to some good use ;-)

@freak4pc: Welcome on board!

> Plan on trying GRDB 3 today, I expect it would be a bit tricky due to some of the custom code I added around manual associations, but maybe I can finally get rid of some of that :)

Yes, there has been yet another rewrite of the association layer since our last conversation about it. In its current shape, I can finally understand it throughout, and so it *will ship*!

Yet, I know that the root `Association` protocol will be tweaked as we add more features (especially "through" associations, that jump over intermediate tables): consider it to be still pretty radioactive.

It's always a good idea to drop a word when you feel like you are a power user and you're pushing GRDB limits: it helps confirming that an experimental and unstable API delivers its promise (conquer the whole world, of course), and can become stable and locked behind semantic versioning. I should create an "experiential" label in [issues](http://github.com/groue/GRDB.swift/issues), to help us grow our common knowledge, confidence, and help :construction: :gift:

> While browsing through the guide it seems the associated object must always be an Optional, is that a limitation or just a selection of a specific use case?

No, it does not have to be optional.

It must only be optional when it may be missing (optional vs required association request). Compare:

```swift
// Required associated object
struct BookInfo: Codable, FetchableRecord {
    var book: Book
    var author: Author
}
let request = Book.including(required: Book.author)
let bookInfos = try BookInfo.fetchAll(db, request)

// Optional associated object
struct BookInfo: Codable, FetchableRecord {
    var book: Book
    var author: Author?
}
let request = Book.including(optional: Book.author)
let bookInfos = try BookInfo.fetchAll(db, request)
```

There are two working hypotheses at play here (I'm 100% sure of none of them, but I can express them):

- It's OK to define nearly as many little codable structs as there are actual associated requests in an app, because those ad-hoc structs should usually be defined as *private* types nested in a controller/viewModel. This explains why I always name them `XXXInfo` in the doc: their meaning is intended to be *local*. Why so? Because global "god objects" that hide the raw database behind high-level methods or observables are, in all rigor, unable to deal well with [data consistency](https://github.com/groue/RxGRDB/blob/GRDB3/README.md#data-consistency) (an unfortunate but unavoidable consequence of the lack of management brought by Realm or Core Data). I expect that users experiment with god database objects because that's what they've always dealt with, until they eventually simplify their code and just fetch what they need when they need. Yes, it's a rebutal of the [Clean Architecture](https://8thlight.com/blog/uncle-bob/2012/08/13/the-clean-architecture.html), which can only work with managed data (Realm or Core Data).
- It's OK if the eventual NOT NULL constraint of the book.authorId database column is not expressed right into the association, and if the developer has thus to always choose whether she needs to perform an "optional" or "required" join. I experimented with embedding the required/optional flag into the association itself, but I eventually learned that even not null associations need to be optionally loaded, because of filters.

Sorry, I can never stop when I'm asked a question ;-)

freak4pc:
Never ever apologize, I always enjoy reading your well-thought replies. Thanks 😻 :D

fpillet:
@groue while we're on the topic of limits, is there a way to pass contextual data (i.e. a UserInfo of sorts) for objects that need it to complete initialization ?

In my datamodel I'm doing things where initializing many objects require being passed some external additional info provider, and I end up writing a custom `fetchAll` method to manually pass it at init time.

Any thoughts on this ?

groue:
@fpillet:

> @groue while we're on the topic of limits, is there a way to pass contextual data (i.e. a UserInfo of sorts) for objects that need it to complete initialization ?
> 
> In my datamodel I'm doing things where initializing many objects require being passed some external additional info provider, and I end up writing a custom `fetchAll` method to manually pass it at init time.
> 
> Any thoughts on this ?

Of course :smile:

This is a use case that is not very well supported. See #324 for a variant. I'll explain below why this is not supported yet, and what are your current options.

As you see, the `ìnit(row:)` gets no context, and can't fail with nil or an error.

That's tough! Again, there are working hypothesis at play there, but also room for present and future optimizations.

---

The hypothesis, as I like to put it, is that a local database is not some JSON loaded from a remote server. This allows GRDB to focus on trusted and controlled databases. If you can't decode a row into a record, this is because the database contains garbage (and you're warmly invited to fix it instead of dragging garbage behind you), or because your type is *not a record*.

Some types are not records, just as not all model objects can be decoded from raw JSON. Sometimes app models don't match their storage or network representations. I'll even add *former versions* of the app models to the picture, since they sometimes pop out when you have to write non trivial migrations from earlier versions of the database.

Of course, all those layers often match pretty closely. GRDB, just like Swift Codable, high-level Swift JSON libs, etc. blur the frontier as they provide handy and concise APIs that let one throw a few lines of code and start working. Open-source competitive pressure at work :sweat_smile:

But, hey, trivial code eventually can only deal with trivial apps :grimacing: I sometimes have an empathy sweat when I witness, in the [Swift forums](https://forums.swift.org), the pressure on the golden-hearted @itaiferber so that he adds more convenience APIs around Codable.

---

The plain `ìnit(row:)` allows optimizations, and especially late decoding of records.

For example, FetchedRecordsController decodes records on demand: `playerController.record(at: IndexPath(...))`. OK, I admit this does not go very far yet.

But I sometimes envision that we can deal with thousands and thousands of records, in a thread-safe way, without consuming much memory, with requests results cached on disk. I also see [snapshots](https://github.com/groue/GRDB.swift#database-snapshots) as an early stone in this high wall.

Now you know what's behind `init(row:)`.

---

> In my datamodel I'm doing things where initializing many objects require being passed some external additional info provider, and I end up writing a custom `fetchAll` method to manually pass it at init time.

Yes, the custom fetchig method is required today. Maybe you write extensions that wrap something along:

```swift
struct Player: TableRecord { ... } // No FetchableRecord adoption
let request = Player.all()
let context = Context(...)
let players = try Row.fetchAll(db, request).map {
    Player(row: $0, context: context)
}
```

I can provide one advice here: root your extensions on [cursors](https://github.com/groue/GRDB.swift#cursors). Avoid arrays until as late as possible, because arrays have several drawbacks: they are not lazy, they consume memory, and they have lost touch with low level C SQLite functions, which means that they involve slow conversions from copies of database values.

You know the three fetching methods, `fetchCursor`, `fetchAll`, and `fetchOne`. Follow their example. As soon as you need a customized `fetchAll`, it is recommended that you build it on top of a customized `fetchCursor`. For example:

```swift
extension Player {
    /// A cursor over all players
    static func fetchCursor(_ db: Database, context: Context) throws -> MapCursor<RowCursor, Player> {
        return try Row
            .fetchCursor(db, Player.all())
            .map { Player(row: $0, context: context) }
    }
    
    /// All players
    static func fetchAll(_ db: Database, context: Context) throws -> [Player] {
        return try Array(fetchCursor(db, context: context))
    }
    
    /// The first player
    static func fetchOne(_ db: Database, context: Context) throws -> Player? {
        return try fetchCursor(db, context: context).next()
    }
}
```

---

Now I'd also like to stress an information that may drive your other questions and suggestions. The doc is full of requests of raw rows, values, and records, because they are the most common requests. But there can be requests of *anything*. There is no constraint on the type associated with requests. It just happens that the only requests that have ready-made fetching methods are requests of rows, values, and records.

Nothing prevents you from defining a whole new realm of requests. Let's say, requests of `ContextFetchableRecord`:

```swift
protocol ContextFetchableRecord {
    associatedtype Context
    init(row: Row, context: Context)
}
```

You'd then write the whole stack of convenience methods around ContextFetchableRecord, just like GRDB ships with a built-in stack around FetchableRecord.

```swift
struct Player: TableRecord, ContextFetchableRecord {
    static let databaseTableName = "players"
    init(row: Row, context: Foo) { ... }
}

let foo = Foo(...)
let players = try Player.fetchAll(db, "SELECT ...", context: foo)
let players = Player.filter(...).fetchAll(db, context: foo)
// etc
```

As a starting point, check out this [gist](https://gist.github.com/groue/f7634ffc9829d472867c65e8a01c2e8f). That's much more involved, but very reusable. If it can fit your domain, this is an idea worth exploring.

For the rest, I wait for your ideas and suggestions!

groue:
> The only way to to let GRDB pass this information without changing the API would be to propagate it via `DatabaseQueue`, passing a generic context in `Configuration`, then having the query interface propagate it to `Row`.

After I sent the previous message, I wandered around database coloring too.

Deferred decoding of records (that I wish to keep) means that records can be decoded long after touch has been lost with the database and its context. That would imply that rows themselves embed the context. No problem with that: `Row` is already a pretty fancy type.

I just wonder about the consequences of having DatabaseQueue, DatabasePool, and Row becoming generic. And when protocols get associated types... Things generally become less easy for library users.

There's a big exploration to do here!

> User would lose some of the type safety, having to cast to the appropriate contextual type at effective class instantiation time though.

A real possibility, and a shared hesitation ;-)

> For now I'm gonna give a go to your ContextFetchableRecord technique.

I hope I did not send you in a dead-end, or a time pit 🤞

hartbit:
I'm wondering how you suggest we consume joined requests containing hasMany associations? For example, lets imagine the `Book` example from the documentation actually has many `authors`. How could we consume this joined request?

```swift
struct BookInfo: FetchableRecord, Decodable {
    var book: Book
    var authors: [Author]
}

let request = Book.including(required: Book.authors)
```

groue:
Hello @hartbit,

The `Book.including(required: Book.authors)` request fetches database rows, each one of them made of one book and one author. It can be decoded into a (Book, Author) pair, but not into a (Book, [Author]) pair.

Support for such requests is not included in the initial feature set.

You'll have to do by hand what future versions of Associations API may do for you. Let's see how.

> I'm wondering how you suggest we consume joined requests containing hasMany associations? For example, lets imagine the Book example from the documentation actually has many authors. How could we consume this joined request?

First of all, please let me reframe your question: you say that books have many authors. And the implicit context of the library example adds that authors have many books, too.

This kind of relationship (symmetric hasMany) requires, in the database, a joining table, let's say `authorship`, that links to both books and authors.

Since I'm not sure that this is the actual context of your question, I will answer another one, hoping this is what you intended. consider the initial library: books belong to their authors, and authors have many books. Two tables. Books have a foreign key that links to authors: "authorId". You want to load all authors with their books, in an array of (Author, [Book]) pairs (or an equivalent struct).

This can not be efficiently done in a single database query, and this is the reason why it isn't included in the initial Association API. We'll have to wait for future versions.

You need two queries:

1. load authors
2. load books of those authors

It should give something along;

```swift
struct AuthorInfo {
    var author: Author
    var books: [Book]
}

// Apply any desired filtering and ordering on those base requests:
let authorsRequest = Author.all()
let booksRequest = Book.all()

let authors = try authorsRequest.fetchAll(db)
let authorIds = authors.map { $0.id }
let books = try booksRequest.filter(authorIds.contains(Column("authorId"))).fetchAll(db)
let groupedBooks = Dictionary(grouping: books, by: { $0.authorId }) // fortunately preserves books ordering
let authorInfos = authors.map { author in
    AuthorInfo(author: author, books: groupedBooks[author.id])
}
```

In an extension (pardon my live sketching):

```swift
extension FetchRequest where RowDecoder == Author {
    func fetchAll(_ db: Database, includingBooks booksRequest: QueryInterfaceRequest<Book>) throws -> [AuthorInfo] {
        let authors = try fetchAll(db)
        let authorIds = authors.map { $0.id }
        let books = try booksRequest
            .filter(authorIds.contains(Column("authorId")))
            .fetchAll(db)
        let groupedBooks = Dictionary(
            grouping: books,
            by: { $0.authorId })
        return authors.map { author in
            AuthorInfo(author: author, books: groupedBooks[author.id])
        }
    }
}

let authorInfos: [AuthorInfo] = dbQueue.read { db in
    let authors = Author.order(Column("name"))
    let books = Book.order(Column("publicationDate").desc)
    try authors.fetchAll(db, includingBooks: books)
}
```

You see that... the Association API hasn't been used at all in this example. Sorry 😉 

Of course, feel free to ask another question about symmetric hasMany, if this was really your topic. And I'll have to talk, again, about the future: ideally we'd have "through" associations that hop across intermediate tables, but we haven't them yet. Associations are a wiiiiiide subject!

groue:
I think we're pretty close from a release. If any of you has any trouble or pet peeve, it is time to get vocal!

hartbit:
I'd like to give it a try, but it won't be before next week. Also, I'm currently stuck from using the latest commit on this branch, because the latest commit on RxGRDB/GRDB3 doesn't compile with it. Could you do something about that?

Thanks!

groue:
> Also, I'm currently stuck from using the latest commit on this branch, because the latest commit on RxGRDB/GRDB3 doesn't compile with it. Could you do something about that?

Your initial PR on RxGRDB has been since obsoleted by https://github.com/RxSwiftCommunity/RxGRDB/pull/32. Check out the [Migration guide from GRDB 2](https://github.com/groue/GRDB.swift/blob/GRDB3/Documentation/GRDB2MigrationGuide.md) where you'll learn about upgrading your Podfile and other interesting changes, and also check the [RxGRDB changelog](https://github.com/groue/RxGRDB/blob/GRDB3/CHANGELOG.md).

fpillet:
Fantastic!

Now ... I just tried compiling on Xcode 10 with macOS 10.13.5 and ... it now needs CSQLite too because the logging function has disappeared from SQLite header :(

groue:
Thanks Florent :-)

- GRDB 3 is stable software, not beta software: it targets Xcode 9.4.
- It did check Xcode 10, of course 😉 There is some known trouble with GRDBCustom and GRDBCipher: see #367. As far as I know, the main framework GRDB has no trouble with Xcode 10. But maybe I'm wrong? If you have any trouble, and need help, please open a new issue, and provide detail, as usual!

fpillet:
Hrm ok so I need to check with all the combinations of toolchains + IDE

Have you tried compiling on macOS 10.13.5 ?

groue:
> Have you tried compiling on macOS 10.13.5 ?

No, not yet.

fpillet:
That's what I'm using, and there it's failing to locate `registerErrorLogCallback`. macOS 10.13.5, Xcode 9.4 with the standard Xcode 9.4 toolchain. So unless an SDK installed by Xcode 10 is causing this trouble, this may be worth a look.

Importing CSQLite in `Database.swift` fixes the issue.

PR #319, opened 2018-03-11
Associations & Joins

groue:
This pull request enhances the query interface so that it can build and consume some joined requests. There is much room for future enhancements, but we have a shippable API already.

Documentation: [GRDB Associations](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md)

@hartbit, @smhk, @Sroka,  @cfilipov, @hdlj, @mtissington, @pakko972, @palpatim, @pierlo, @schveiguy, and @swiftlyfalling, you have shown your interest in GRDB in general and sometimes on join requests in particular: I'd be glad if you would have a look at the proposed documentation. I'll be happy answering your questions, addressing your concerns, and generally improve this PR with your help.


hartbit:
This looks amazing! 🤩 Is there any chance this API won’t make it in GRDB 3? What is the big-picture ETA for GRDB 3? The best way for me to start playing with this now is to integrate it into the application we’re working on. But I can only do so if I won’t have to completely rip the join code out later on and if GRDB 3 is released production-read-ish before we release.

groue:
> This looks amazing! 🤩 Is there any chance this API won’t make it in GRDB 3?

Thanks David! Yes, it *will* ship in GRDB3. I'm now confident enough that the design is not a dead-end. As the benevolent dictator of this lib, and after two failed attempts, I now that this is the way to go.

I have three concerns on the proposed API:

- that it has not been bike-shedded and polished enough.
- that it is too open, and allows writing unsound code that we'll never be able to support and will bite many users
- that the only "warning" you'll find in the [doc](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md) hides a huge conceptual bug.

> What is the big-picture ETA for GRDB 3?

Weeks.

There are unchecked items in TODO.md in the GRDB3 section. Not all of them will ship, because I may change my mind. But none of them is as hard as associations. Yet they all take a little of my free time, and there is also a lot of documentation to update, including the [Medium articles](https://medium.com/@gwendal.roue/).

> The best way for me to start playing with this now is to integrate it into the application we’re working on. But I can only do so if I won’t have to completely rip the join code out later on and if GRDB 3 is released production-read-ish before we release.

I understand. It's difficult to commit to beta software.

A welcome contribution would be a [witty QA mind](https://twitter.com/sempf/status/514473420277694465?lang=en).

hartbit:
> Thanks David! Yes, it will ship in GRDB3.

Great!

> Weeks.

That sounds promising. As we are not planning on releasing before June, I feel confident going forward with integrating those new APIs and beta-testing them at the same time. Worst case scenario, we'll release with Release Candidate level of quality, and I'm confident enough with the level of quality of GRDB to go ahead with that.

I'll try to document my comments here as I use those new APIs.

groue:
@freak4pc, since this PR may also match some of your interests, it may a be a good idea to have your opinion, too?

hartbit:
The APIs are not doing two separate fetches. Where did you get that impression from? The documentation shows the raw sql generated by each API and they all use joins:

https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#joining-methods

freak4pc:
I'll try to clarify my point :) 

This example: 
`let books = try author.fetchAll(db, Author.books)` 

Would fetch all books of a specific author. 

The question is what happens if I want to fetch all authors and their respective books - that is the _join_ piece exactly. Based on the examples I saw in the docs I couldn't find a straightforward example of this.

freak4pc:
Yes but this is exactly the structure I had a problem with - 
Why do I need an intermediary object at all in this case? This isn't a many-to-many relationship, there shouldn't be a "central" object IMO. Unless this is a limitation of GRDB, but a true SQL JOIN can fetch an author with its connected books in its respective objects without any "intermediary" table (or object, in this case)

hartbit:
For me, this fits my mental model of SQL very well: there is an intermediary object because it represents every row of the request result. Its a solution that evolves well whatever the kind of JOIN result you have, even to many-to-many relationships. A solution which would store books in a `books` property on `Author` would like "nice" for one-to-many relationships but would fall short for any more general cases.

freak4pc:
So I guess we are in a disagree about this - even if it fits your mental model of this, this isn't how SQL Joins look like usually. 

A SQL Join would usually provide an object with its foreign-keyed objects, without any intermediary object.

Using an intermediary object is very much like a having SQL View (e.g. a virtual table). So it highly depends on what kind of solution you're trying to architect here, but this isn't exactly a join per-se (structure-wise, even though the underlying is a query)

hartbit:
> So I guess we are in a disagree about this - even if it fits your mental model of this, this isn't how SQL Joins look like usually.

Imagine a more real-world database where books can be authored by multiple authors (many-to-many relationship). In that case, an SQL JOIN like this:

```sql
SELECT authors.*, books.*
FROM authors
INNER JOIN authorBooks ON authorBooks.authorId = authors.id
INNER JOIN books ON books.id = authorBooks.bookId
```

couldn't be represented without an intermediate object.

For an "expert" API like GRDB joins, I think we need to embrace that joins are complicated beasts and provide an API that allows maximum flexibility.

freak4pc:
The question is- are we trying to solve the 10% or the 90% ? Would you agree a double inner-join isn't that common of a use case? 

What I'm asking about is: 
```SQL
SELECT authors.*, books.*
FROM authors
INNER JOIN books ON books.authorId = authors.id
```

E.g. an author and his book

hartbit:
I think its very personal, but my joins are never that simple.

freak4pc:
And that's absolutely fine - but I can honestly tell you the majority of people would want and object, and its directly-relationed objects with it, and not a more complex issue, so I think its definitely possible to model a solution that would allow for the 90% to have a more concise syntax, while the more complex issues can be solved with an intermediary object.

But using an intermediary object for this basic scenario is quite the bummer TBH.

groue:
> Hey @groue , Awesome work!
> 
> I started reading but had to stop to get some work done, here is some of the feedback / questions I had time to go through: 

Hello Shai, thanks for your feedback (thanks David, too)! Let's answer your questions!!

> 
> `let books = try author.fetchAll(db, Author.books)`
> 
> How would you feel abbout
> 
> `let books = try author.books.fetchAll(db)` ? 
> 
> given the `hasMany` type having that method?

You have seen that Author.books is a static propery of Author (it contains an association). There is no `books` property on Author instance that is automatically defined.

I have been wondering if requesting all books from an author had to be done from an author (give me all books given this association), or from the association (give me all books given this author).

Since other ORMs usually load associated records by the mean of methods on the source record, it's been my choice too:

```swift
// Editor's pick: Fetch books from author
let books = author.fetchAll(db, Author.books)

// Discarded: Fetch books from the association:
let books = Author.books.fetchAll(db, from: author)
```

Now fetching methods are not enough. We also need requests, which can feed database observation tools. That's why we also have:

```swift
let request = author.request(Author.books)
request.rx.fetchAll(in: dbQueue).subscribe { books in
    print("new books: \(books)")
}
```

So: `author.fetchAll(db, Author.books)` is a subjective choice to put the origin record on the right of the expression, like other ORMs (but with stranger syntax because there is no automatic generation of properties).

And the documentation on this topic has to be improved, because it mixes requests (advanced feature) and fetching methods (that everybody needs) on the same level.

> ---
> 
> Also, `hasMany` is a method of one of the protocols or a free function? If the latter that might not be the best idea I think ?
> 
> *Edit*: Noticed its a static method on the record, seems fine to me. I'm trying to think if there's a "prettier" syntax but I think this actually looks great.

Well spotted, and thanks. Yes, "hasOne", "hasMany", "belongsTo" have the nice historical log of abstractions that are well understood by developers.
 
> ---
> 
> For this API:
> `Book.including(optional: Book.author)`
> 
> What is the expected output here? I find the name a bit non-descriptive. Would it get a book including its optional author if exists? Would it get a Book that itself includes some author? 

It is described [here](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#includingoptional):

"This request fetches all books, including their author's information. Books that don't have any author (a null authorId column) are fetched as well."

I'm not sure I understand your question, which may mean that I totally failed at explaining the four joining methods, and that you built a mental model of them that is at odds with their actual behavior. No offense, this happens all the time when APIs are not clear enough.

I don't know if would still ask this same question *now*. If you do, would you be able to ask it again, but with different words so that I can figure out what's in your mind, and give a proper answer?


> 
> ---
> 
> Not sure If I missed this but is there a way to fetch an object along with its relation? 
> e.g. for the book example with the `hasMany()` on top being able to do `let authors = try Author.fetchAll(db)` that would also fetch the underlying books connected by relation? Or would that have to still be an imperative/manual operation?

Excellent question. This deserves a new entry in the [Future Directions](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#future-directions) paragraph.

No. Today, there is no API to load all authors will all their books, and you'd have to do it manually. I had the gut feeling that it was not a feature required for an initial release.

> I saw the "BookInfo" structure that seems to try and solve that, but introducing a third data type isn't necessarily what I'd imagine doing (if it's at all possible).

1000 points question again. Not only are those third data types the very reason why we can ship this joining API. But they also are a very recommendable way to consume database values.

The previous attempt at providing an API that generates joins would not use those "third data types". It would return tuples instead. It did address, for example, the "all authors will all their books" request by returning `[(Author, [Book])]`.

But without third data types, we have big issues:

1. Swift has no variadic tuples or generics. This means that we're stuck with hard-coded limits (join 2 types, join 3 types, end.). This just *doesn't scale*.
2. Tuples are handy. But "third data types" are even more handy. Types have an identity. Types can be extended with convenience properties and methods. They can be fed from the database, but also remote JSON data, whatever. They provide *excellent* support for view models.

And "Third-party types" have another benefit (because I have a hidden agenda): by fostering the decoding of joined requests into third-party types, I grant GRDB users with data consistency for free. And the topic of lazy loading of associated records vanishes. What's better than this "BookInfo" type that contains exactly what I need? Why feed it manually with several db requests performed at various times in my app when I can get it for free?

As David said:

> I find this solution fairly elegant. I guess we could have a solution where a tuple is returned. But I actually quite like being able to define my own structured types with the types I expected to get back.

That's exactly my expectation.

You later asked:

> [Shai]: The question is what happens if I want to fetch all authors and their respective books - that is the join piece exactly. Based on the examples I saw in the docs I couldn't find a straightforward example of this.

Unless I'm mistaken, such a requests needs two SQL queries, not a single one (please correct me if I'm wrong):

```sql
SELECT * FROM authors WHERE ...
SELECT * FROM books WHERE authorId IN (...)
```

Such requests can currently not be expressed with the `FetchRequest` protocol (and not be observed by observation tools, btw). FetchRequest has to build a single SQLite statement in its `prepare()` method.

I was expecting that future additions would add another protocol on top of `FetchRequest`, in order to support more general requests. Such general requests could not be iterated with cursors, but they would still be able to feed observation tools.

This makes me think that I may have to rename `FetchRequest` into `StatementRequest`, so that `FetchRequest` can later be used for the more general "fetching" protocol.

You again:

> a true SQL JOIN can fetch an author with its connected books in its respective objects without any "intermediary" table (or object, in this case)

I know that you like model objects to have properties to their associated objects. You call that "modelling", and it's a fair habit that is widely supported by many years of software design, ORM APIs, accross many years and thousands of lines of code:

```swift
class Book {
    var title: String
    var author: Author
}
```

I want to break this habit :metal:

It's OK when your models are managed, uniqued, and lazy-loaded.

GRDB records are much more low-level: a fetched record is an in-memory cache of database values. It can be a plain Swift struct.

That's why we need cache invalidation, and why I put so much energy on providing easy-to-use database observation techniques :-)

This means that the above Book class with its author property, when it exists, *belongs to the application*. The application is sole responsible of its management (which is not easy).

By ditching uniquing, lazy loading, and not managing records, GRDB was able to simplify multi-threaded code. The cost is that records have `authorId` properties instead of `author`. And we have those "third-party types".

I wholeheartedly understand your point. And I jump with passion in the opposite direction, with many many types focused on precise tasks.

My personal experience (which guides me a lot) shows that those types are often private types inside view controllers (or view models or whatever). Add I like to focus on a local and isolated type, instead of bloating a unique model type with convenience methods used by a single screen of an app.


> 
> ---
> 
> Docs-wise I would **bold** terms specific to the GRDB codebase such as "The **BelongsTo** association between a book...", personal taste :)

Very good idea. There are so many new concepts in this doc!

> ---
> 
> I had a problem with the fact tables _must_ be singular, until I got to the last row there and noticed you can actually define a custom name. I'm not sure I'm fully clear on the documentation there, it notes an example regarding "plural" but the written code is singular. What am I missing?

As David has noticed, it's surely a typo. I have tried to be consistent (with the only exception of the "demographics" table which is for me the needle that reminds me that something is not "extra clear" here).

I deeply regret this "singular table" convention. I really want to remove it.

There it lies:

```swift
let author = Book.belongsTo(Author.self)
let request = Book.including(required: author)
let row = try Row.fetchOne(db, request)!
print(row.debugDescription)
// ▿ [id:1, authorId:2, title:"Moby-Dick"]
//   unadapted: [id:1, authorId:2, title:"Moby-Dick", id:2, name:"Herman Melville"]
//   - author: [id:2, name:"Herman Melville"]
```

You see the `author: [id:2, name:"Herman Melville"]` line? We need this "author" key to come from somewhere. And it's better be singular, because it then fits the `author` property of "third-party types" like a glove.

This key comes from the association `Book.belongsTo(Author.self)`.

That association has not much anything to chew on: the Book type (irrelevant), and the Author type, which comes with its `databaseTableName` static property, which contains "author" when the table has a singular name. And that's why singular names are preferred. There is no other reason.

I've tried to build "author" from the Author type itself, by playing with `String(describing: Author.self)`. But it often returns an awful string that requires careful parsing.

To remove the "singular table" convention, we just need to be able to generate "author" from the Author type.


> 
> ---
> 
> The syntax for manually defining a foreign key is a bit verbose/strange IMO :
> 
> ```
> enum ForeignKeys {
>     static let author = ForeignKey(["authorId"]))
>     static let translator = ForeignKey(["translatorId"]))
> }
> ```
> 
> Why use an enum if using static lets?

Right. This "namespace" enum has no purpose, and pollutes the doc.

> Could I use a struct if I'd want to? Why does that "ForeignKey" initializer accept an array?

As David said above, because foreign keys can span several columns.

About the "foreign key" name. It's hard-core, very low-level. I kept it because the two types at both ends of an association use the same foreign keys so disambiguiate their own associations:

```swift
struct Book: TableRecord {
    // Define foreign keys
    static let authorForeignKey = ForeignKey(["authorId"]))
    static let translatorForeignKey = ForeignKey(["translatorId"]))
    
    // Use foreign keys to define associations:
    static let author = belongsTo(Person.self, using: authorForeignKey)
    static let translator = belongsTo(Person.self, using: translatorForeignKey)
}

struct Person: TableRecord {
    static let writtenBooks = hasMany(Book.self, using: Book.authorForeignKey)
    static let translatedBooks = hasMany(Book.self, using: Book.translatorForeignKey)
}
```

I agree that this is clearly not the most elegant part of the API.


-----

Thanks for your feedback so far! I already a nice list of improvements to do, and I hope I could answer your questions and concerns, even if it may look like we disagree sometimes.

groue:
> But using an intermediary object for this basic scenario is quite the bummer TBH.

I've given my point of view above, but yours is still interesting. Do would have an idea of what it would look like (the definition of two book/author types, and the ideal API that loads what you need)?

freak4pc:
> I've given my point of view above, but yours is still interesting. Do would have an idea of what it would look like (the definition of two book/author types, and the ideal API that loads what you need)?

Given your explanation on top I don't think you'd really want to do this :) 

I would imagine for a behavior that causes the "secondary" object to be fetched along with the first without an intermediary object, but you've fully explained your stance on it and I'm fully OK with it. I don't think I have any serious beef with it at this point, it's just a API design choice.

groue:
@freak4pc I'm happy we understand each other. Now if I write bold words and use the 🤘 emoji, this does not mean that I don't want to listen (although heavy rock music can damage ears).

Maybe I can smooth your concerns by telling that supporting classic modeling is just an `init(row:)` away:

```swift
class Book {
    static let author = belongTo(Author.self)
    
    var title: String
    lazy var author: Author { /* fetch the author */ }
    
    init(row: Row) {
        self.title = row["title"]
        if let authorRow = row.scoped(on: "author") {
            self.author = Author(row: authorRow)
        }
    }
}

let books = try Book.including(required: Book.author).fetchAll(db)
for book in books {
    print("\(book.title) by \(book.author.name)")
}
```

I just can't really foster such pattern in the documentation, because lazy loading of associated records breaks the data consistency guarantee (and I morally can't give dangerous advice). It's a little too smart, you see?

freak4pc:
Yeah I'm aware of the lazy method but I can get this done without this PR, so I didn't even mention it :) 

No worries, I totally what your architectural standpoint is and it's 100% good! 💯

groue:
For information, some tests fail because the various tested SQLite versions don't all agree on the `DatabaseRegion` fetched by joined requests. Database region define regions such as "columns a and b from table t". They provide support for transaction observers that want to spot changes in the results of a request (FetchedRecordsController and RxGRDB).

As far as I can see, [tests of database regions](https://github.com/groue/GRDB.swift/blob/91e88fbc41e4485243d2790221be26ea94c5eeef/Tests/GRDBTests/DatabaseRegionTests.swift#L440-L541) have to be adapted, but they don't reveal any observer breakage (although I'll have to add a few integration tests that lift remaining doubts).

groue:
I did start using @freak4pc's feedback:

> How would you feel about
>
> `let books = try author.books.fetchAll(db)` ?

After a good second read, I find your idea stellar.

The [Requesting Associated Records](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#requesting-associated-records) chapter has been rewritten accordingly. `author.fetchAll(db, Author.books)` and similar methods that "fetch the contents of a relation from a source record" have been removed.

> For this API: `Book.including(optional: Book.author)`
>
> What is the expected output here? I find the name a bit non-descriptive.

The [Joining Methods](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#joining-methods) chapter has been fully rewritten, and I hope it's much more clear now.

groue:
> Docs-wise I would bold terms specific to the GRDB codebase such as "The BelongsTo association between a book...", personal taste :)

Good idea, done.

> The syntax for manually defining a foreign key is a bit verbose/strange IMO :
> 
> `enum ForeignKeys {...}`

Yes. I have removed the surprising enum in the [Foreign Keys](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#foreign-keys) chapter.

freak4pc:
> After a good second read, I find your idea stellar.
>
> The Requesting Associated Records chapter has been rewritten accordingly. author.fetchAll(db, Author.books) and similar methods that "fetch the contents of a relation from a source record" have been removed.

Wooo! 🎉 Love this new API ! Looking forward :) Also documentation updates look awesome.

groue:
Thanks Shai! David, the missing [Columns Selected by an Association](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#columns-selected-by-an-association) chapter is now there!

groue:
Two questions by @hartbit, based on https://gist.github.com/hartbit/0acb80e57c3a90a9017b1012ab8349f5

- [Would it be worth erroring out if a select in a joining causes the select to be discarded?](https://twitter.com/dhartbit/status/973916179327868928). Problem: `.joining(required: assoc.select(...))` is wrong because `joining` discards selection. But it currently does it quite silently.

- [I've found it more readable to define all selects together (similarly to a SQL query)](https://twitter.com/dhartbit/status/973917642946904065).

groue:
OK, @hartbit, thanks for your questions.

**You ask to be able to gather all selected values at the front of [your request](https://gist.github.com/hartbit/0acb80e57c3a90a9017b1012ab8349f5)**, instead of sprinkling little selection snippets here and there.

Practically speaking, though, your request has something very specific: it counts and aggregates. Aggregates will be addressed specifically in the future, with something that will likely be called "annotations" (inspiration from [Django ORM](https://docs.djangoproject.com/en/2.0/topics/db/aggregation/#joins-and-aggregates)):

```swift
// SELECT author.*, COUNT(DISTINCT book.id) FROM ... GROUP BY ...
let req = author.annotate(Author.books.count)
```

Annotations will free the user from joining, grouping, and counting the associated records in the correct SQL way.

This will be more welcoming to users who don't know SQL. And even to users who know SQL, because forgetting a DISTINCT is easy.

So, in the initial release, your request won't be addressed. I need to work on aggregations, annotations, merging of association trees, hasOneThrough and hasManyThrough associations first.

**You then ask to error/warn out if a select in a joining causes the select to be discarded**:

Indeed `...joining(assoc.select(...))` is paradoxical because it both ask to discard the selection (joining), and selects values.

But [all associations come with a selection](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#columns-selected-by-an-association), even though it is often implicit. Furthermore, you may have to join an association that you did not define:

```swift
func f(association: Association) {
    let r = request.joining(required: association)
}
```

You don't want this function to have a GRDB-imposed precondition that the input association should not have any explicit selection. You really don't.

So I think of your request as the normal reflex of an early user who hasn't learnt yet the difference between `joining` and `including`, and gets bitten. I think that the proper answer is:

- **Repeated use** (you're unlikely to make the same mistake again, or to remain clueless when you do it again)
- **Documentation**: the [Joining Methods](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#joining-methods) better explains now, I hope, the difference between the four choices.
- **Tooling**: joined request are complex beasts, that's a plain fact, and there's no point pretending it's easy. The trivial zone is very very small. Many mistakes will happen. The doc talks about `Row.debugDescription` for a better understanding of the request results. The doc should give more hints at the ways to read and control the generated SQL. And there should be more high-level APIs (like annotations above).

hartbit:
I've fine with your analysis :) I'm already very happy to have what we have. By the way, I think I understood what you said on Twitter, but could you expand your point of view on making `Column` optionally contain a table name to help disambiguate simple cases without overusing table aliases? I see that as allowing columns to contain an implicit table alias.

hartbit:
Can I propose one extra joining method?

```
// SELECT *
// FROM author
// LEFT JOIN book ON book.authorId = author.id
// WHERE book.id IS NULL
//
// or
//
// SELECT *
// FROM author
// LEFT JOIN book ON book.authorId = author.id
// WHERE NOT EXISTS (
//     SELECT 1
//     FROM books
//     WHERE books.authorId = author.id
// )
Author.excluding(Author.books)
```

I could use it my projects to remove rows which can't benefit from `ON DELETE CASCADE`:

```
// Delete all authors that have no books.
Author.excluding(Author.books).deleteAll(db)
```

groue:
@hartbit TODOified!

groue:
> By the way, I think I understood what you said on [Twitter](https://twitter.com/groue/status/973919600319574016), but could you expand your point of view on making Column optionally contain a table name to help disambiguate simple cases without overusing table aliases? I see that as allowing columns to contain an implicit table alias.

Hello @hartbit!

A little context: in the current proposed API, columns don't know which table they refer to, and one has to use a **table alias** when one wants to juggle with several tables:

```swift
let publishDate = Column("publishDate")
let deathDate = Column("deathDate")
let authorAlias = TableAlias()
let posthumousBooks = Book
    .joining(required: Book.author.aliased(authorAlias))
    .filter(publishDate >= authorAlias[deathDate])
```

Since defining table aliases is not a funny job, you ask if we could avoid them. And why not perform table disambiguation in the columns themselves:

```swift
let bookPublishDate = Column("book", "publishDate")
let authorDeathDate = Column("author", "deathDate")
let posthumousBooks = Book
    .joining(required: Book.author)
    .filter(bookPublishDate >= authorDeathDate)
```

We may someday do with less table aliases, but I know that the enhanced columns you propose aren't the answer.

Sometimes one has to join the same table twice. Not always, of course. But when you do, I want you to be able to rely on a few simple rules, always the same, so that you don't have to learn (and forget) too many specific cases. And more than everything else, I want you to trust the engine. Enhanced columns fail both goals ;-)

1. Enhanced column are not a reliable way to perform disambiguation. The example below would not work:
    
    ```swift
    let bookPublishDate = Column("book", "publishDate")
    let personDeathDate = Column("person", "deathDate")
    let posthumousBooks = Book
        .joining(required: Book.author)      // book.authorId -> person.id
        .including(required: Book.publisher) // book.publisherId -> person.id
        .filter(bookPublishDate >= personDeathDate) // Which deathDate is it?
    ```
    
    It would fail with a runtime error. If not because GRDB would spot the inconsistency, that would be because SQLite could not parse `person.deathDate` IN `SELECT ... FROM book JOIN person person1 ... JOIN person person2 ... WHERE book.publishDate >= person.deathDate`.

2. Because of the previous point, enhanced columns would be ticking bombs waiting to explode. You could not organize your code into logical units. You'd had to write each request from scratch, in isolation, and you could never reuse anything.
    
    Table aliases, on the other side, allow you to write the following code:

    ```swift
    extension Book {
        // The request for posthumous books
        static func posthumous() -> QueryInterfaceRequest<Book> {
            let authorAlias = TableAlias()
            return Book
                .joining(required: Book.author.aliased(authorAlias))
                .filter(Column("publishDate") >= authorAlias[Column("deathDate")])
        }
    }
    
    // Extend posthumous books with the publisher:
    let books = Book
        .posthumous()
        .including(required: Book.publisher)
    ```

So... table aliases are verbose, yes, but they have advantages :-)

hartbit:
It all makes sense now: I understand why they are not the right solution. Thanks for taking the time to explain to slow brain like mine :-)

groue:
Thanks for challenging the proposal, @hartbit 😄

groue:
https://twitter.com/sgrif/status/976263167931109376

**Thank you very much** for giving a look, @sgrif.

> This definitely should have been split into multiple PRs. It's hard to even find the right file to comment on. I'm going to do it here instead (note: I am only going to read through the docs, not the actual implementation or tests)

My fault: my tweet should have linked straight to the doc. This is where I intended to get your advice. I'm glad you picked that way :-)


>> [AssociationsBasics.md#L76](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L76)
>>
>> With GRDB associations, we can streamline these operations (and others), by declaring the connections between books and authors.
>
> is it worth calling out that there's also a difference between N+1 and not here? That seems much more important than "streamlining"

Yes it is, but I have tried to closely follow [Active Record introduction](http://guides.rubyonrails.org/association_basics.html) here. I thought that I could learn from it :-)

>> [AssociationsBasics.md#L87-L90](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L87-L90)
>>
>> ```
>> extension Book {
>>     static let author = belongsTo(Author.self)
>>     var author: QueryInterfaceRequest<Author> {
>>         return request(for: Book.author)
>>     }
>> }
>> ```
>
> I'm not sure if this is just a limitation of Swift or not, but this *really* feels like it's so common it should be one line.

It's a limitation of Swift. We don't have any macro or built-in code generation tool, so everything has to be written down. Now I'm super happy that we only need two declarations. Strictly speaking, the second one is only strongly recommended.

>> [AssociationsBasics.md#L100-L110](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L100-L110)
>>
>> (Magical code that loads (book, author) pairs
>>
>> I'd like to know more details on what's happening here. Is this doing a join? (note: that's typically less efficient than doing two separate queries).

We're still in the introduction, inspired from the Active Record Guide. Such mysteries are disclosed later on (in both guides).

> I'm also curious to know if this struct actually needs to exist, or if one can just have `(Book, Author?)` "just work

That's a very good question. A previous attempt at dealing with association and joins had me look very closely at https://docs.diesel.rs/diesel/associations/index.html. For example, you can read:

> In Diesel, data and its associations are considered to be separate. If you want to pass around a user and all of its posts, that type is `(User, Vec<Post>)`.

I could not agree more. This is how we can deal with users independently from their posts when we need to, without bending the `User` type so that it can accomodate plain users vs. enriched users.

Swift has no variadic templates and tuples which could support an API that could load users, users with their posts, users with their posts and pets, users with their posts and pets and parents etc. This does not scale. Learning to write simple queries would not help writing more complex queries, and one would have to pick a completely different fetching technique.

That's why I kept the idea, but replaced the `(Book, Author?)` tuple with a hand-made `BookInfo` struct. [More complex queries](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#fetching-values-from-associations) just need a bigger struct: the technique scales well with the size of the queries. Finally, you can shape the struct in the way you want (to explain this, we'd need to talk about the consumption of a complex nested hierarchical query into a flat struct - [Decoding a Joined Request with a Decodable Record](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#decoding-a-joined-request-with-a-decodable-record)).

I don't know if Rust can dynamically build variadic tuples. Anyway, Swift can't, so I had to drift a little from Diesel's current way of doing it.


> [AssociationsBasics.md#L261](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L261) Conventions are great, but I sure hope every one of these is configurable
> Oh you literally say that they are. Never mind.

From a maintainer of Active Record, I wouldn't expect less :smile:

>> [AssociationsBasics.md#L274](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L274)
>>
>> **Database table names should be singular and camel-cased.**
>
> https://twitter.com/sgrif/status/976265023138885632

Glad you spotted one! I don't see myself porting [ActiveSupport's inflections](http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html) yet. Am I wrong?

>> [AssociationsBasics.md#L722-L758](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L722-L758)
>>
>> Associations can be combined in order to build more complex requests.
>> You can join several associations in parallel ... You can chain associations in order to jump from a record to another ...
>
> Ok this is the thing I wanted to see. The API you've provided *does not* make it clear to me that a join is occuring (which is important to know if I need the join to be done for filtering purposes).
>
> I'm also curious why you defaulted to joining here, since it's often less efficient than performing two queries. Diesel deliberately gives you the choice between the two (if you want a join, you do so explicitly, otherwise we have helpers for loading associated records)

I did not expect this :-)

Here it is Active Record inspiration that talks:

1. With Active Record, you don't quite tell how stuff should be fetched, and you enjoy consuming the results.
2. Active Records has years of experience picking the best SQL for a request, and has changed its mind several times. Hiding the joining technique allows such change.

Active Record is special, too, in that it does not really put any focus on the difference between ON and WHERE clauses. In practice it [puts the focus on WHERE](http://guides.rubyonrails.org/active_record_querying.html#retrieving-filtered-data-from-multiple-tables):

```ruby
# SELECT people.id, people.name, comments.text
# FROM people
# INNER JOIN comments
#         ON comments.person_id = people.id
# WHERE comments.created_at > '2015-01-01'
Person
  .select('people.id, people.name, comments.text')
  .joins(:comments)
  .where('comments.created_at > ?', 1.week.ago)
```

GRDB can do that, but can't foster it too much because Swift programmers don't like strings at all. It's thus easier to define ON clause than WHERE clauses:

```swift
// SELECT ...
// FROM people
// INNER JOIN comments
///        ON comments.person_id = people.id
//        AND comments.created_at > '2015-01-01'
Person.joining(required: Person.comments.filter(createdAt > ...)))
```

I don't know yet if fostering ON clauses over WHERE clauses will turn out good, bad, or neutral.

So, my best answer to your question is two-fold:

1. I'm hoping that the [four joining methods](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#joining-methods) can build a conceptually sound way to navigate in the database, and that behind the initial surprising APIs, a consistent system of simple rules that scale well can emerge.
2. The inner implementation can improve over the time and enhance the generated SQL.


>> [AssociationsBasics.md#L774-L787](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L774-L787)
>>
>> ```swift
>> // SELECT book.*
>> // FROM book
>> // JOIN person ON person.id = book.authorId
>> //            AND person.countryCode = 'FR'
>> let frenchAuthor = Book.author.filter(Column("countryCode") == "FR")
>> let request = Book.joining(required: frenchAuthor)
>> ```
>
> This API is weird to me. I would expect the join and filter to be separate.

This is the return of the ON vs. WHERE. I'm listening.

> What if the query I give you has something which semantically changes the whole thing, like a group by clause? What happens if it has an order clause? Does it override the outer?

I think about this question a lot, but haven't made the whole tour of this wide landscape.

You're right: in the `req.joining(assoc)` expression, the `req` subject can be any request, including a request that is already joined, or grouped, or ordered, etc. The `assoc` argument also can be a plein association, or a filtered association, or an association joined on other associations.

This may build requests that refer to the same association several times in "conflicting" ways, or that build database sets that are impractical to express in SQL.

Now, groups work. @hartbit has a [working example](https://gist.github.com/hartbit/0acb80e57c3a90a9017b1012ab8349f5#file-grdbassoc-swift).


>> [AssociationsBasics.md#L814-L824](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L814-L824)
>>
>> When you sort both the base record on the associated record, the request is sorted on the base record first, and on the associated record next:
>
> Ok well that's probably the right answer with that API, but I still think it's a bad sign that you have to state what happens in the documentation. It's not obvious. My question about group by still stands.

I could remove the `order` method from associations, and force the user to sort the whole query. There is an example in the chapter on [Table Aliases](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#table-aliases).

But this would make it impractical to define pre-sorted associations:

```swift
struct Author {
    static let sortedBooks = hasMany(Book.self).order(Column("title"))
}
```

And if associations can be sorted, I *have* to document what happens when you sort both the main record and the association. The expected behavior is not obvious, and this is precisely why I had to document it.

Did you have something else in your mind?


>> [AssociationsBasics.md#L895](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L895)
>>
>> Table Aliases
>
> Heh. The bane of my existence (our theoretical API for that is basically the same, but it's much harder to do when you enforce that it's verified at compile time)

We need to find a balance. Too much type safety, and your API won't let you run any SQL snippet without painful ceremony. Not enough type safety, and you don't look serious enough. A core value of GRDB is that is "never punishes you for writing SQL". I walk on a razor blade.

>> [AssociationsBasics.md#L1247-L1254](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L1247-L1254)
>>
>> Known Issues - You can't chain a required association on an optional association
>
> I think you should solve this before stabilizing the API. It will have greater ramifications on the public API surface than you expect.

This one is tricky. You know it. I feel it, too.

To understand it well, consider this SQL request made of left joins which fits well `A.including(optional: A.b.including(optional: B.c))` (two optionals):

```sql
SELECT a.*, b.*, c.*
FROM a
LEFT JOIN b ON ...
LEFT JOIN c ON ...
-- a.id  b.id  c.id
-- 1     1     1
-- 2     2     NULL
-- 3     NULL  NULL
```

It returns three rows. The second one is interesting: it has one A, one B, but no C (the second row). This second row is not our friend.

The request `A.including(optional: A.b.including(required: B.c))` (an optional then a required, currently not supported) is supposed to return all As that have, or not, a (B that must have a C). This means that you expect an SQL query that outputs:

```sql
-- a.id  b.id  c.id
-- 1     1     1
-- 2     NULL  NULL
-- 3     NULL  NULL
```

But I don't know how to write an SQL query that outputs this results.

And...... Tadaaa! If I can't *generate SQL* that outputs the expected rows, I can still work on the *row consumption* side: if the second row does not contains values for both B and C, I can pretend there is no B.

All right, @sgrif, thanks for the kick-off. I may be able to solve this known issue, after all.

>> [AssociationsBasics.md#L1259-L1267](https://github.com/groue/GRDB.swift/blame/GRDB3-Associations/Documentation/AssociationsBasics.md#L1259-L1267)
>>
>> Joining two associations with the same association key at the same level is undefined behavior
>
> I'm not familiar with your library, but this should be easy to prevent at compile time shouldn't it?
>
> Either way, calling it "undefined behavior" is a bit heavy handed. You're not going to delete my hard drive. Just say it's unsupported.

"Unsupported" is much better, yes.

No I won't be able to catch this at compile time.

The real trouble here is **association reuse**. For example, let's consider that books have authors, and authors have hats and coats. You could join author plus hat first, and later author plus coat:

```swift
let request = Book.including(required: Book.author.including(optional: Author.hat))
let request2 = request.including(required: Book.author.including(optional: Author.coat))
```

In this case, the *intent* of the user is not to perform two SQL joins on the author table. The *intent* is to get books with its author, and eventually its author's hat. And eventually its author's coat.

That's, at least, the seed I'd like to plant in GRDB users' heads.

To support such logic, I'll eventually have to spot association reuse, and perform some kind of merging. Implementation-wise, it will be a merge of [Trees of Association Keys](https://github.com/groue/GRDB.swift/blob/GRDB3-Associations/Documentation/AssociationsBasics.md#the-structure-of-a-joined-request). The algorithm is yet to be found. 


> Ok that's all I've got. I skimmed over a good chunk of it. Sorry, it's *very* dense documentation.

That's already immensely helpful. Thanks you **a lot**, @sgrif. Wish I had to write Rust at my job ;-)

Issue #333, opened 2018-04-13
Memory inefficiency in DATETIME parsing


sobri909:
Hey 😄 I've finally found some time to get back to database code. Yay!

So anyway, I've found an interesting one while dealing with very large result sets that include `DATETIME` fields.

I've got some result sets that can get up into the tens or hundreds of thousands (the one I'm looking at right now is ~40k rows), and I need to walk the entire sequence in a single run, as efficiently as possible. But while walking through the results I've been seeing memory use balloon to over 600MB (at which point iOS kills the app).

So I figured the problem was all those model instances in memory. But turns out it wasn't. The model instances themselves are only taking up a handful of megabytes. The *real* memory eater is a "leak" in `Scanner` (as used [here](https://github.com/groue/GRDB.swift/blob/master/GRDB/Core/Support/Foundation/DatabaseDateComponents.swift#L135) in `DatabaseDateComponents`). 

Turns out those scanners are pooping out mallocs like crazy, and expecting someone else to clean up after them. Once the sequence has finished iterating, the mallocs do eventually get cleaned up. But if the result set is large enough it won't get that far, and will instead get killed off by iOS. For my 40k rows result set, I'm seeing around 600MB of wasteful mallocs, then iOS terminating the process before iteration completes. 

Basically `Scanner` is a slob.

I'm currently successfully working around it by using a cached DateFormatter: 

```swift
internal lazy var dateFormatter: DateFormatter = {
    let formatter = DateFormatter()
    formatter.dateFormat = "yyyy-MM-dd HH:mm:ss.SSS"
    formatter.timeZone = TimeZone(abbreviation: "UTC")
    return formatter
}()
```

```swift
if let dateString = value.storage.value as? String {
    return (column, store.dateFormatter.date(from: dateString))
}
```

But that obviously isn't a portable solution, because of the hardcoded `dateFormat`.

While hunting for better solutions I stumbled onto [this blog post](http://jordansmith.io/performant-date-parsing/), which makes use of `vsscanf()`, which looks super fun! But I haven't had a chance to experiment with it yet.

sobri909:
I'm guessing the *most* efficient solution would be to hand the job over to SQLite, and fetch the datetimes as `strftime('%s')` then run them through `Date(timeIntervalSince1970:)`. But I couldn't see a way to tidily fit that in. And given that SQLite is seemingly returning datetimes as strings by default (or am I wrong on that?), there might be a good reason why using unix timestamps aren't desired.

[...]

PR #334, opened 2018-04-16
vsscanf() based date string parser

sobri909:
I've temporarily left in the `useScanfStrategy` debug bool, to make it easy for you to do your own comparison testing, if you want.

### DIFF
--- a/Tests/GRDBTests/FoundationDateTests.swift
+++ b/Tests/GRDBTests/FoundationDateTests.swift
@@ -259,7 +259,7 @@ class FoundationDateTests : GRDBTestCase {
             XCTAssertEqual(calendar.component(.hour, from: date), 1)
             XCTAssertEqual(calendar.component(.minute, from: date), 2)
             XCTAssertEqual(calendar.component(.second, from: date), 3)
-            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 600_000) // permit precision loss down to 3 sig figs
+            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 10) 
         }
     }
 }
--- a/Tests/GRDBTests/FoundationDateTests.swift
+++ b/Tests/GRDBTests/FoundationDateTests.swift
@@ -179,7 +179,7 @@ class FoundationDateTests : GRDBTestCase {
             XCTAssertEqual(calendar.component(.hour, from: date), 1)
             XCTAssertEqual(calendar.component(.minute, from: date), 2)
             XCTAssertEqual(calendar.component(.second, from: date), 3)
-            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 10)
+            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_000_000) < 10)  // We actually get 4_000_008. Some precision is lost during the DateComponents -> Date conversion. Not a big deal.
         }
     }
 
@@ -259,7 +259,7 @@ class FoundationDateTests : GRDBTestCase {
             XCTAssertEqual(calendar.component(.hour, from: date), 1)
             XCTAssertEqual(calendar.component(.minute, from: date), 2)
             XCTAssertEqual(calendar.component(.second, from: date), 3)
-            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 10) 
+            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_000_000) < 10)  // We actually get 4_000_008. Some precision is lost during the DateComponents -> Date conversion. Not a big deal.
         }
     }
 }
--- a/Tests/GRDBTests/FoundationDateTests.swift
+++ b/Tests/GRDBTests/FoundationDateTests.swift
@@ -179,7 +179,7 @@ class FoundationDateTests : GRDBTestCase {
             XCTAssertEqual(calendar.component(.hour, from: date), 1)
             XCTAssertEqual(calendar.component(.minute, from: date), 2)
             XCTAssertEqual(calendar.component(.second, from: date), 3)
-            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_000_000) < 10)  // We actually get 4_000_008. Some precision is lost during the DateComponents -> Date conversion. Not a big deal.
+            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 10)
         }
     }
 
@@ -259,7 +259,7 @@ class FoundationDateTests : GRDBTestCase {
             XCTAssertEqual(calendar.component(.hour, from: date), 1)
             XCTAssertEqual(calendar.component(.minute, from: date), 2)
             XCTAssertEqual(calendar.component(.second, from: date), 3)
-            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_000_000) < 10)  // We actually get 4_000_008. Some precision is lost during the DateComponents -> Date conversion. Not a big deal.
+            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 10) 
         }
     }
 }
--- a/Tests/GRDBTests/FoundationDateTests.swift
+++ b/Tests/GRDBTests/FoundationDateTests.swift
@@ -259,7 +259,7 @@ class FoundationDateTests : GRDBTestCase {
             XCTAssertEqual(calendar.component(.hour, from: date), 1)
             XCTAssertEqual(calendar.component(.minute, from: date), 2)
             XCTAssertEqual(calendar.component(.second, from: date), 3)
-            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 10) 
+            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 600_000) // permit precision loss down to 3 sig figs
         }
     }
 }
--- a/GRDB.xcodeproj/project.pbxproj
+++ b/GRDB.xcodeproj/project.pbxproj
@@ -259,6 +259,7 @@
 		566B91331FA4D3810012D5B0 /* TransactionObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 566B91321FA4D3810012D5B0 /* TransactionObserver.swift */; };
 		566B91361FA4D3810012D5B0 /* TransactionObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 566B91321FA4D3810012D5B0 /* TransactionObserver.swift */; };
 		566B91391FA4D3810012D5B0 /* TransactionObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 566B91321FA4D3810012D5B0 /* TransactionObserver.swift */; };
+		56707201208A509C006AD95A /* DateParsingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 567071FA208A509C006AD95A /* DateParsingTests.swift */; };
 		567156181CB142AA007DC145 /* DatabaseQueueReadOnlyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 567156151CB142AA007DC145 /* DatabaseQueueReadOnlyTests.swift */; };
 		5671FC201DA3CAC9003BF4FF /* FTS3TokenizerDescriptor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5671FC1F1DA3CAC9003BF4FF /* FTS3TokenizerDescriptor.swift */; };
 		5671FC231DA3CAC9003BF4FF /* FTS3TokenizerDescriptor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5671FC1F1DA3CAC9003BF4FF /* FTS3TokenizerDescriptor.swift */; };
@@ -843,6 +844,7 @@
 		566B91221FA4CF810012D5B0 /* Database+Schema.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "Database+Schema.swift"; sourceTree = "<group>"; };
 		566B912A1FA4D0CC0012D5B0 /* StatementAuthorizer.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StatementAuthorizer.swift; sourceTree = "<group>"; };
 		566B91321FA4D3810012D5B0 /* TransactionObserver.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TransactionObserver.swift; sourceTree = "<group>"; };
+		567071FA208A509C006AD95A /* DateParsingTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = DateParsingTests.swift; sourceTree = "<group>"; };
 		567156151CB142AA007DC145 /* DatabaseQueueReadOnlyTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = DatabaseQueueReadOnlyTests.swift; sourceTree = "<group>"; };
 		567156701CB18050007DC145 /* EncryptionTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = EncryptionTests.swift; sourceTree = "<group>"; };
 		5671FC1F1DA3CAC9003BF4FF /* FTS3TokenizerDescriptor.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = FTS3TokenizerDescriptor.swift; sourceTree = "<group>"; };
@@ -1574,6 +1576,7 @@
 				56DE7B361C42BBBB00861EB8 /* PerformanceCoreDataTests.sqlite */,
 				56DE7B2D1C42B23B00861EB8 /* PerformanceModel.xcdatamodeld */,
 				563363F81C960711000BE133 /* PerformanceRealmTests.realm */,
+				567071FA208A509C006AD95A /* DateParsingTests.swift */,
 				56DE7B271C41302500861EB8 /* FetchNamedValuesTests.swift */,
 				56DE7B291C4130AF00861EB8 /* FetchPositionalValuesTests.swift */,
 				56DE7B2B1C41311900861EB8 /* FetchRecordClassTests.swift */,
@@ -2136,6 +2139,7 @@
 			files = (
 				56D3BE711F4EB1A00034C6D2 /* FetchRecordStructTests.swift in Sources */,
 				560C98231C0E23BB00BF8471 /* InsertRecordClassTests.swift in Sources */,
+				56707201208A509C006AD95A /* DateParsingTests.swift in Sources */,
 				56DE7B2C1C41311900861EB8 /* FetchRecordClassTests.swift in Sources */,
 				56DE7B281C41302500861EB8 /* FetchNamedValuesTests.swift in Sources */,
 				56D507831F6D7B2E00AE1C5B /* InsertRecordStructTests.swift in Sources */,
--- a/GRDB.xcodeproj/project.pbxproj
+++ b/GRDB.xcodeproj/project.pbxproj
@@ -625,6 +625,9 @@
 		56FF455A1D2CDA5200F21EF9 /* RecordUniqueIndexTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 56FF45551D2CDA5200F21EF9 /* RecordUniqueIndexTests.swift */; };
 		6340BF801E5E3F7900832805 /* RecordPersistenceConflictPolicy.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6340BF7F1E5E3F7900832805 /* RecordPersistenceConflictPolicy.swift */; };
 		6340BF841E5E3F7900832805 /* RecordPersistenceConflictPolicy.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6340BF7F1E5E3F7900832805 /* RecordPersistenceConflictPolicy.swift */; };
+		C96C0F2B2084A442006B2981 /* SQLiteDateParser.swift in Sources */ = {isa = PBXBuildFile; fileRef = C96C0F242084A442006B2981 /* SQLiteDateParser.swift */; };
+		C96C0F2C2084A459006B2981 /* SQLiteDateParser.swift in Sources */ = {isa = PBXBuildFile; fileRef = C96C0F242084A442006B2981 /* SQLiteDateParser.swift */; };
+		C96C0F2D2084A45A006B2981 /* SQLiteDateParser.swift in Sources */ = {isa = PBXBuildFile; fileRef = C96C0F242084A442006B2981 /* SQLiteDateParser.swift */; };
 		DC2393C81ABE35F8003FF113 /* GRDB-Bridging.h in Headers */ = {isa = PBXBuildFile; fileRef = DC2393C61ABE35F8003FF113 /* GRDB-Bridging.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		DC3773F919C8CBB3004FCF85 /* GRDB.h in Headers */ = {isa = PBXBuildFile; fileRef = DC3773F819C8CBB3004FCF85 /* GRDB.h */; settings = {ATTRIBUTES = (Public, ); }; };
 /* End PBXBuildFile section */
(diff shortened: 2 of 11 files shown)


groue:
Thanks a lot, @sobri909. I'll answer as shortly as I can!

sobri909:
Thanks @groue 😄

Another couple of implementation notes: 

- The code in the original [blog post ](http://jordansmith.io/performant-date-parsing/) reused a single `DateComponents` instance across all requests, presumably to avoid any instantiation costs there. I didn't try using that optimisation here. Improved timing performance wasn't the priority, and it felt like something that could easily introduce bugs during later edits.

- `SQLiteDateParser` could've been generalised to return `DateComponents` instead of `DatabaseDateComponents`. But ... meh. Can always do that later, if it ever needs to happen.

sobri909:
I've cleaned up all the obvious test fails. But travis is failing out on what looks like unrelated stuff now (testCopyTransfersChanges, etc). I'll have a poke around in those tests tomorrow morning.

groue:
Hello @sobri909,

> The code in the original blog post reused a single DateComponents instance across all requests, presumably to avoid any instantiation costs there. I didn't try using that optimisation here. Improved timing performance wasn't the priority, and it felt like something that could easily introduce bugs during later edits.

💯 I fully agree.

We can avoid the static methods, shared allocated pointers, and the mutex with `withUnsafeMutablePointer`:

```swift
struct SQLDateComponentsParser {
    struct SQLDateComponents {
        var year: Int32 = 0
        var month: Int32 = 0
        var day: Int32 = 0
    }
    
    func parse(string: String) -> SQLDateComponents {
        var components = SQLDateComponents()
        let parseCount = withUnsafeMutablePointer(to: &components.year) { yearP in
            withUnsafeMutablePointer(to: &components.month) { monthP in
                withUnsafeMutablePointer(to: &components.day) { dayP in
                    withVaList([yearP, monthP, dayP]) { vaList in
                        vsscanf(string, "%d-%d-%d", vaList)
                    }
                }
            }
        }
        return components
    }
}

let sqlComponents = SQLDateComponentsParser().parse(string: "2018-04-16")
sqlComponents.year  // 2018
sqlComponents.month // 4
sqlComponents.day   // 16
```

> SQLiteDateParser could've been generalised to return DateComponents instead of DatabaseDateComponents. But ... meh. Can always do that later, if it ever needs to happen.

This is interesting, especially considering that [GRDB3](https://github.com/groue/GRDB.swift/pull/313) is on the way: we can break things, fix design issues, and why not question [DatabaseDateComponents](https://github.com/groue/GRDB.swift/blob/v2.10.0/README.md#datecomponents).

We'll see that it's not that simple (but I'm still quite open to suggestions):

DatabaseDateComponents, today, groups a plain DateComponents, and a format (YMD, etc), so that one knows the format of the parsed date string.

It is important to distinguish those formats. A missing component is not equivalent to a component that is explicitly set to 0: "2018-04-16" != "2018-04-16 00:00:00".

This format is exposed to the end user, for whatever useful purpose. It is also used internally, when GRDB decodes Date from String, via DatabaseDateComponent. For example, "12:00:00" can be converted to DatabaseDateComponent, with format `.HMS`, but it can't be converted to Date because it lacks YMD information. See [Date.init?(databaseDateComponents:)](https://github.com/groue/GRDB.swift/blob/v2.10.0/GRDB/Core/Support/Foundation/Date.swift#L60).

Now, there is information redundancy here, since components in DateComponents are optional. For example, to know if an SQLite string contains time information, one could just check the `hour` component:

```swift
// Why not?
let dateComponents: DateComponents = row["date"]
if dateComponents.hour == nil {
    // parsed "YYYY-MM-DD"
} else {
    // parsed... something... else (???)
}
```

But I think that this is easier to do with a plain and boring switch on a format enum, as today:

```swift
// Current API
let dbComponents: DatabaseDateComponents = row["date"]
switch dbComponents.format {
case .YMD:
    // parsed "YYYY-MM-DD"
case .YMD_HM:
    // parsed "YYYY-MM-DD HH:MM"
...
}
```

Finally, removing DatabaseDateComponents would expose us to a *writing* problem: not all DateComponents can be converted to valid SQLite date strings. When you store a DateComponents will nil components except year, what are you supposed to store in the database?

```swift
var components = DateComponents()
components.year = 2018

// What here?
// - "2018" is not a valid SQL date
// - "2018-01-01" is valid, but...
// - ...why not "2018-01-01 00:00:00.000" then?
// - fatal error, in order to complain about ambiguous input?
let string = String.fetchOne(db, "SELECT ?", arguments: [components])
```

DatabaseDateComponents does not suffer from this ambiguity. The user never doubts which format is used. All nil components in the requested format are given default values when needed:

```swift
var components = DateComponents()
components.year = 2018

// "2018-01-01 00:00"
let dbComponents = DatabaseDateComponents(components, format: .YMD_HM)
let string = String.fetchOne(db, "SELECT ?", arguments: [dbComponents])
```

> I've cleaned up all the obvious test fails. But travis is failing out on what looks like unrelated stuff now (testCopyTransfersChanges, etc). I'll have a poke around in those tests tomorrow morning.

Those tests check that some record information is preserved by some transformations (copy, roundtrip through the database, etc). And it happens that the involved records have a date property :-) Those tests are not flaky, and have not been modified in a while: it's very likely that they spot an actual regression.

groue:
> it's very likely that [failing tests] spot an actual regression.

I'd look at [testRecordIsNotEditedAfterFullFetch](https://github.com/groue/GRDB.swift/blob/v2.10.0/Tests/GRDBTests/RecordEditedTests.swift#L161-L177) first: it should be the easier to track in the debugger. The `Person.init(row:)` initializer reads a Date that does not convert to the same database string as the database string it was parsed from. Hence a wrong output from `hasDatabaseChanges`.

Tests/GRDBTests/FoundationDateTests.swift
```diff
@@ -179,7 +179,7 @@ class FoundationDateTests : GRDBTestCase {
             XCTAssertEqual(calendar.component(.hour, from: date), 1)
             XCTAssertEqual(calendar.component(.minute, from: date), 2)
             XCTAssertEqual(calendar.component(.second, from: date), 3)
-            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_000_000) < 10)  // We actually get 4_000_008. Some precision is lost during the DateComponents -> Date conversion. Not a big deal.
+            XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_560_000) < 10)
```
groue: This change needs to be reverted.

The test was making sure that GRDB handles fractional seconds is a compatible way with SQLite. Extra precision has to be [ignored](https://github.com/groue/GRDB.swift/blob/v2.10.0/GRDB/Core/Support/Foundation/DatabaseDateComponents.swift#L268-L270):

```swift
let seconds = try Double.fetchOne(db, "SELECT strftime('%f','2015-07-22 01:02:03.00456')")!
print(seconds) // 3.005, not 3.00456
```

sobri909: @groue I'm already on it 😉 Another commit coming in a sec. 

Although SQLite *is* maintaining the precision for me locally. But we can't depend on that.

groue: > Although SQLite is maintaining the precision for me locally. But we can't depend on that.

Interesting. What do you mean?

sobri909: SQLiteDateParser is preserving precision to [9 sig figs](https://github.com/sobri909/GRDB.swift/blob/dad163771ff18368e54acd9c1248caab8ee94e66/GRDB/Core/Support/Foundation/SQLiteDateParser.swift#L101-L105). So that test is passing for me locally. (And on travis too, so far. Yay!)

Although sim tests != real device. And having the tests demand that level of precision would be brittle for any changes in SQLite engine. SQLite I don't think is making any promises on how much precision it maintains.

So I'm about to commit a change to the tests that permits precision loss down to 3 sig figs. (The old tests were *demanding* that precision loss, instead of just permitting it).

sobri909: @groue The tests were failing when precision was maintained, instead of merely permitting precision loss. So the tests were wrong 😉

The latest commit allows precision loss to the same degree as the Scanner based parser, and also passes when precision is maintained. 

SQLiteDateParser will maintain all the precision it's given, instead of discarding it.

sobri909: From the [SQLite docs](https://www.sqlite.org/lang_datefunc.html): 

> Exactly three digits are shown in the examples because only the first three digits are significant to the result, but the input string can have fewer or more than three digits and the date/time functions will still operate correctly. Similarly, format 12 is shown with 10 significant digits, but the date/time functions will really accept as many or as few digits as are necessary to represent the Julian day number.

groue: > @groue The tests were failing when precision was maintained, instead of merely permitting precision loss. So the tests were wrong 😉

You're right: "00456" was stripped to "004", when SQLite reads 0.005 fractional seconds, as shown above.

> The latest commit allows precision loss to the same degree as the Scanner based parser, and also passes when precision is maintained.

I understand you better know. But I don't agree with your "permit precision loss down to 3 sig figs" comment in https://github.com/groue/GRDB.swift/pull/334/commits/3bcc5b806425b3f94bc927e3360cdc566481c425. I want actual precision loss, not maximum precision loss.

sobri909: The previous tests would fail when precision was maintained (because `abs(4_560_000 - 4_000_000) > 10`). So the only way to pass the test would've been to discard the precision?

groue: I'm sorry for this nitpicking, especially since this PR changes code that had a bug, as you have well shown - I'm not here to give lessons. But I value above all *compatibility with SQLite* whenever we have to reproduce in Swift some inner SQLite algorithm.

sobri909: No worries 😄

I don't *think* there's any SQLite compatibility issues (unless I'm missing something - I'm not intimately familiar with SQLite yet). Their docs say they accept as many sig figs as given, and just don't promise to make use of any beyond three, when using `strftime()` etc.

So maintaining whatever precision we're given (and receive back) shouldn't create compatibility issues. I think?

groue: > The previous tests would fail when precision was maintained (because abs(4_560_000 - 4_000_000) > 10). So the only way to pass the test would've been to discard the precision?

The original test was:

```swift
// (parsing "2015-07-22 01:02:03.00456")
// We actually get 4_000_008. Some precision is lost during the DateComponents -> Date conversion. Not a big deal.
XCTAssertTrue(abs(calendar.component(.nanosecond, from: date) - 4_000_000) < 10)
```

So it was testing that GRDB 2.10 parses `.00456` into .004 (and strips the `56` digits)

This test was not correct, since SQLite parses `.00456` into .005 (as shown by `strftime('%f','2015-07-22 01:02:03.00456')`). There is some rounding here, not performed by GRDB 2.10.

**An ideal test would assert identical parsing between GRDB and SQLite.** In all rigor, I'm not yet sure of the correct way to reach this ideal 😅

sobri909: I think a serious issue is potential overflow [here](https://github.com/sobri909/GRDB.swift/blob/3bcc5b806425b3f94bc927e3360cdc566481c425/GRDB/Core/Support/Foundation/SQLiteDateParser.swift#L19). Because it's quite possible that someone could create date strings with more than 9 points of precision.

I'm going to fix that now.

sobri909: > since SQLite parses .00456 into .005

Only if passed through `strftime()`. The tests I changed are only testing storage and retrieval of `Date` instances. And SQLite preserves full precision in that case, because it's simply storing and returning strings.

groue: > So the only way to pass the test would've been to discard the precision?

Yes, yes: we'll have to discard precision as long as SQLite does.

sobri909: Except that SQLite is only discarding precision when manipulating dates, not when storing them. (Or at least, it's not discarding the precision when they're stored as strings). 

So there's no reason to discard the precision when doing storage and retrieval? There's only a need to be aware of potential precision loss, when manipulating the dates SQL-side.

groue: (our messages have overlapped, and it may look like I dismiss your answers - I don't)

> Their docs say they accept as many sig figs as given, and just don't promise to make use of any beyond three, when using strftime() etc.

OK, I'll look at that.

sobri909: Hah. No worries 😄

> this does not mean that I'll let GRDB drift from actual, observed SQLite behavior

Likewise! The docs do say that it's fine to store more precision, and that the date/time functions will simply ignore it. But it will be worth testing to make sure that's definitely true.

groue:
@sobri909, I own you an apology, because I didn't guide you well.

Our above conversation has me think, and I'm not sure yet how I will conclude. I have been fuzzy, and am actually unable right now to precisely define what I mean by "compatibility with SQLite".

Besides, I feel uncomfortable that GRDB currently does not allow one to get the extra digits via date components. This does not fit well with the "experts welcome" design goal of this lib.

So maybe I'll rally your view, eventually.

So:

1. I suggest that this PR does what it claims, and no more: just replace Scanner with a more efficient date parsing technique, with total compatibility with previous versions of GRDB (hence *truncate* behind the three first digits). Not a single test is changed, despite the flaws you have noticed.

2. We both sleep on our discussion, and eventually come up with a desired enhancement to date/date components parsing.

What do you think?

sobri909:
@groue Sounds fine to me!

I'll change the string length truncations in the last commit to truncate down to max of 3 sig figs. Then later if you want to allow preservation of more sig figs, it's an easy two line change.

sobri909:
Hm. Travis is always timing out in the `"TID=GRDBiOS (Xcode 9.2, iOS <MIN>)"` tests set. It's stalling seemingly after doing `EnumeratedCursorTests`, so possibly during `FTS3PatternTests`.

Seems unrelated, and possibly just travis being moody, as it often is. But seems suspicious that it's always in the same place, even though that place doesn't appear to be related to the changes at all.

Not sure what to make of that...

sobri909:
Ok, now that the tests are passing, I can finally read and reply to your refactor thoughts!

*Aside: Except for those `iOS <MIN>` iPhone 4s tests timing out, which probably just mean travis is "holding it wrong"* 😏

For `withUnsafeMutablePointer()`, I've done some more reading, and I *think* I understand what it's for now.

It seems like it's for when you want to use Swift vars for C / Objective-C functions that expect pointers? So you can use `withUnsafeMutablePointer()` without having to worry about manually managing alloc / dealloc of the vars?

But it has an awkward API that would require 7 levels of nesting, because you can only hand in one var per call 😞 And it seems like its automatic memory management wouldn't be useful in this case, because we currently don't need to ever dealloc the vars. So there's nothing to be gained?

> avoid the static methods, shared allocated pointers, and the mutex 

Haha. I consider those the main feature of the parser 😄 Because it only ever allocates its working memory once, it doesn't need to alloc/dealloc anything on each run, and isn't at risk of leaving any allocations behind like `Scanner` does. 

For `DatabaseDateComponents` vs `DateComponents`, I agree with everything there! The extra decorations from `DatabaseDateComponents` all make sense, and make life easier. 

My only thought was to change `SQLiteDateParser` so that it didn't know anything about `DatabaseDateComponents`, and just returned `DateComponents` instead. Then whichever class consumes that result could inspect the components and decide which of the supported fixed formats could be used in the final `DatabaseDateComponents`. 

But that's just rearranging the furniture without any real gain. Until there's something useful to be gained from a reshuffle, it doesn't seem worth the time?

groue:
> Ok, now that the tests are passing, I can finally read and reply to your refactor thoughts!
> 
> *Aside: Except for those `iOS <MIN>` iPhone 4s tests timing out, which probably just mean travis is "holding it wrong"* 😏

[The job](https://travis-ci.org/groue/GRDB.swift/jobs/367537165) ends with "No output has been received in the last 10m0s"... I agree: don't bother about it.

> For `withUnsafeMutablePointer()`, I've done some more reading, and I *think* I understand what it's for now.
> 
> It seems like it's for when you want to use Swift vars for C / Objective-C functions that expect pointers? So you can use `withUnsafeMutablePointer()` without having to worry about manually managing alloc / dealloc of the vars?

Yes, it frees us from any heap allocation:

```swift
let i = 1 // stack allocated
withUnsafeMutablePointer(to: &i) { p in
    // use pointer to 1
}
```

> But it has an awkward API that would require 7 levels of nesting, because you can only hand in one var per call 😞

It's ugly and awkward, true. But never mind: the compiler optimizes out all calls to `withUnsafeMutablePointer`, and:

> And it seems like its automatic memory management wouldn't be useful in this case, because we currently don't need to ever dealloc the vars. So there's nothing to be gained?
>
> > avoid the static methods, shared allocated pointers, and the mutex 
> 
> Haha. I consider those the main feature of the parser 😄 Because it only ever allocates its working memory once, it doesn't need to alloc/dealloc anything on each run, and isn't at risk of leaving any allocations behind like `Scanner` does. 

We gain heap allocations (there is none), and since we no longer have any global constants, we can remove the mutex: we can parse concurrently from various threads.

> For `DatabaseDateComponents` vs `DateComponents`, I agree with everything there! The extra decorations from `DatabaseDateComponents` all make sense, and make life easier. 
> 
> My only thought was to change `SQLiteDateParser` so that it didn't know anything about `DatabaseDateComponents`, and just returned `DateComponents` instead. Then whichever class consumes that result could inspect the components and decide which of the supported fixed formats could be used in the final `DatabaseDateComponents`. 
> 
> But that's just rearranging the furniture without any real gain. Until there's something useful to be gained from a reshuffle, it doesn't seem worth the time?

I understand. But we need to know the parsed format, and the parser is quite competent at that.

sobri909:
> The static version has seven heap allocations (that's not much), but we have zero.

Really? I expected them to be stack, due to being static lets that request explicit allocation size at compile time...

Although having just done some googling and reading, some people are suggesting that Swift's `UnsafeMutablePointer` is always allocated on the heap. Hmmm...

I can't see a good reason why the compiler wouldn't optimise these specific declarations to the stack. But also can't find anyone saying that the compiler *will* do that. Sigh. 

Although even if they are heap, they're still only allocated once, so it's still only 7 allocations for the entire app session. There's no real performance cost there. It's just academic. And given the uglier code required for `withUnsafeMutablePointer()`, I'd still vote against it on stack vs heap grounds 😉

I think the more promising gain of using `withUnsafeMutablePointer()` would be getting rid of the mutex. That could give a small performance bonus.

If I get a chance, I'll try that later this week. I've had to jump over to some other tasks now, so don't have time to play with it again just yet 😞

sobri909:
Here's the [mailing list thread](https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20160926/003502.html) that discusses stack/heap for `UnsafeMutablePointer`. 

And now I remember that Swift's static lets are instantiated lazily. So yeah, pretty much guaranteed to be heap. Sigh. It's still only academic in terms of performance cost, but I'm disappointed anyway 😂

groue:
UnsafeMutablePointer is a pointer. Whether this pointer points to heap-allocated memory or stack memory is not its business. And it will even sometimes be optimized out by the compiler (making our stack/heap questions irrelevant). Consider how it compiles the two functions below (both return 2):

```swift
// test.swift

func foo() -> Int {
    var i = 1
    withUnsafeMutablePointer(to: &i) {
        $0.pointee = 2
    }
    return i
}

func bar() -> Int {
    return 2
}
```

```
$ xcrun -sdk macosx swiftc -emit-assembly -O test.swift 
...
__T04test3fooSiyF:
	pushq	%rbp
	movq	%rsp, %rbp
	movl	$2, %eax
	popq	%rbp
	retq
...
__T04test3barSiyF:
	pushq	%rbp
	movq	%rsp, %rbp
	movl	$2, %eax
	popq	%rbp
	retq
...
```

They are identical :-)

However, explicit allocations are not optimized out:

```swift
func baz() -> Int {
    let p = UnsafeMutablePointer<Int>.allocate(capacity: 1)
    defer { p.deallocate(capacity: 1) }
    p.pointee = 2
    return p.pointee
}
```

```
__T04test3bazSiyF:
	pushq	%rbp
	movq	%rsp, %rbp
	movl	$8, %edi
	movl	$7, %esi
	callq	_swift_rt_swift_slowAlloc
	movq	$2, (%rax)
	movl	$8, %esi
	movl	$7, %edx
	movq	%rax, %rdi
	callq	_swift_rt_swift_slowDealloc
	movl	$2, %eax
	popq	%rbp
	retq
```

groue:
Ha, the nested `withUnsafeMutablePointer(to:)` will be able to go when [SE-0210](https://github.com/apple/swift-evolution/blob/master/proposals/0210-key-path-offset.md), currently [in review](https://forums.swift.org/t/review-se-0210-add-an-offset-of-method-to-memorylayout/12023), is accepted :-)

groue:
Right, I'll help you 👍 Thanks for the push access on your fork!

groue:
Work in progress!

groue:
I added DateParsingTests, which parses 50000 date strings:

```
// New: vsscanf
testParseDateComponents: 0.34s
testParseDate:  0.482s

// Old: Scanner
testParseDateComponents: 0.868s
testParseDate:  1.08s
```

That's an great speed improvement :-) I don't know how to test for memory consumption, though.

groue:
Since 316d28746bdab05cdc2669228a0a3a3070dcb3d0:

```
testParseDateComponents: 0.296s
testParseDate:  0.385s
```

groue:
OK, @sobri909, I'm done.

sobri909:
Hah. I see you're going back and forth with the different ways to parse the nanoseconds, like I did 😉

Ironically I think my first attempt at it was the cleanest and most dependable (by just using `%f`), but it made it a major hassle to distinguish between an input of `12`, `12.0`, `12.00`, etc. Preserving the input string's level of precision turned out to be much easier if the nanoseconds are parsed separately from the seconds. Sigh. 

Also I'm guessing you're not a Vim user? Adding trailing whitespace to empty lines breaks Vim's ability to navigate between code blocks with `{` and `}`.

sobri909:
> I don't know how to test for memory consumption, though.

For this, I've got some super super old code sitting around somewhere, from when I had to be aware of memory consumption when doing some large image processing stuff during the iOS 4 days. I'll see if I can find it, and if it still works.

sobri909:
Ok, I've found this method, from some code back in 2012: 

```objc
- (void)calcMemory {
  mach_port_t host_port;
  mach_msg_type_number_t host_size;
  vm_size_t pagesize;

  host_port = mach_host_self();
  host_size = sizeof(vm_statistics_data_t) / sizeof(integer_t);
  host_page_size(host_port, &pagesize);

  vm_statistics_data_t vm_stat;

  if (host_statistics(host_port, HOST_VM_INFO, (host_info_t)&vm_stat, &host_size)
      != KERN_SUCCESS) {
    NSLog(@"failed to fetch vm statistics");
  }

  // stats in bytes
  natural_t mem_used = (vm_stat.active_count + vm_stat.inactive_count
      + vm_stat.wire_count) * pagesize;
  natural_t mem_free = vm_stat.free_count * pagesize;
  self.totalMemory = mem_used + mem_free;

  //NSLog(@"used: %umb free: %umb total: %dmb", mem_used / 1024 / 1024,
  //    mem_free / 1024 / 1024, self.totalMemory / 1024 / 1024);
}
```

Just trying to make sense of it now...

groue:
> Hah. I see you're going back and forth with the different ways to parse the nanoseconds, like I did 😉

Yeah! I understand you! I had to add tests because I was sweating! I just wanted to dig a little more deeper in the raw parsing to gain a few cycles (more on that below).

> Also I'm guessing you're not a Vim user? Adding trailing whitespace to empty lines breaks Vim's ability to navigate between code blocks with { and }.

No, I use Xcode, and GRDB project uses Xcode conventions throughout. I guess a configurable professional editor like vim can handle that with a few mods 😅 

>> test for memory consumption
>
> Ok, I've found this method, from some code back in 2012

You know, we were supposed to fix the original issue, #333. If it looks fixed to you, we'll be able to merge this PR!

I'm very happy with it, because we've learned and improved many things:

- I didn't know about vsscanf
- Using vsscanf was an opportunity to have Date and DatabaseDateComponents adopt StatementColumnConvertible, the protocol for values that efficiently feed from raw SQLite data (like Int, Double, String, etc.)
- The profiler shows that 50% of time is spent building the vaList that feeds vsscanf. To me this means that vsscanf itself is surely fast, but also that it's pretty expensive to call it from Swift. I conclude that eventual lightning-fast date parsing in GRDB will... not use vsscanf ;-)

groue:
We're ready for a merge, then :-)

groue:
This PR will land in GRDB 3.0

sobri909:
Great! 😄

**Semi related:** Apologies for not having tried 3.0 yet. I've been doing some db level refactors in LocoKit that I wanted to get stable first, before throwing any new variables into the equation. But it looks like those parts of my refactor are looking settled now, so I'll probably switch to 3.0 today / tomorrow. (More likely tomorrow, as I've got a flight to catch today).

PR #350, opened 2018-05-12
Use `String(format: ...)`

fpillet:
... instead of `NSString(format: ...) as String`


fpillet:
Additional info: when compiling on Linux, can't use NSString because transparent bridging `as String` is not available. Therefore, we can safely use `String(format: ...)` for these format strings as they only use `CVarArg`-compatible types

groue:
Thank you Florent! I'm sure those lines were written during Swift 2 times.

groue:
Happy to welcome you in contributors 💯

PR #351, opened 2018-05-13
Annotate `DatabaseQueue.write(_:)` with `@discardableResult`

freak4pc:
The returned result from `write(_:)` isn't necessarily needed, for example when just updating, or writing a new record in a database queue. 

e.g. 

```swift
try dbQueue.write { db in 
    try myModel.update(db)
}
```

### DIFF
--- a/GRDB/Core/DatabaseQueue.swift
+++ b/GRDB/Core/DatabaseQueue.swift
@@ -230,6 +230,7 @@ extension DatabaseQueue {
     ///
     /// - parameter block: A block that executes SQL statements.
     /// - throws: An eventual database error, or the error thrown by the block.
+    @discardableResult
     public func write<T>(_ block: (Database) throws -> T) rethrows -> T {
         return try writer.sync { db in
             var result: T? = nil


groue:
Hello @freak4pc,

I'm reluctant to accept this pull request. Look at the standard library behavior:

```swift
var i = 1

// no warning
withUnsafePointer(to: &i) { _ in
    return ()
}

// warning: Result of call to 'withUnsafePointer(to:_:)' is unused
withUnsafePointer(to: &i) { _ in
    return 1
}
```

`withUnsafePointer` (and other stdlib methods that return the result of their closure argument) are not declared with the @discardableResult attribute. That would be dangerous.

Swift is nice, and does not emit this warning when the closure returns Void, as in the 1st example above.

I too wish that Swift were even nicer, and would not emit this warning for single-line closures that call a @discardableResult method. This should be pitched and discussed with core team and community on [Swift forums](http://forums.swift.org).

But the proper way to avoid this warning, in the current state of Swift, is to explicitly discard the value, as you would do with any other stdlib method:

```swift
try dbQueue.write { db in
    _ = try makeResultWithSideEffect()
}
```

Did I miss something that should have GRDB chose another path?

freak4pc:
Thanks for the feedback @groue. 

I accept your opinion, but I don't agree to be 100% honest - As the benevolent dictator of this library its 💯% up to you 😉 

I'll reiterate on my point again. Looking at the standard library, or general frameworks - the `@discardableResult` annotation is used for methods where its _unlikely_ you'll want to use the return value in the major percentage of use cases. 

For example `Set.insert(_:)` has a return value, but also declares `@discardableResult` since using the return value is often not the main use case of that method. 

When _reading_ from a database, the return value is critical - but I'd argue that when _writing_ into the database, you usually don't expect a return value since *you didn't ask for any value* (much like the Set's insert , for example) 

Hope you give this another consideration - I believe that he usage of this flag is recommended where it creates an expected behavior for a developer that uses a specific method. If 90% of the time the developer needs to `_ = method()` to silence the warning, I believe this is not the correct planning.

freak4pc:
> I'm not sure at all that it is the case for dbQueue.write.

I think my argument is that most people do a write transaction without necessarily returning a value from it, but given you're against it I'm OK to close

groue:
All right @freak4pc! I hope I was able to clearly expose my reasoning, and the proper way out of this warning 😅

PR #353, opened 2018-05-19
Linux compilation support

fpillet:
Main framework compiles and runs on Linux.
Still fighting with SPM to get tests compiling (due to issue with building CSQLite in some conditions)


fpillet:
Sorry wrong target branch :(

PR #354, opened 2018-05-19
Main framework compiles on Linux

fpillet:
Main framework compiles and runs on Linux.
Still fighting with SPM to get tests compiling (due to issue with building CSQLite in some conditions)


fpillet:
So the SPM issue is not with CSQLite but is the dreaded `missingLinuxMain` problem. Even though I added a `LinuxMain.swift` file in the `Tests` folder, SPM still complains. Will need to dig deeper on this.

groue:
@fpillet, this is huge news, and I hope @zmeyc is happy, too :-)

For full information, we had a Twitter conversation in which we agreed on ignoring NS-prefixed foundation types on the Linux platform, since those are unlikely to exist at runtime in a Linux project. Also, this PR is a preliminary step towards Linux support, that will be announced and given proper support in a post-GRDB 3.0 release.

I have two requests before the PR is merged:

1. The PR currently breaks the SPM package on macOS, in release configuration:

    ```
    $ make test_SPM
    swift build -c release
    <unknown>:0: error: unable to execute command: Segmentation fault: 11
    ```
    
    I could spot that this crash happens because of the changes in UUID.swift:

    ```
    $ git reset --hard 34db8efea0546983b51d941e78131c903449a840 # Top of GRD3
    $ git reset 4f5fdf45e865fc4e790bd26bf0ac2f950517b909 # Your PR
    $ git checkout -- GRDB/Core/Support/Foundation/UUID.swift # Revert UUID.swift
    $ make test_SPM # success
    ```
    
    But since I don't really know how to compile GRDB on Linux right now, I could not restore UUID support on macOs while preserving Linux. Do you think you could have a look?

2. I'd rather see LinuxMain.swift removed from the PR.
    
    The lack of Objective-C runtime on Linux requires the project to explicitly write the full list of tests. This is brittle, because it is easy to add a test and forget to extend LinuxMain.swift. Ole Begemann wrote a nice blog post on the subject: https://oleb.net/blog/2017/03/keeping-xctest-in-sync/
    
    In the [previous attempt](https://github.com/groue/GRDB.swift/pull/205) at compiling GRDB in Linux, we used Sourcery to generate LinuxMain.swift. Maybe there is a new recommended technique today. But any way, LinuxMain.swift won't be part of the distribution.

With those two updates, `#if os(Linux)` will be welcome ❤️🐧  !

fpillet:
Ok I'm on it. Since this is the only actual code change I made, I'm not surprised it happens at this very spot, even though running tests locally here worked fine and the compiler didn't crash.

Thanks for the additional info about using Sourcery to auto-generate LinuxMain, it's a nice idea! I'll remove my LinuxMain.swift from the PR.

groue:
Thanks Florent! We'll look at Sourcery or any other technique later.

Oh, an important reminder since we don't have Linux tests yet: the last Linux PR had some oddities with date coding and/or decoding. If your project uses dates, be watchful!

fpillet:
Interestingly, the code I changed for UUID works fine in a Playground where I tested it before committing. The compiler chokes when emitting SIL for this perfectly valid code :( I'll revert to your previous implementation for this method then.

zmeyc:
Hi! Maybe it will be easier to use this PR as a base: https://github.com/groue/GRDB.swift/pull/205/commits
Most of the tests were succeeding in it. I think last thing I tried to fix was NSDecimalNumber losing precision on Linux. Also it has Dockerfile for testing Linux builds on OS X. I can help upgrading it to latest version if needed.
@fpillet @groue

groue:
@zmeyc: I'm not sure #205 is the best base: it was written for GRDB 0.106.1 and Swift 2.2, when we're on the way to GRDB 3 and Swift 4.1. Too much things have changed since, and I expect merging/rebasing to be a painful job. Besides, Foundation on Linux has evolved significantly (let's hope for the best).

However, your amazing Docker/Sourcery job has an intact value. It is likely to be cherry-picked soon :-)

zmeyc:
@groue  @fpillet I agree that it's probably easier to manually patch it in, but we were using mid 2017 Swift 3.1 snapshots, so it's not that old. :) Imo it's worth keeping NS-types too, we fixed incompatibilities in them which were minor. I think it's worth browsing the entire patch and checking which of these workarounds are still needed, especially Date-related ones. I can repatch that PR into latest GRDB if you would like me to.

groue:
@zmeyc, I promise I will make sure your hard work won't be lost. Yes your entire patch will be carefully studied. You're the first father of GRDB on Linux, and this won't be forgotten. Yes I'll be happy when you jump in again in the project, so that we can finally announce full Linux support. @fpillet's patch does not come from nowhere: I guided him to #205 when he expressed his intent, and he was happy to read our conversations.

**Right now**, this PR only makes sure @fpillet can run GRDB on Linux. This is not a full Linux release yet. It will come later. Don't rush: we still have time.

groue:
@fpillet, thanks for the changes. Is the PR ready for you?

groue:
Thanks Florent!

groue:
Related: https://forums.swift.org/t/generating-linuxmain-breaks-when-renaming-a-test/12894

PR #356, opened 2018-05-23
Fixing type used in 'Demographics BelongsTo Country' example

pocketpixels:
I am reading this for the first time, so I might be wrong, but I strongly suspect the type in the belongsTo() call should be Country?


groue:
Hello @pocketpixels, you're right!

Thanks for jumping in this young documentation for a young feature.

PR #362, opened 2018-05-31
Fix buffer overflow in vsscanf date parsing

hartbit:
The buffer is 10 bytes long including the null terminator so we can only scan 9 characters.

### DIFF
--- a/GRDB/Core/Support/Foundation/SQLiteDateParser.swift
+++ b/GRDB/Core/Support/Foundation/SQLiteDateParser.swift
@@ -55,7 +55,7 @@ class SQLiteDateParser {
                             withUnsafeMutablePointer(to: &parserComponents.second) { secondP in
                                 parserComponents.nanosecond.withUnsafeMutableBufferPointer { nanosecondBuffer in
                                     withVaList([yearP, monthP, dayP, hourP, minuteP, secondP, nanosecondBuffer.baseAddress!]) { pointer in
-                                        vsscanf(cString, "%4d-%2d-%2d%*1[ T]%2d:%2d:%2d.%10s", pointer)
+                                        vsscanf(cString, "%4d-%2d-%2d%*1[ T]%2d:%2d:%2d.%9s", pointer)
                                     }
                                 }
                             }
@@ -100,7 +100,7 @@ class SQLiteDateParser {
                 withUnsafeMutablePointer(to: &parserComponents.second) { secondP in
                     parserComponents.nanosecond.withUnsafeMutableBufferPointer { nanosecondBuffer in
                         withVaList([hourP, minuteP, secondP, nanosecondBuffer.baseAddress!]) { pointer in
-                            vsscanf(cString, "%2d:%2d:%2d.%10s", pointer)
+                            vsscanf(cString, "%2d:%2d:%2d.%9s", pointer)
                         }
                     }
                 }


groue:
Oh my. Thanks @hartbit!

groue:
@sobri909

> What's the max sig figs of an SQLite date string? I think initially that printf format was set to read all the way to the max sig figs that a date string might have. So shortening it by one will drop a digit of precision.

The widest of the [official SQLite date formats](https://www.sqlite.org/lang_datefunc.html) is `YYYY-MM-DD HH:MM:SS.SSS` with three trailing decimal digits.

Yet we discovered in #334 than SQLite actually four digits (at least the 3.19.3 version installed on my mac):

```
sqlite> -- %f: fractional seconds as 'SS.SSS' string
sqlite> select strftime('%f', '1970-01-01 00:00:00.0004');
00.000
sqlite> select strftime('%f', '1970-01-01 00:00:00.0005');
00.001
```

I initially wanted to have GRDB perform the exact same parsing as SQLite, but it is impossible: the above behavior of strftime can not be infered from the documentation.

GRDB 3 parses up to 9 decimal digits (the maximum precision of Foundation.DateComponents.nanosecond), but truncates after the third digit:

```swift
let dateComponents = try! DatabaseQueue().inDatabase {
    try DatabaseDateComponents.fetchOne($0, "SELECT '1970-01-01 00:00:00.123456789'")!.dateComponents
}
dateComponents.nanosecond // 123000000
```

As [you said](https://github.com/groue/GRDB.swift/pull/334#issuecomment-381886866):

> I'll change the string length truncations in the last commit to truncate down to max of 3 sig figs. Then later if you want to allow preservation of more sig figs, it's an easy two line change.

And I'm fine with it: we comply to the official doc, are compatible with GRDB2, and we can still extend the precision later. I'm happy that the performance improvements of #334 were merged before we had to brainstorm this delicate precision question!

> Which isn't a big deal, because I think the current implementation is then later on dropping all precision beyond 3 sig figs anyway, to maintain parity with older GRDB versions.

Yes.

> But if that later precision dropping is removed, the better fix would be to enlarge the buffer by one, so that no precision is lost?

Do you think we should have read 10 digits instead (considering that the goal is to feed NSDateComponents.nanosecond and that we currently truncate after the third digit)?

Do you also think that we should have that nanosecond precision brainstorming now?

> Unless I counted wrong, and SQLite date strings aren't that long anyway. Hunting about now for an example...

This does not matter: people can feed any string, and people write buggy code. We want GRDB to behave well in all cases.

#334 spots weird dates, and this is good:

```swift
// Fatal error: could not convert database value "My birthday" to DatabaseDateComponents
try! DatabaseQueue().inDatabase {
    _ = try DatabaseDateComponents.fetchOne($0, "SELECT 'My birthday'")
}
```

#334 is also pretty tolerant (NSScanner-based GRDB 2.10.0 used to crash on this one):

```swift
let dateComponents = try! DatabaseQueue().inDatabase {
    try DatabaseDateComponents.fetchOne($0, "SELECT '1970-01-01 hello'")!.dateComponents
}
dateComponents.year // 1970
```

I'm not super happy with this, but I also can live with it: the most glaring programmer errors are properly handled. I suspect that I'll one day rewrite the date parser to make it more strict.

sobri909:
> GRDB 3 parses up to 9 decimal digits (the maximum precision of Foundation.DateComponents.nanosecond)

Aah, that's the number I was looking for this morning, but was too sleepy to find. 

Yeah, if `DateComponents.nanosecond` maxes out at 9, then I agree that this commit is correct. I was just uncertain this morning as to whether the right number was 9 or 10 😉

> Do you think we should have read 10 digits instead (considering that the goal is to feed NSDateComponents.nanosecond and that we currently truncate after the third digit)?

My understanding of the SQLite docs is that SQLite will preserve the full precision that it is given, but will only make use of up to 3 sig figs when performing date manipulations. So from that perspective, I think it's worthwhile for GRDB to retain as much precision as it's given, and leave it up to SQLite to discard precision when it wants to.

But for this particular commit, I don't see a problem. The parser is creating `DateComponents` instances, and that class maxes out at 9, so 9 is the right number 😊

groue:
> But for this particular commit, I don't see a problem. The parser is creating DateComponents instances, and that class maxes out at 9, so 9 is the right number

I don't want to bikeshed to much, but we *could* parse 10 digits, and apply rounding as SQLite does in the strftime examples above:

```swift
// If we read 10 digits and round:
func nanosecond(_ string: String) -> Int {
    return try! DatabaseQueue().read {
        try DatabaseDateComponents
            .fetchOne($0, "SELECT ?", arguments: [string])!
            .dateComponents
            .nanosecond
    }
}
nanosecond("1970-01-01 00:00:00.0000000004") // 0
nanosecond("1970-01-01 00:00:00.0000000005") // 1
}
```

sobri909:
Ooh I see what you're saying now. If a date string comes in with greater than 9 points of precision, then the parser can at least attempt to round it sensibly. 

Well that doesn't seem too bikesheddy to me. I thought for a moment there you were suggesting preserving full Double precision in general. Which the parser couldn't do anyway.

Issue #379, opened 2018-07-03
Swift 4.2 branch fails to build as a dependency


QuestofIranon:
### What did you do?

added the 4.2 branch as a dependency for a swift 4.2/ XCode 10 project
`pod 'GRDB.swift', :git => 'https://github.com/groue/GRDB.swift.git', :branch => 'swift4.2'`

### What did you expect to happen?

The library to build with Swift 4.2 set as the version

### What happened instead?

Fails to build
![image](https://user-images.githubusercontent.com/2449351/42246839-0623a216-7ed3-11e8-8b5e-919dc3b7ed59.png)

I have to set it to build in 4.0 everytime a podfile update is required.

### Environment

**GRDB flavor(s):**  GRDB
**GRDB version:** Swift 4.2 Branch
**Installation method:** CocoaPods `pod 'GRDB.swift', :git => 'https://github.com/groue/GRDB.swift.git', :branch => 'swift4.2'`
**Xcode version:** Version 10.0 beta 2 (10L177m)
**Swift version:** 4.2
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** High Sierra 10.13.5

groue:
Hello @QuestofIranon,
.
You're using beta software, and a branch that is not advertised anywhere

groue:
Hello @QuestofIranon,

**You're backcountry skiing**: beta Xcode + unadvertised branch. This is not for everyone.

I recommend that you step back:

- `pod 'GRDB.swift', '~> 3.1'`
- If Xcode 10 keeps on complaining, add at the end of your Podfile:
    
    ```ruby
    post_install do |installer|
      installer.pods_project.targets.each do |target|
        if target.name == 'GRDB.swift'
          target.build_configurations.each do |config|
            config.build_settings['SWIFT_VERSION'] = '4.1'
          end
        end
      end
    end
    ```

groue:
I'm reopening this issue until Xcode 10 is out (see https://github.com/groue/GRDB.swift/issues/383#issuecomment-404127426)

groue:
GRDB 3.3.0-beta1 is out. It supports Xcode 10 beta 6 and Swift 4.2.

PR #380, opened 2018-07-05
Working xcode10/Swift 4.2 build

QuestofIranon:
this brings the branch up to date and building with xcode10/swift 4.2 per #379


groue:
Hello again @QuestofIranon, thanks for your pull request.

It won't be accepted.

Requiring Swift 4.2 and Xcode 10 is a **breaking change**, which means that we'd have to bump GRDB to version 4. But Xcode has a compatibility mode that lets it compile Swift 4.1: we can thus preserve the Swift 4.1+ / Xcode 9.3+ requirement even after Xcode 10 has shipped. For more information, ask Google about the advantages of semantic versioning (the first one is that people are happier when things that don't have to break don't break).

For your next pull request, please ask beforehand if the direction you aim at has a chance of being merged in. For example, why didn't you answer in #379? Engage conversation.

Last advice: don't ignore the [pull request template](https://raw.githubusercontent.com/groue/GRDB.swift/master/.github/PULL_REQUEST_TEMPLATE.md) when you submit one.

See you soon, if you are in a mood for preparing Xcode 10 release.

QuestofIranon:
What is the purpose of your Swift4.2 branch if not to prepare for the XCode 10 launch?

groue:
It is a branch for experimenting with Xcode 10. It has been created in a time I didn't know much yet about this Xcode release. It's a private branch of mine, if you want. I may trash it. Or not.

GRDB does make promises that you can rely on, and those are all expressed in the README of the master branch.

There is another promise, non-written, which is that there is always a stable version of GRDB that runs with the latest stable version of Xcode. More precisely, count a few days after Xcode has shipped for GRDB to be upgraded if necessary. You will be able to run GRDB with Xcode 10 after it has officially shipped.

There is a promise I don't make, which is that beta software gets the same level of support than stable software. When you run beta, you may be lucky, and get some help (as you had in #379). But you may also be left alone fixing your issue. Running beta is risky.

And everything else is wild speculation ;-)

My turn to ask a question: why didn't you simply follow my advice in #379?

QuestofIranon:
I did take your advice
> Unless you are willing to contribute, I recommend that you step back:

groue:
Good. So there is nothing left that needs to be solved.

Issue #378, opened 2018-07-01
Nuking db during development


bellebethcooper:
Hello,

Sorry for another question so soon! I've scoured the docs re: migrations and haven't been able to find the answer I'm looking for. I'm developing an app that's a client for a web app, so I'm using GRDB to persist the user's data locally after requesting it from the server. While I'm in the development phase, there are many times I want to force the app to request server-side data in order to test changes I'm making, and I'm struggling to figure out how to do this.

When using Realm previously there was a way to skip doing any migrations and just nuke everything in the db. This is essentially what I want: a temporary way to make each new load of the app have an empty database until I'm ready to send it out to users and start using migrations.

Is something like this possible with GRDB?

Thanks a lot!

groue:
Hello again @bellebethcooper. GRDB won't nuke your database automatically :-)

My personal practice is to check a `RESET` environment variable in the scheme editor (in the "Run options" tab if I recall correctly). The app checks for the presence of this environment variable, and erases the database if it is set. When the database schema is ready, I uncheck this environnement variable.

It is easier to erase  the database when it is stored inside its dedicated repository: you juste have to remove the directory, and all database files are removed altogether (main db file, -wal and -shm files, other temporary SQLite files...)

groue:
I didn't know this need was common. Do you think we should provide an API for that? How does Realm do?

bellebethcooper:
@groue That sounds like a great approach. I'll give it a go; thanks!

I don't know how common this is, but for me there are two reasons to use it. One is during initial development, when my schema can change a lot (especially if the API I'm working with is being developed at the same time). The other is in a case like the bookmarking client I'm working on, where the app is a mobile client for a web service, and the server acts as the single source of truth, so I'm really just using GRDB (and previously Realm) for local caching to speed up loading for the user, but I don't really need to worry about migrations and so on, since I can just nuke the local database and load fresh data from the server if I need to.

In Realm, it's really simple to set this up. There's a `Bool` used in the Realm config called `deleteRealmIfMigrationNeeded`. [Docs here](https://www.realm.io/docs/swift/latest/api/classes/realm/configuration#/s:10RealmSwift0A0C13ConfigurationV06deleteA17IfMigrationNeededSbvp).

Thanks again for the quick response and for sharing your approach to this!

groue:
[...]

PR #381, opened 2018-07-06
Nuking db during development

groue:
This is a pull request that addresses #378 by @bellebethcooper.

---

A DatabaseMigrator can automatically wipe out the full database content, and recreate the whole database from scratch, if it detects that a migration has changed its definition:

```swift
var migrator = DatabaseMigrator()
migrator.eraseDatabaseOnSchemaChange = true
```

Beware! This flag can destroy your precious users' data!

Yet it may be useful in those two situations:

1. During application development, as you are still designing migrations, and the schema changes often.
    
    In this case, it is recommended that this flag does not ship in the distributed application:
    
    ```swift
    var migrator = DatabaseMigrator()
    #if DEBUG
    // Speed up development by nuking the database when migrations change
    migrator.eraseDatabaseOnSchemaChange = true
    #endif
    ```

2. When the database content can easily be recreated, such as a cache for some downloaded data.

The `eraseDatabaseOnSchemaChange` option triggers a recreation of the database if the migrator detects a *schema change*. A schema change is any difference in the `sqlite_master` table, which contains the SQL used to create database tables, indexes, triggers, and views.

---

This pull request also adds the two following methods:

- `DatabaseWriter.erase()`
- `DatabaseWriter.vacuum()`

Those are not documented in the main README, and only exposed in the reference documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Release Notes
 =============
 
+## Next Version
+
+- [#381](https://github.com/groue/GRDB.swift/pull/381): Nuking db during development
+
+
 ## 3.1.0
 
 Released June 17, 2018 &bull; [diff](https://github.com/groue/GRDB.swift/compare/v3.0.0...v3.1.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,13 @@ Release Notes
 
 ## Next Version
 
+### New
+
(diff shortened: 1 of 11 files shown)


bellebethcooper:
@groue This looks great! So easy to use, also. Thanks for the quick turnaround! I think this will be super useful in future :)

groue:
Shipped in [v3.2.0](https://github.com/groue/GRDB.swift/blob/master/CHANGELOG.md)!

Issue #424, opened 2018-10-02
How to ensure database not in use when using erase


bellebethcooper:
Hello! I think I'm probably misunderstanding something, or missing a step, but I'm stuck trying to erase my GRDB database when a user logs out of my app, as the database seems to be in use. I haven't found any clues in the code or docs as to what I'm doing wrong, so I'd appreciate any pointers! Thanks!

### What did you do?

When the user logs out of my app, I tried to erase the database like so:

```swift
if let dbQueue = dbQueue {
            do {
                try dbQueue.erase()
            } catch {
                print("error erasing db: \(error)")
            }
        }
```

### What did you expect to happen?

All data would be removed from the database, so it's clean for the next user logging in

### What happened instead?

I got the following error:

`SQLite error 1: destination database is in use`

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 3.2
**Installation method:** CocoaPods
**Xcode version:** 10.0
**Swift version:** 4.2
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 10.13.6

groue:
Hello @bellebethcooper,

The `erase` method has been [introduced](https://github.com/groue/GRDB.swift/pull/381) recently, after your much welcomed [suggestion](https://github.com/groue/GRDB.swift/issues/378). It is tested, I happily put `DatabaseMigrator.eraseDatabaseOnSchemaChange` to some good use. We generally know that it works.

So what's wrong?

The "destination database is in use" snippet appears [once](https://www.sqlite.org/src/artifact/78d3cecfbe28230a) in SQLite source code:

```c
/*
** Check that there is no open read-transaction on the b-tree passed as the
** second argument. If there is not, return SQLITE_OK. Otherwise, if there
** is an open read-transaction, return SQLITE_ERROR and leave an error 
** message in database handle db.
*/
static int checkReadTransaction(sqlite3 *db, Btree *p){
  if( sqlite3BtreeIsInReadTrans(p) ){
    sqlite3ErrorWithMsg(db, SQLITE_ERROR, "destination database is in use");
    return SQLITE_ERROR;
  }
  return SQLITE_OK;
}
```

So it looks like your app has an "open read-transaction" somewhere, whatever it means.

Does it ring a bell to you?

Is there any other connection (instance of DatabaseQueue or DatabasePool) for the same database?

Could you please run the following code, and tell if it prints anything?

```swift
do {
    try dbQueue.erase()
} catch {
[...]

PR #425, opened 2018-10-03
Fix "destination database is in use " error during database backup

groue:
This PR addresses #424.

The `DatabaseWriter.erase()` method internally uses the SQLite [Online Backup API](https://www.sqlite.org/c3ref/backup_finish.html):

```swift
extension DatabaseWriter {
    func erase() throws {
        // In order to erase the content, we backup
        // an empty in-memory database into self:
        DatabaseQueue().backup(to: self)
    }
}
```

This technique used to fail with an error of code 1 (SQLITE_ERROR), message "destination database is in use" when live SQLite prepared statement would prevent the backup from starting (see the regression tests for the precise failure setup).

It happens that Record methods keep prepared statements in a cache, in order to optimize insertions, updates, deletions, and tests for record existence.

This PR thus clears the statement cache *before* starting a backup, when we used to do it after a successful backup.

The error "destination database is in use" can still happen during a backup, if the client application keeps its own prepared statements alive. To fix this error, the client application must release all prepared statements before starting the backup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ Release Notes
 
 ## Next Version
 
+### Fixed
+
+- [#425](https://github.com/groue/GRDB.swift/pull/425): Fix "destination database is in use " error during database backup
+
+
 ### New
 
 - [#422](https://github.com/groue/GRDB.swift/pull/422): Refining Association Requests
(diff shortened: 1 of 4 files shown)


groue:
The fix is ready, @bellebethcooper 👍

I want to finish #423 first, though: you will have to keep the workaround we discussed in #424 for a little while.

bellebethcooper:
No stress, the workaround is just fine. Thanks! 👍

Issue #430, opened 2018-10-12
Subclassing Record vs using a Struct


jfahrenkrug:
### What did you do?

I'm new to the library. Thank you! I love it so far, but I'm having trouble deciding whether to make my models be
(a) subclasses of `Record` or (b) structs that implement `Codable, FetchableRecord, MutablePersistableRecord`.

The documentation points out that `Record` has change detection, which is something I'd like. On the other hand it seems as if the `struct` approach gives me JSON encoding for free, which I also like. When subclassing from `Record`, it seems as if I need to do the JSON encoding and decoding by hand in the `encode(to)` and `init(row:)` methods.

### What did you expect to happen?

It would be great if I'd get the same encoding/decoding luxury in `Record` subclasses. Maybe I'm just doing it wrong :-D.

### What happened instead?

It seems as if I get none of the encoding/decoding logic for free that I get when using structs.
There's very little documentation on how to use `Record` subclasses in advanced scenarios (like having JSON columns), so I might just be doing it wrong.

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 3.4.0
**Installation method:** CocoaPods
**Xcode version:** 10.0
**Swift version:** 4.2
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** High Sierra

### Demo Project

(none)

groue:
> I want those people [who subclass Record] to [...] feel welcomed.

> There's very little documentation on how to use Record subclasses in advanced scenarios (like having JSON columns)

> Maybe Record could be enhanced so that it provides a better support for Codable.

As you can see, there's a palatable tension between the claimed goal, and the reality 😅

Should it be addressed in the documentation, or in the code of the library? That's the question.

groue:
As a support for reflexion, and an eventual new paragraph in the Record documentation, here is a quick tour of features and their availability:

| Feature | Record subclass | Record struct | Codable record |
| ------- | --------------- | ------------- | -------------- |
| Fetching methods | X | X | X |
| Persistence methods | X | X | X |
| SQL generation via the query interface | X | X | X |
| Comparison with other records | X | X | X |
| Self-comparison | X |   |   |
| Synthesized serialization |   |   | X |
| JSON columns |   |   | X |

jfahrenkrug:
Hi @groue,

Thank you so much for your quick and thorough reply! It's very helpful to understand the reasons behind the current state of things :)

To answer your question:

>  Do you have your own opinion on this question?

[...]

PR #431, opened 2018-10-13
Working on a Codable JSON sync example

KyleLeneau:
Based on this [issue](https://github.com/groue/GRDB.swift/issues/430) that came up, I tried to take my first take at making the Codable changes for JSON sync. 

### One more thing…

- [x] Updated CHANGELOG.md

### DIFF
--- a/GRDB.xcworkspace/contents.xcworkspacedata
+++ b/GRDB.xcworkspace/contents.xcworkspacedata
@@ -20,7 +20,7 @@
          location = "group:Issue156.playground">
       </FileRef>
       <FileRef
-         location = "group:/Users/groue/Documents/git/groue/GRDB.swift/Playgrounds/CustomizedDecodingOfDatabaseRows.playground">
+         location = "group:CustomizedDecodingOfDatabaseRows.playground">
       </FileRef>
       <FileRef
          location = "group:Associations.playground">
--- a/Playgrounds/JSONSynchronization.playground/Contents.swift
+++ b/Playgrounds/JSONSynchronization.playground/Contents.swift
@@ -24,10 +24,10 @@ configuration.trace = { print($0) }
 let dbQueue = DatabaseQueue(configuration: configuration)
 
 
-// Create a database table
+// Create the database table to store the players
 
 try dbQueue.inDatabase { db in
-    try db.create(table: "players") { t in
+    try db.create(table: "player") { t in
         t.column("id", .integer).primaryKey()
         t.column("name", .text)
         t.column("score", .integer)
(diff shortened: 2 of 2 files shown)


jfahrenkrug:
@KyleLeneau This is fantastic! Thank you so much! I don't want to be greedy, but extending this example to nested records at one point would be fantastic as well: Maybe a buch of `Team`s that have an array of nested `Player`s.

groue:
Thank you @KyleLeneau! Modernizing playgrounds often drops under my radar :-) I'll give a look at this PR very shortly.

KyleLeneau:
@jfahrenkrug & @groue So there are two ways this JSON could be coming into the system that I can think on and both have different logic provide the sync we'd want to showcase in an example. What are your thoughts?

#### Option 1:
Where the players and teams are sent independently because they come from two different backend queries (like iCloud).
```javascript
{
    "players": [
        { "id": 1, "name": "Arthur", "score": 1000, "teamId": 1 },
        { "id": 2, "name": "Barbara", "score": 2000, "teamId": 1 },
        { "id": 3, "name": "Craig", "score": 500, "teamId": 2 },
    ],
    "teams": [
        { "id": 1, "name": "Red Sox" },
        { "id": 2, "name": "Dodgers" }
    ]
}
```

#### Option 2:
This option feels more natural to some but it totally depends on the backend and how it's structured.

This presents a challenge though cause you can't actually model the Codable models like this and still have them work with GRDB (or so I think/suspect) since the `players` property is a `QueryInterface` for the association. So to achieve this I think I would need a lot of mapping from input Codables to storage ones.
```javascript
{
    "teams": [
        {
            "id": 1,
            "name": "Red Sox"
            "players": [
                { "id": 1, "name": "Arthur", "score": 1000 },
                { "id": 2, "name": "Barbara", "score": 2000 },
            ]
        },
        {
            "id": 2,
            "name": "Dodgers"
            "players": [
                { "id": 3, "name": "Craig", "score": 500 },
            ]
        }
    ]
}
```

jfahrenkrug:
Great questions!

> This presents a challenge though cause you can't actually model the Codable models like this and still have them work with GRDB (or so I think/suspect) since the players property is a QueryInterface for the association.

I just faced exactly this problem yesterday. I get some JSON where only the top-level has an ID. It then contains 2 levels of nested data underneath.

It's not a problem to use the associations with this setup. It doesn't bother the JSONDecoder. However, I had to first de-serialize the JSON data "the old way": `let json = try JSONSerialization.jsonObject(with: jsonData, options: [])`, then work my way down the tree, create and persist parents before children and then add the parent's freshly created ID as a `parentId` to the child object I'm about to persist. 

For the teams/players example, it would look like this (untested code!):

    struct Player: Codable, FetchableRecord, MutablePersistableRecord {
        // Associations
        static let team = belongsTo(Team.self)
        var team: QueryInterfaceRequest<Team> {
            return request(for: Player.team)
        }
        
        // Fields
        var id: Int64?
        var name: String
        var score: Int16
        var teamId: Int64
        
        private enum CodingKeys: String, CodingKey, ColumnExpression {
            case
            id,
            name,
            score,
            teamId
        }
        
        mutating func didInsert(with rowID: Int64, for column: String?) {
            id = rowID
        }
    }
    
    struct Team: Codable, FetchableRecord, MutablePersistableRecord {
        // Associations
        static let players = hasMany(Player.self)
        var players: QueryInterfaceRequest<Player> {
            return request(for: Team.players)
        }
        
        // Fields
        var id: Int64?
        var name: String
        
        private enum CodingKeys: String, CodingKey, ColumnExpression {
            case
            id,
            name
        }
        
        mutating func didInsert(with rowID: Int64, for column: String?) {
            id = rowID
        }
    }

Those were the models. To persist this (without syncing and assuming a 1-n relationship):

    let json = try JSONSerialization.jsonObject(with: jsonData, options: [])
    
    if let json = json as? [String: Any] {
        if let teams: Array = json["teams"] as? Array<[String: Any]> {
            let decoder = JSONDecoder()
            try DataManager.shared.dbQueue.write { db in
                
                // Walk through the teams and create them one-by-one
                for teamDict in teams {
                    print("Team \(String(describing: teamDict["name"]))")
                    
                    // Ugly: re-encode team so we can use the JSONDecoder to create a persistable struct
                    let encodedTeam = try JSONSerialization.data(withJSONObject: teamDict, options: .prettyPrinted)
                    var team = try decoder.decode(Team.self, from: encodedTeam)
                    try team.insert(db)
                    
                    if let players: Array = teamDict["players"] as? Array<[String: Any]> {
                        for playerDict in players {
                            var playerWithTeamId = playerDict
                            
                            // Set the parent's ID in the JSON
                            playerWithTeamId["teamId"] = team.id
                            
                            print("Player \(String(describing: playerWithTeamId["name"]))")
                            
                            let encodedPlayer = try JSONSerialization.data(withJSONObject: playerWithTeamId, options: .prettyPrinted)
                            var player = try decoder.decode(Player.self, from: encodedPlayer)
                            try player.insert(db)
                        }
                    }
                }
            }
        }
    }

It's a bit cumbersome, but it works. The decoding/re-encoding/decoding is ugly, there must be a better way.

jfahrenkrug:
@groue Great point! The last thing I want is to complicate the sample code! What do you think about a sort-of "cookbook" section somewhere in the documentation that deals with ways on how to solve some real-world problems? 
You are absolutely right: the model layer of the network and of the database rarely ever match. That would be a great example to describe somewhere because it's probably the case for 90+% of GRDB's users who talk to a remote API.
But if you think it's best to not mention those cases in the documentation at all, I'm fine with that too :) Maybe a blog post would be a better spot for it then.

groue:
> What do you think about a sort-of "cookbook" section somewhere in the documentation that deals with ways on how to solve some real-world problems?

I think, @jfahrenkrug, that real-world problems are sometimes better addressed in a blog post or a dedicated article than in a complex sample code. A dedicated article is super cool: it can address other questions than just GRBD. It can freely pick its audience and level. It can be opinionated, and displease some people without fear. And above all, a dedicated article doesn't have to be maintained by the library author ;-) I already have a lot of documentation backlog, as you see (and there are more).

jfahrenkrug:
@groue Agreed! :-D

KyleLeneau:
> You're dealing with an application problem, not a GRDB problem.

@groue I agree, I came to the same conclusion last night after digging into it a bit more and was hoping one of you would say the same thing =)

> real-world problems are sometimes better addressed in a blog post or a dedicated article than in a complex sample code. A dedicated article is super cool: it can address other questions than just GRDB. It can freely pick its audience and level. It can be opinionated, and displease some people without fear. And above all, a dedicated article doesn't have to be maintained by the library author ;-) I already have a lot of documentation backlog, as you see (and there are more).

Haha, yep! A blog post to sync data from the API could take on so many different forms and I don't think that should be here. Example, I would rather see a blog post on how to sync with iCloud instead =)

Given all of that, I think this PR is ready to be looked at. My 2-level changes are only local on my computer but this examples moves the playground to use Codable models like the original example did (thus de-emphasizing the use of `Record` class).

groue:
@KyleLeneau, this is a great pull request, and a real help. I have nothing to add, but a big thank you.

groue:
It aims `master` instead of `development`, but they are identical today: no fuss. LGTM!

groue:
@KyleLeneau, please drop me a line if you would be happy with a push access on the repository.

groue:
Oooops, you were already invited, sorry 🤣 I need a rest.

PR #478, opened 2019-02-15
Swift 5: SQL interpolation

groue:
## Motivation

Even though [record](https://github.com/groue/GRDB.swift/blob/master/README.md#records) types and the [query interface](https://github.com/groue/GRDB.swift/blob/master/README.md#the-query-interface) are great SQL generation tools, developers sometimes need to write raw SQL queries.

Executing raw SQL queries is an error-prone task, for multiple reasons:

- SQL injection
    
    ```swift
    // DON'T DO THAT
    let name = "Robert'; DROP TABLE student; --"
    let id = 42
    try db.execute("UPDATE student SET name = '\(name)' WHERE id = \(id)")
    ```

- Implicit dependency on the database schema
    
    ```swift
    // Breaks without notice when the database schema changes
    try db.execute(
        "UPDATE student SET name = ? WHERE id = ?",
        arguments: [name, id])
    ```

- Distance between SQL parameters and their values
    
    ```swift
    // Do you spot the bug?
    try db.execute("""
        UPDATE student
        SET firstName = ?, lastName = ?, department = ?, birthDate = ?,
            registrationDate = ?, mainTeacherId = ?
        WHERE id = ?
        """, arguments: [firstName, lastName, department, birthDate,
            registrationDate, mainTeacherId])

## Solution

To address those difficulties, we introduce **SQL interpolation**, based on [Swift 5 String Interpolation](https://github.com/apple/swift-evolution/blob/master/proposals/0228-fix-expressiblebystringinterpolation.md).

```swift
extension Student: TableRecord, Codable {
    static func update(_ db: Database, firstName: String, ...) throws {
        try db.execute(literal: """
            UPDATE \(self)
            SET \(CodingKeys.firstName) = \(firstName),
                \(CodingKeys.lastName) = \(lastName),
                \(CodingKeys.department) = \(department),
                \(CodingKeys.birthDate) = \(birthDate),
                \(CodingKeys.registrationDate) = \(registrationDate),
                \(CodingKeys.mainTeacherId) = \(mainTeacherId)
            WHERE id = \(id)
            """)
    }
}
```

SQL interpolation is a great new feature of GRDB 4. However the normal expectation of developers is that raw SQL is a regular String. We thus make sure that SQL interpolation does not kick in unless explicitly requested with the `literal` argument label:

```swift
// DELETE FROM student WHERE id = 42
try db.execute(sql: "DELETE FROM \(tableName) WHERE id = \(id)")

// DELETE FROM student WHERE id = ?
try db.execute(literal: "DELETE FROM \(Student.self) WHERE id = \(id)")
```

[SQLRequest](https://github.com/groue/GRDB.swift/blob/master/README.md#building-custom-requests) is the only GRDB type which profits from implicit SQL interpolation. This brings GRDB closer to the [Room](https://developer.android.com/topic/libraries/architecture/room) persistence library, and should help sharing requests between iOS and Android applications:

```swift
// GRDB
extension Student: TableRecord, Codable {
    static func filter(name: String) -> SQLRequest<Student> {
        return "SELECT * FROM \(self) WHERE \(CodingKeys.name) = \(name)"
    }
}
```

```kotlin
// Room
@Dao
interface StudentDao {
    @Query("SELECT * FROM student WHERE name = :name")
    fun loadAllByName(name: String): Array<User>
}
```

## Detailed design

**Breaking change**: all methods that accept a raw SQL string now use an explicit `sql` argument label:

```diff
-try db.execute("INSERT ...")
-let rows = try Row.fetchAll(db, "SELECT ...")
+try db.execute(sql: "INSERT ...")
+let rows = try Row.fetchAll(db, sql: "SELECT ...")
```

We introduce the **SQLLiteral** type. SQLLiteral wraps a raw SQL string, and statement arguments:

```swift
let name = "Robert'; DROP TABLE student; --"
let id = 42
let literal: SQLLiteral = "UPDATE student SET name = \(name) WHERE id = \(id)"

literal.sql       // UPDATE student SET name = ? WHERE id = ?
literal.arguments // ["Robert'; DROP TABLE student; --", 42]
```

Use literals with `Database.execute(literal:)` method:

```swift
try db.execute(literal: literal)
```

Build SQLRequest from literals:

```swift
extension Student: TableRecord, Codable {
    static func filter(name: String) -> SQLRequest<Student> {
        // SELECT * FROM student WHERE name = ?
        var literal: SQLLiteral = "SELECT * "
        literal += "FROM \(self) "
        literal += "WHERE \(CodingKeys.name) = \(name)"
        return SQLRequest(literal: literal)
    }
}
```

Embed literals in [query interface requests](https://github.com/groue/GRDB.swift/blob/master/README.md#requests):

```swift
extension Student: TableRecord, Codable {
    static func filter(name: String) -> QueryInterfaceRequest<Student> {
        return Student.filter(literal: "\(CodingKeys.name) = \(name)")
    }
}
```

SQLRequest supports SQL interpolation:

```swift
extension Student: TableRecord, Codable {
    static func filter(name: String) -> SQLRequest<Student> {
        return "SELECT * FROM \(self) WHERE \(CodingKeys.name) = \(name)"
    }
}
```

Interpolation supports the following values:

```swift
// Raw SQL with eventual arguments
// SELECT * FROM student WHERE name = ?
let request: SQLRequest<Student> = """
    SELECT * FROM student
    WHERE \(sql: "name = ?", arguments: [name])
    """

// Other literals
// SELECT * FROM student WHERE name = ?
let condition: SQLLiteral = "name = \(name)"
let request: SQLRequest<Student> = """
    SELECT * FROM student
    WHERE \(literal: condition)
    """

// Values
// SELECT * FROM student WHERE id = ?
let id = 1
let request: SQLRequest<Student> = """
    SELECT * FROM student
    WHERE id = \(id)
    """

// Collections
// SELECT * FROM student WHERE id IN (?,?,?)
let ids = [1, 2, 3]
let request: SQLRequest<Student> = """
    SELECT * FROM student
    WHERE id IN \(ids)
    """

// TableRecord
// SELECT * FROM student
extension Student: TableRecord { }
let request: SQLRequest<Student> = """
    SELECT * FROM \(Student.self)
    """

// Expressions
// SELECT (celcius * 1.8 + 32) AS fahrenheit FROM planet
let request: SQLRequest<Double> = """
    SELECT \(CodingKeys.celcius * 1.8 + 32) AS fahrenheit
    FROM planet
    """

// Ordering
// SELECT * FROM student ORDER BY name DESC
let request: SQLRequest<Student> = """
    SELECT * FROM student
    ORDER BY \(CodingKeys.name.desc)
    """
```


GRDB/QueryInterface/SQLSelectable.swift
```diff
@@ -26,23 +26,21 @@ public protocol SQLSelectable {
...
-        self.arguments = arguments
+    init(literal sqlLiteral: SQLLiteral) {
+        self.sqlLiteral = sqlLiteral
     }
     
     func resultColumnSQL(_ context: inout SQLGenerationContext) -> String {
-        if let arguments = arguments {
-            if context.appendArguments(arguments) == false {
+        if !sqlLiteral.arguments.isEmpty {
+            if context.appendArguments(sqlLiteral.arguments) == false {
```
groue: Have appendArguments return true if arguments are empty

GRDB/Core/SQLInterpolation.swift
```diff
@@ -0,0 +1,33 @@
...
+        sql = ""
+        sql.reserveCapacity(literalCapacity + interpolationCount)
+    }
+
+    /// "SELECT * FROM player"
+    public mutating func appendLiteral(_ sql: String) {
+        self.sql += sql
+    }
+
+    /// "SELECT * FROM \(sql: "player")"
```
groue: rawSQL:

PR #496, opened 2019-03-06
Fix crash when trying to perform fetchCount or ValueObservation.trackCount on joined query

### LINKED ISSUE #495: Hard failure when trying to perform fetchCount or ValueObservation.trackCount on joined query
### What did you do?

Was trying to perform a count observation using a QueryInterfaceRequest that involved a join.

To get a count that matched the query expressed.

### What happened instead?

Crash within the GRDB library at https://github.com/groue/GRDB.swift/blob/v3.6.2/GRDB/QueryInterface/QueryInterfaceQuery.swift#L247

### Environment

**GRDB flavor(s):** (GRDB, GRDBCipher, GRDBCustom?) GRDBCipher
**GRDB version:** 3.6.2
**Installation method:** (CocoaPods, SPM, manual?) Manual
**Xcode version:** 10.1
**Swift version:** 4.2
**macOS version running Xcode:** Mojave 10.14.3

### Demo Project

Example from a playground I had setup can be seen at https://pastebin.com/arCtuYm2 - the last query in that playground (line 95) would cause the crash. Variants with explicit aliases for both tables (must explicitly alias both) work correctly (even if it may not be the most standard SQL generated for the count query `SELECT COUNT(*) FROM (SELECT * FROM ....) `- but the variant without explicit table aliases crashes. Since it looks to be crashing on `alias!` I expect that behind the scenes you generate implicit TableAlias objects if not explicitly done so and for some reason that doesn't happen (correctly) in count variants.

groue:
This PR fixes #495 by @Marus.

### DIFF
--- a/GRDB/QueryInterface/SQLGenerationContext.swift
+++ b/GRDB/QueryInterface/SQLGenerationContext.swift
@@ -19,6 +19,7 @@ public struct SQLGenerationContext {
     
     /// Used for SQLSelectQuery.makeSelectStatement() and SQLSelectQuery.makeDeleteStatement()
     static func queryGenerationContext(aliases: [TableAlias]) -> SQLGenerationContext {
+        // Unique aliases, but with preserved ordering, so that we have stable SQL generation
         let uniqueAliases = aliases.reduce(into: [TableAlias]()) {
             if !$0.contains($1) {
                 $0.append($1)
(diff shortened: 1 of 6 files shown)


groue:
Something is wrong with Travis. Let's merge this one, and prepare a v3.6.3 release.

groue:
🚀 Shipped with GRDB 3.7.0

groue:
Merged into GRDB-4.0 branch

PR #505, opened 2019-03-22
Eager Loading of HasMany associations

groue:
🚧 

This PR "completes" the associations feature set by allowing eager loading of HasMany and HasManyThrough associations:

```swift
// Setup
struct Author: TableRecord, FetchableRecord, Decodable {
    static let books = hasMany(Book.self)
    var id: Int64
    var name: String
}

struct Book: TableRecord, FetchableRecord, Decodable {
    var id: Int64
    var authorId: Int64
    var title: String
}

// Load all authors along with their books
struct AuthorInfo: FetchableRecord, Decodable {
    var author: Author
    var books: [Book]
}

let request = Author.including(all: Author.books)
let infos: [AuthorInfo] = try AuthorInfo.fetchAll(db, request)

for info in infos {
    print("\(info.author.name) wrote:")
    for book in info.books {
        print("- \(book.title)")
    }
}
```


groue:
Closing, because I have no idea what I'm doing. We'll start over from #521.

PR #511, opened 2019-04-04
Fix notification name.

charlesmchen-signal:
Due to a typo, `DatabasePool` erroneously observes the same notification type.  I've fixed the typo.

Per your comment on #510, I've opened this against GRDB-4.0.

Just fixing a typo; no docs need to be changed.

Just fixing a typo; no docs need to be changed.

I haven't tested - the fix seems trivial?

### DIFF
--- a/GRDB/Core/DatabasePool.swift
+++ b/GRDB/Core/DatabasePool.swift
@@ -185,7 +185,7 @@ extension DatabasePool {
         center.addObserver(
             self,
             selector: #selector(DatabasePool.applicationDidEnterBackground(_:)),
-            name: UIApplication.didReceiveMemoryWarningNotification,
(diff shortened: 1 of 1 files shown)


GRDB/Core/DatabasePool.swift
```diff
@@ -185,7 +185,7 @@ extension DatabasePool {
         center.addObserver(
             self,
             selector: #selector(DatabasePool.applicationDidEnterBackground(_:)),
-            name: UIApplication.didReceiveMemoryWarningNotification,
```
charlesmchen-signal: Note the mismatch between this and the selector used on the previous line.  `DatabaseQueue` does something similar, but without the typo.

groue:
Well spotted 👍

Issue #513, opened 2019-04-08
Record.fetchOne(db) does not add "LIMIT 1" into sql query


ealymbaev:
### What did you do?

I did enable trace printing and found out that .fetchOne(db) query does not add "LIMIT 1" to raw sql statement

As long as I know there IS a difference for performance when you add limitation to query. FetchOne requires only one row.

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 3.7
**Installation method:** CocoaPods
**Xcode version:**
**Swift version:** 4.2
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:**

groue:
Hello @ealymbaev,

Thank you for your report. In my own personal usage of GRDB, I don't remember using `fetchOne` for queries which are not meant to return only one row. For example, when fetching by primary key. Or when fetching from a table which contains a single row.

Now, if you know actual use cases of `fetchOne` which means "the first in this set", then would you please share them here, so that we can discuss? There may be real needs for this extra `LIMIT 1`, after all.

alextrob:
I've been caught out by this as well.

My use case was having a lot of data cached in a table and I wanted to query what the most recently cached item was. It looked something like this:

```swift
let latest = try dbQueue.read { db in
    CacheRecord.order([Column("cachedAt").desc]).fetchOne(db)
}
```

Adding `.limit(1)` to the query is an easy fix but easy to overlook.

groue:
Hi @alextrob, thanks for your input, it's very valuable.

OK, we can look at a way to add an automatic `LIMIT 1` to `fetchOne`. The impact of this change will not be huge, but you were able to convince me we can avoid surprises 👍 

---

Currently, `fetchOne` in an extension method on the [FetchRequest](https://github.com/groue/GRDB.swift/blob/master/README.md#fetchrequest-protocol) protocol. This protocol does not know which fetching method is called when it generates SQL statements, so it can not, in its current shape, deliver what we need:

```swift
protocol FetchRequest {
    // Called for fetchOne, fetchAll, fetchCursor
    func prepare(_ db: Database) throws -> (SelectStatement, RowAdapter?)
}
```

FetchRequest is the base protocol for [custom requests](https://github.com/groue/GRDB.swift/blob/master/README.md#custom-requests) as well as the two built-in requests QueryInterfaceRequest, and SQLRequest:

```swift
[...]

PR #515, opened 2019-04-11
Add "LIMIT 1" to `fetchOne` requests

alextrob:
Adds "LIMIT 1" to `fetchOne` requests when appropriate. More discussion and reasoning about this is in #513.

"LIMIT 1" is _not_ added when querying by primary key or a unique key.

```swift
// SELECT * FROM player WHERE rowid = 1
try Player.fetchOne(key: 1)

// SELECT * FROM player WHERE email = "test@example.com"
// (assuming `email` is unique)
try Player.fetchOne(key: ["email": "test@example.com"])
```

But it is added in other cases where only one record needs to be fetched.

This PR includes a change to the `FetchRequest` protocol. `prepare` now has a `singleResult` parameter that conforming types can optionally use to generate a more suitable query.

```swift
protocol FetchRequest {
    func prepare(_ db: Database, forSingleResult singleResult: Bool) throws -> (SelectStatement, RowAdapter?)
}
```

For the built-in `FetchRequest` conformers:

* `QueryInterfaceRequest` uses `singleResult` to add (or not add) "LIMIT 1" to a query.
* `SQLRequest` ignores `singleResult` because when you opt in to raw SQL you probably don't expect GRDB to modify your query.

### Pull Request Checklist

- [x] <strike>This pull request is submitted against the `development` branch.</strike> (@groue [requested](https://github.com/groue/GRDB.swift/issues/513#issuecomment-481656845) that this target the `GRDB-4.0` branch).


Tests/GRDBTests/RecordMinimalPrimaryKeyRowIDTests.swift
```diff
@@ -380,6 +381,7 @@ class RecordMinimalPrimaryKeyRowIDTests : GRDBTestCase {
             
             let fetchedRecord = try MinimalRowID.filter(key: ["id": record.id]).fetchOne(db)!
             XCTAssertTrue(fetchedRecord.id == record.id)
+            XCTAssertEqual(lastSQLQuery, "SELECT * FROM \"minimalRowIDs\" WHERE (\"id\" = \(record.id!)) LIMIT 1")
```
groue: I'm very happy you added this test.

I'm somewhat sorry, because this technique for fetching one record by primary or unique key wasn't listed in my "exhaustive" list in #513:

```swift
let request = Player.filter(key: 123) // or Player.filter(key: ["email": "me@example.com"])
let player = try dbQueue.read { db in try request.fetchOne(db) }
```

For the context, this alternative way to perform `fetchOne(_:key:)` is useful because such requests can be observed so that the application is notified of db changes:

```swift
let observation = ValueObservation(trackingOne: Player.filter(key: 123))
try observation.start(in: dbQueue) { player: Player?
    print("Player has changed: \(player)")
} 
```

**What about trying to remove `LIMIT 1` here as well?**

One way to achieve this goal is to add a flag in SQLSelectQuery. When this flag is set, we could ignore the `limitOne` hint. And maybe we'd no longer need the `primaryKeyOrUnique` hint after that?

This would be a safe technique because all request derivation methods can only *increase* the amount of filtering, never relax it:

```swift
// Player 123, along with its team, but only if it is active:
//
// SELECT player.*, team.*
// FROM player
// JOIN team ON team.id = player.teamId
// WHERE player.id = 123 AND player.isActive = 1
let request = Player
    .filter(key: 123)
    .filter(Column("isActive") == true)
    .including(required: Player.team)
```

We'd need to make sure that all SQLSelectQuery derivation methods transfer this flag from one query to another. But after a quick glance, it looks like all those methods are already information-preserving.

groue: NB: this change is not required for this pull request to enter review. We can proceed without it if you prefer. But if you agree with the suggestion, and if you have a little more time, this change would be appreciated.

groue: Hint: look for `filter<PrimaryKeyType: DatabaseValueConvertible>(key:` and `filter(key:`.

GRDB/Core/FetchRequest.swift
```diff
@@ -15,10 +15,20 @@ public protocol FetchRequest: DatabaseRegionConvertible {
     /// Returns a tuple that contains a prepared statement that is ready to be
     /// executed, and an eventual row adapter.
     ///
+    /// Default implementation uses `prepare(db, hint: nil)`.
```
groue: Second question: what is the purpose of keeping this method in the protocol? Is there any reason for a type to implement both methods? Isn't `prepare(_:hint:)` sufficient?

Tests/GRDBTests/RecordMinimalPrimaryKeyRowIDTests.swift
alextrob: Good suggestion @groue, and thanks for the hints 😄 

Definitely agree it'd be good to remove LIMIT 1 from the SQL there. I can look into it.

GRDB/Core/FetchRequest.swift
alextrob: Because the protocol's `public` I had this idea of not wanting to break compatibility with anything already calling `prepare(_:)`. Also thought `hint` could be confusing when you don't need it.

I'd have preferred to declare it as:

```
func prepare(_ db: Database, hint: FetchRequestHint? = nil) throws -> (SelectStatement, RowAdapter?)
```

… but you can't have a default argument in a protocol.

Anyway, I'm not too attached to it. Happy to take it out if you'd prefer.

groue: All right, I understand. The good news is that we can break things, because we are targeting the next major version. GRDB 4 will break several things (see #479 for the list of changes). So we can keep a single method.

I agree the hint adds a layer of complexity to the API, but I don't expect many users to write their own concrete type which conforms to FetchRequest. This is an advanced technique. If they do, I guess they'll be able to handle it.

But we have to document clearly that this hint is only an opportunity for optimization, and can safely be ignored.

Towards this goal of clarity, I think we'll also have to look for alternative names for `hint` and `FetchRequestHint`, and wonder if we really want an optional. Those currently sound too much like an implementation detail, and not enough like a public API.

I don't know... The radical solution is:

```swift
func prepare(_ db: Database, forSingleResult singleResult: Bool) -> ... {
    if singleResult {
        // Add LIMIT 1
        return ...
    } else {
        // Don't add LIMIT 1
        return ...
    }
}
```
 
But a boolean has little room for further extensions (even if I can't foresee any right now). I agree with your idea of an enum:

```swift
func prepare(_ db: Database, forCardinality cardinality: FetchCardinality) -> ... {
    switch cardinality {
    case .single:
        // Add LIMIT 1
        return ...
    case .multiple:
        // Don't add LIMIT 1
        return ...
    }
}
```

But "cardinality" can't really be extended either. Not really an improvement over a boolean. Let's go all general:

```swift
func prepare(_ db: Database, forPurpose purpose: FetchPurpose) -> ... {
    switch purpose {
    case .singleResult:
        // Add LIMIT 1
        return ...
    case .multipleResults:
        // Don't add LIMIT 1
        return ...
    }
}
```

I wish we could have merged `FetchRequest.fetchCount` here:

```swift
func prepare(_ db: Database, forPurpose purpose: FetchPurpose) -> ... {
    switch purpose {
    case .singleResult:
        // Add LIMIT 1
        return ...
    case .multipleResults:
        // Don't add LIMIT 1
        return ...
    case .count:
        // Build the counting statement
        return ...
    }
}
```

But we can't choose this last route, because `fetchCount` is currently an optional method with a non-trivial default implementation: users would now have real difficulties adopting the FetchRequest protocol.

Well.. As you see, we have much room in our way to the best design. I'm curious about your ideas as well. Our goal: turn our implementation strategy into a public API that can make sense for people who **don't have a clue** about our implementation strategy. I mean that most users are utterly indifferent about how the library works, and only care about their needs. Good public APIs acknowledge this hard fact. It may look difficult, but this can also be a funny game :-)

A technique that has constantly helped me finding an answer to this kind of question has been the README. We currently write, below the introduction to the FetchRequest protocol:

> The `prepare` method returns a prepared statement and an optional row adapter. The [prepared statement](#prepared-statements) tells which SQL query should be executed. The row adapter helps presenting the fetched rows in the way expected by the row decoders (see [row adapter](#row-adapters)).

We can try to update this paragraph so that it explains how to use the optional hint. Experience shows that when the documentation looks like it fills its teaching goal, the API is good as well.

Tests/GRDBTests/RecordMinimalPrimaryKeyRowIDTests.swift
groue: Cool, thank you!

GRDB/Core/FetchRequest.swift
groue: Applying my own advice makes the plain and simple boolean approach look not so bad:

> The `prepare(_:forSingleResult:)` method accepts a database connection, a `singleResult` hint, and returns a prepared statement and an optional row adapter. The `singleResult` hint can safely be ignored, but conforming types can use it as an opportunity to append `LIMIT 1` to the returned [prepared statement](#prepared-statements). The optional row adapter helps presenting the fetched rows in the way expected by the row decoders (see [row adapter](row-adapters)).

GRDB/QueryInterface/QueryInterfaceRequest.swift
```diff
@@ -39,9 +43,19 @@ extension QueryInterfaceRequest : FetchRequest {
...
+    /// - parameter hint: A hint about how the query should be prepared.
     /// - returns: A prepared statement and an eventual row adapter.
     /// :nodoc:
-    public func prepare(_ db: Database) throws -> (SelectStatement, RowAdapter?) {
+    public func prepare(_ db: Database, hint: FetchRequestHint?) throws -> (SelectStatement, RowAdapter?) {
+        let query: SQLSelectQuery
+
+        switch hint {
+        case .limitOne?:
+            query = self.query.limit(1)
```
groue: Here we break the following code:

```swift
let player = try Player.limit(1, offset: 10).fetchOne(db)
```

We have to honor the eventual offset, if present.

GRDB/Core/FetchRequest.swift
alextrob: I'm coming around to the idea of a plain old boolean. Complicating the API for unknown future cases probably isn't worth it. If other scenarios do come up in the future, it can be changed in a way that really makes sense.

groue: > Complicating the API for unknown future cases probably isn't worth it.

Indeed, if we really can't find any extension for the "hint" today, then let's not try to outsmart ourselves. As you say, we'll always be able to revise our opinion in the future, if needed!

Tests/GRDBTests/RecordMinimalPrimaryKeyRowIDTests.swift
alextrob: Ok I think I've got that implemented as you described it. Tests are updated to exclude LIMIT 1 from the SQL in that scenario.

GRDB/Core/FetchRequest.swift
alextrob: Cool. I've switched it to a boolean and updated the README.

GRDB/QueryInterface/QueryInterfaceRequest.swift
alextrob: Whoops! Should be fixed in the latest commits (and added a test to catch it).

alextrob:
Ready for another look @groue 😊

groue:
@alextrob, this is really great.

I added a couple commits:

- 46791c8 remove an extra initializer which was not needed.
- 4716f91 makes `filter(key:)` sole responsible for the expectsSingleResult flag. I understand that you did not "dare" going this route, but I think this is better now: other types and protocols remain totally unaware of this flag.

Now I'll look at tests, which I'm sure are excellent, and we'll merge this PR very shortly :-)

alextrob:
Nice one. I like your changes @groue 👍

groue:
@alextrob, this looks very good. I have removed tests about the lack of extra `LIMIT 1` for raw sql requests, because it's more a matter of implementation difficulty than a hard choice. I'll now just add a couple more tests that fill some holes. You can consider your part as done, and well done on top of that!

groue:
Merged. Thank you very much @alextrob and @ealymbaev: I did not immediately recognize the value of this change, but I'm now totally convinced.

Alex, I'd be happy to grant you push access to the repository. Drop me a line if you are interested!

alextrob:
Thanks @groue! Happy I got to contribute 😄

groue:
Invitation sent 🤝

Issue #530, opened 2019-05-13
Wrong number of statement arguments with "like '%?%'"


Timac:
### What did you do?

I am trying to perform some SQLite queries using the like operator.
For example I would like to delete all rows from mytable where the title column contains the string "something":

```
do {
	try databasePool.write { db in
		try db.execute(sql: "DELETE FROM mytable where title like '%?%'", arguments: ["something"])
	}
} catch let error as DatabaseError {
	debugPrint("Failed with error \(error)")
} catch {
	debugPrint("Failed with unknown error")
}
```

### What did you expect to happen?

The SQLite query looks valid. I would expect the query to delete all rows from mytable where the substring is found.

### What happened instead?

I see instead an error `SQLite error 21: wrong number of statement arguments: 1`
Note that if I use instead the following, I don't see the error and the expected rows are deleted:

```
try db.execute(sql: "DELETE FROM mytable where title like '%\(something)%'")
```

A similar issue occurs when using SELECT:

```
try db.execute(sql: "SELECT * FROM mytable where title like '%?%'", arguments: ["something"])
```

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 4.0 (revision 0e2daff329040ddf8694cdd1be1bc3e5ba0d4138)
**Installation method:** manual
**Xcode version:** 10.2.1
**Swift version:** 5
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 10.14.4

groue:
Please see https://github.com/Nike-Inc/SQift/issues/9 for an identical issue. It is a misunderstanding of what SQLite parameters are.

Timac:
@groue Thanks a lot for the clarification!
Since I am not the first one asking this question and probably not the last one, I added a paragraph to the README to clarify this point.

See https://github.com/groue/GRDB.swift/pull/531

groue:
That's lovely, @Timac! You're right, this is a FAQ :-) I'll soon look into your PR.

sobri909:
It also works if you put the percent marks in the argument. 

```swift
let notes = Stalker.store.notes(
    where: "body LIKE ? AND deleted = 0 ORDER BY date DESC",
    arguments: ["%\(query)%"])
```

PR #531, opened 2019-05-14
Add documentation to clarify #530 Wrong number of statement arguments with "like '%?%'"

Timac:
Add a paragraph "? as a parameter" in the "Good to Know" section of the README to explain how to build like pattern like '%?%'.

See related issue:
https://github.com/groue/GRDB.swift/issues/530

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,7 @@ Documentation
 #### Good to Know
 
 - [Avoiding SQL Injection](#avoiding-sql-injection)
+- [? as a parameter](#-as-a-parameter)
 - [Error Handling](#error-handling)
 - [Unicode](#unicode)
 - [Memory Management](#memory-management)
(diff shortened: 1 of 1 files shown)


groue:
@Timac,

Thank you for your contribution.

I'm not sure we can publish it right away, considering how things are presented:

1. The new documentation section is all clear given the context of #530. But we can't expect readers to know about this kind of issue before they meet it. I'm afraid most people will miss your goal because they don't know what your goal is and why they should be concerned.
2. User who face the "Wrong number of statement arguments" error will Google for it. They won't be able to find your new documentation section because it does not mention the *verbatim error*.
3. You provide one workaround for a particular case, without exposing other valid workarounds (such as building the whole parameter in Swift, instead of building it with SQL)

May I suggest you rewrite it as a FAQ entry instead, focused on the "Wrong number of statement arguments" error. Introduce the error with a sample erroneous LIKE query. Then explain what this error means and what `?` is. As I did in https://github.com/Nike-Inc/SQift/issues/9, link to SQLite documentation. Make your readers able to *learn* when they read your documentation, not only find a *quick fix*. Quick fixes are for Stack Overflow, not for GRDB documentation.

groue:
@Timac, I close this PR due to lack of activity. Don't hesitate reopening when you are ready!

Timac:
Hey @groue,

Sorry for the delay. I definitively agree with all the points you wrote.
I made a new pull request to the `master` branch which should follow your suggestions.

See: https://github.com/groue/GRDB.swift/pull/538

PR #538, opened 2019-05-27
Add FAQ to clarify "Wrong number of statement arguments" error with "like '%?%'"

Timac:
Add an FAQ about the SQLite error 21 "wrong number of statement arguments" with LIKE queries. The documentation contains an example of a LIKE query exhibiting the issue as well as explanation about the problem and a workaround.

See related issue:
#530


README.md
```diff
@@ -289,6 +289,7 @@ Documentation
 #### Good to Know
 
 - [Avoiding SQL Injection](#avoiding-sql-injection)
+- [? as a parameter](#-as-a-parameter)
```
groue: I guess this line can be deleted.

README.md
```diff
@@ -8506,6 +8507,7 @@ FAQ
 - [Generic parameter 'T' could not be inferred](#generic-parameter-t-could-not-be-inferred)
 - [SQLite error 10 "disk I/O error", SQLite error 23 "not authorized"](#sqlite-error-10-disk-io-error-sqlite-error-23-not-authorized)
 - [What Are Experimental Features?](#what-are-experimental-features)
+- [SQLite error 21 "wrong number of statement arguments" with LIKE queries](#sqlite-error-21-wrong-number-of-statement-arguments-with-like-queries)
```
groue: You will focus the "wrong number of statement arguments" error on LIKE queries only. All right 👍

README.md
```diff
@@ -8645,6 +8647,29 @@ There is an exception, though: *experimental features*, marked with the "**:fire
...
 
 
+### SQLite error 21 "wrong number of statement arguments" with LIKE queries
+
+You may get the error "wrong number of statement arguments" when executing a LIKE query similar to:
+
+```swift
+let name = textField.text
+try dbQueue.read { db in
+    try db.execute(sql: "SELECT * FROM students where title like '%?%'", arguments: [name])
```
groue: One does not use `execute` with SELECT queries. I suggest instead:

```swift
let name = textField.text
let players = try dbQueue.read { db in
    try Player.fetchAll(db, sql: "SELECT * FROM player WHERE name LIKE '%?%'", arguments: [name])
}
```

Stylewise, note:

- Usage of the Player type, one of the few example types used in GRDB documentation.
- Uppercase SQL keywords, as everywhere in GRDB documentation.
- Singular table name, as [recommended](https://github.com/groue/GRDB.swift/blob/master/README.md#create-tables).

README.md
```diff
@@ -8645,6 +8647,29 @@ There is an exception, though: *experimental features*, marked with the "**:fire
...
+}
+```
+
+What is important to understand here is that SQLite only interprets `?` as a parameter when it is a placeholder for a whole value (int, double, string, blob, null). In this incorrect query, `?` is just a character in the `'%?%'` string value: it is not a query parameter, and is not processed in any way. See [https://www.sqlite.org/lang_expr.html#varparam](https://www.sqlite.org/lang_expr.html#varparam) for more information about SQLite parameters.
+
+To provide the parameter, one option is to let SQLite build the like pattern using the string concatenation operator `||`:
```
groue: Since you provide a single example of code that works, then please provide the one that lets Swift build the parameter instead:

```swift
let name = textField.text
let players = try dbQueue.read { db in
    let pattern = "%\(name)%"
    try Player.fetchAll(db, sql: "SELECT * FROM player WHERE name LIKE ?", arguments: [pattern])
}
```

Why so? Because it can trivially be ported to the query interface when the user decides to use it:

```swift
let name = textField.text
let players = try dbQueue.read { db in
    let pattern = "%\(name)%"
    try Player.filter(Column("name").like(pattern)).fetchAll(db)
}
```

You don't have to mention the query interface request in your FAQ. But we have to write documentation in a way that smoothes the transition from raw SQL to generated SQL.

Timac:
Hey @groue ,

I did the changes as requested. I also change the FAQ to provide both solutions:
- letting Swift build the pattern (with a mention about the query interface)
- using the || operator

I think that having these different examples might help future readers.

Let me know if you need any other changes. Thanks again for providing such a great framework!

README.md
Timac: This line has been removed.

groue:
OK, @Timac, I slightly modified the copy, removed the mention of the query interface which is not the topic here, and removed the second solution because one solution is all that people are looking for here.

Thank you very, very much 👍

groue:
By the way, welcome to the GRDB contributors! 🤝

Just drop me a line if you'd like to be granted with a push access to the repository.

Timac:
@groue Happy to have helped!

PR #541, opened 2019-05-30
ValueObservation.fetch

groue:
This PR adds the `fetch` and `fetchFirst` methods to [ValueObservation](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservation).

`fetch` returns the observed value:

```swift
let observation = ValueObservation.trackingAll(Player.all())
let players: [Player] = try dbQueue.read { db in
    try observation.fetch(db)
}
```

Not all observations are guaranteed to emit values. The `compactMap` operator, for example, can mute an observation. This is why the `fetch` method is not available on all observations.

> Implementation-wise, `fetch` is defined on observations whose reducer conforms to the ImmediateValueReducer protocol. CompactMapValueReducer does not conform to this protocol.

When `fetch` is not available, one can still use `fetchFirst`, which returns an optional:

```swift
let observation = ValueObservation
    .trackingOne(Player.filter(key: 42))
    .compactMap { $0 } // filters out non-existing player
let player: Player? = try dbQueue.read { db in
    try observation.fetchFirst(db)
}
```

Until we figure out if those new ValueObservation `fetch` and `fetchFirst` methods are actually sound and interesting, they will be marked [**:fire: EXPERIMENTAL**](http://github.com/groue/GRDB.swift/blob/master/README.md#what-are-experimental-features), and remain undocumented in the main README.

They are an experiment inspired by [Room](https://developer.android.com/topic/libraries/architecture/room) the Android database library. In Room, the developer can easily expose APIs that returns values, LiveData, or RxJava Observables ([documentation](https://developer.android.com/training/data-storage/room/accessing-data)):

```kotlin
@Dao
interface PlayerDao {
    // Simple queries
    @Query("SELECT * FROM player")
    fun loadAllPlayers(): List<Player>
    
    // Observable queries with LiveData
    @Query("SELECT * FROM player")
    fun loadAllPlayers(): LiveData<List<Player>>
    
    // Reactive queries with RxJava
    @Query("SELECT * FROM player")
    fun loadAllPlayers(): Flowable<List<Player>>
)
```

I want to see if ValueObservation can bring this versatility to GRDB and [RxGRDB](https://github.com/RxSwiftCommunity/RxGRDB) as well:

```swift
extension PlayerManager {
    // Define one observation
    private let playersObservation = ValueObservation.trackingAll(Player.all())
    
    // Use it to fetch
    func players() throws -> [Player] {
        return try dbQueue.read { db in
            try playersObservation.fetch(db)
        }
    }
    
    // Use it with RxGRDB
    func players() -> Observable<[Player]> {
        return playersObservation.rx.fetch(in: dbQueue)
    }
}
```

ValueObservation is a versatile type, thanks to its `map`, `combine`, etc. [transformation methods](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservation-transformations). ValueObservation has a great potential:

```swift
extension PlayerManager {
    struct HallOfFame {
        var totalPlayerCount: Int
        var bestPlayers: [Player]
    }
    
    private let hallOfFameObservation =
        ValueObservation
            .combine(
                ValueObservation.trackingCount(Player.all()),
                ValueObservation.trackingAll(Player
                    .order(Column("score").desc)
                    .limit(10)))
            .map { totalPlayerCount, bestPlayers in
                HallOfFame(
                    totalPlayerCount: totalPlayerCount,
                    bestPlayers: bestPlayers)
            }
    
    // Use it to fetch
    func hallOfFame() throws -> HallOfFame {
        return try dbQueue.read { db in
            try hallOfFameObservation.fetch(db)
        }
    }
    
    // Use it with RxGRDB
    func hallOfFame() -> Observable<HallOfFame> {
        return hallOfFameObservation.rx.fetch(in: dbQueue)
    }
}
```

There are several problems:

- The type of PlayerManager.hallOfFameObservation is an awful behemoth: ` ValueObservation<DistinctUntilChangedValueReducer<MapValueReducer<CombinedValueReducer2<DistinctUntilChangedValueReducer<RawValueReducer<Int>>, FetchableRecordsReducer<Player>>, PlayerManager.HallOfFame>>>`.
    
    Because of their complex types, observations can not easily be used in contexts where Swift requires explicit types (functions, computed properties...).
    
    And we thus can not build complex observations step by step.
    
    This is why the sample code above defines a huge expression. Difficult to build. Difficult to debug. Difficult to read. It just looks too smart to be good. And its inner components can not be reused elsewhere.
    
    We may have to wait for [SE-0244 Opaque Result Types](https://github.com/apple/swift-evolution/blob/master/proposals/0244-opaque-result-types.md) in Swift 5.1 in order to address the root problem of the observation types.

- ValueObservation apis are heavy and verbose. We need to bring a little taste there :sweat_smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one expection: 
 - [#537](https://github.com/groue/GRDB.swift/pull/537): Remove useless parenthesis from generated SQL
 - [#538](https://github.com/groue/GRDB.swift/pull/538) by [@Timac](https://github.com/Timac): Add FAQ to clarify "Wrong number of statement arguments" error with "like '%?%'"
 - [#539](https://github.com/groue/GRDB.swift/pull/539): Expose joining methods of both requests and associations
+- [#541](https://github.com/groue/GRDB.swift/pull/541): ValueObservation.fetch
 
 ### Documentation Diff
 
--- a/GRDB.xcodeproj/project.pbxproj
+++ b/GRDB.xcodeproj/project.pbxproj
@@ -482,6 +482,8 @@
 		567F45AC1F888B2600030B59 /* TruncateOptimizationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 567F45A71F888B2600030B59 /* TruncateOptimizationTests.swift */; };
 		568068311EBBA26100EFB8AA /* SQLRequestTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 568068301EBBA26100EFB8AA /* SQLRequestTests.swift */; };
 		568068351EBBA26100EFB8AA /* SQLRequestTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 568068301EBBA26100EFB8AA /* SQLRequestTests.swift */; };
+		5686050C22A01A9700CD4C12 /* ValueObservationTrackingFetchTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5686050B22A01A9700CD4C12 /* ValueObservationTrackingFetchTests.swift */; };
+		5686050D22A01A9700CD4C12 /* ValueObservationTrackingFetchTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5686050B22A01A9700CD4C12 /* ValueObservationTrackingFetchTests.swift */; };
 		568735A21CEDE16C009B9116 /* Betty.jpeg in Resources */ = {isa = PBXBuildFile; fileRef = 5687359E1CEDE16C009B9116 /* Betty.jpeg */; };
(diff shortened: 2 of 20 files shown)


groue:
@hartbit, may I ask for your opinion on this one? You're very good at vetting APIs: I wonder what your eagle eye would say about the whole picture here. I'm looking for advice without fear of sharp chirurgy.

hartbit:
@groue I'm on my way to San Jose for WWDC 😄 I'll look at this Sunday. Is that okay?

groue:
Take your time David! Actually I'm not sure it was a good idea to ping you: I did not take the time to explain much, and you may wonder what I'm after. But I'll appreciate your feedback anyway. Surprises are always cool :-)

groue:
Happy WWDC 😉

groue:
Closing. I've imported the good bits in the development branch, but this idea is just not ready yet. We'll see what the release of Combine and Swift 5.1 brings. We can expect improvements to RxGRDB as well.

PR #550, opened 2019-06-13
Asynchronous Database Access Methods

groue:
This PR adds asynchronous database access methods on DatabaseQueue, Pool, and Snaphot:

```swift
protocol DatabaseReader {
    #if compiler(>=5.0)
    func asyncRead(_ block: @escaping (Result<Database, Error>) -> Void)
    #endif
}

protocol DatabaseWriter {
    #if compiler(>=5.0)
    func asyncWrite<T>(
        _ updates: @escaping (Database) throws -> T,
        completion: @escaping (Database, Result<T, Error>) -> Void)
    #endif

    func asyncWriteWithoutTransaction(_ updates: @escaping (Database) -> Void)
}
```

`asyncWrite` and `asyncRead` use the Result type from the Standard Library, and are only available with a Swift 5+ compiler (Xcode 10.2+).

I'm worried that `asyncWriteWithoutTransaction` is easier to use than the safer alternative `asyncWrite`. Let's hope the very long name sounds like a warning for the users.

Those methods will fuel RxGRDB and GRDBCombine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one expection: 
 - [#542](https://github.com/groue/GRDB.swift/pull/542): Move eager loading of hasMany associations to FetchRequest
 - [#546](https://github.com/groue/GRDB.swift/pull/546) by [@robcas3](https://github.com/robcas3): Fix SPM errors with Xcode 11 beta
 - [#549](https://github.com/groue/GRDB.swift/pull/549) Support for Combine
+- [#550](https://github.com/groue/GRDB.swift/pull/550) Asynchronous Database Access Methods
 
 ### Documentation Diff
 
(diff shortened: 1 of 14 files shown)


groue:
Yeah @sobri909, it took years of iterations, and the advent of the standard `Result`, before we could shape the asynchronous APIs 😅

sobri909:
@groue I didn't even know this was in the works! Very cool surprise 😄

Also, apologies for being absent for the past many months. My workload has been insane, and drowned out my chances to work on anything else. But I do still read every GRDB email 😂 There's always fun new things turning up! I just never have the time to start using them or work on them. Sigh.

I'm keen to have a play with `GRDBCombine` though, once I get a chance. That looks extremely promising!

groue:
Don't worry :-) I'm having a lot of fun, and it's a pleasure to share it 👍

PR #553, opened 2019-06-15
Complete Reactive Support

groue:
This PR adds missing pieces for the companion libraries RxGRDB and GRDBCombine:

- The [experimental](https://github.com/groue/GRDB.swift/blob/master/README.md#what-are-experimental-features) `ImmediateValueReducer` protocol guarantees an Observable/Publisher that a ValueObservation emits an initial value. It makes it possible to define a Single/Future. All value observations have this quality, but the ones that are filtered with `compactMap`.

- Optional values were lacking the fetchOne method: `Optional<String>.fetchOne(...)` was not available because `String.fetchOne(...)` was just enough. This lack of uniformity was a hindrance for companion libraries that are highly generic and don't really care about such detail. So now optional values have the fetchOne method, with the exact same behavior as the non-optional one: all return nil if there was no row to fetch, or if the row contains a NULL value.

### DIFF
--- a/GRDB/Core/DatabaseSnapshot.swift
+++ b/GRDB/Core/DatabaseSnapshot.swift
@@ -127,7 +127,7 @@ extension DatabaseSnapshot {
             // Deal with initial value
             switch observation.scheduling {
             case .mainQueue:
-                if let value = try unsafeReentrantRead(observation.fetchFirst) {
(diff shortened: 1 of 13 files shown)


groue:
No, everything is ill-advised or too early.

PR #562, opened 2019-06-28
Fix crash when using more than one DatabaseCollation

### LINKED ISSUE #561: Crash using more than one DatabaseCollation
### What did you do?

Attempt to use a `DatabaseCollation` after installing multiple.

Using a custom collation successfully calls the collation function.

### What happened instead?

We see a bad access crash

### Environment

**GRDB flavor(s):** GRDB (linking custom SQLite)
**GRDB version:** 4.1.1
**Xcode version:** 10.2.1
**Swift version:** 5
**macOS version running Xcode:** 10.14.5

### Demo Project

No demo project, but here is what the code looks like:

```swift
func installCollation(_ pool: DatabasePool) {
    let standardCollation = DatabaseCollation("STANDARD") {
        $0.localizedStandardCompare($1)
    }
    let unicodeCollation = DatabaseCollation("UNICODE") {
        $0.localizedCaseInsensitiveCompare($1)
    }
    pool.add(collation: standardCollation)
    pool.add(collation: unicodeCollation)
}
```

### Investigative Notes
After the second call to `add(collation:`, `DatabasePool`'s internal collation `Set` only has one element. I would expect it to have 2. I believe this is because `DatabaseCollation`'s hash function always hashes to 0, which results in every call to `add` ejecting the previous collation from the set. If you then go to use one of the collations, which has been ejected, it will crash attempting to use the collation function from previously-ejected `DatabaseCollation` instances, as they have been deallocated (no longer retained by the `Set`).

I can workaround this by storing these collations in globals (ensuring the instances remain valid for the lifetime of the app):
```swift
private let standardCollation = DatabaseCollation("STANDARD") {
    $0.localizedStandardCompare($1)
}

private let unicodeCollation = DatabaseCollation("UNICODE") {
    $0.localizedCaseInsensitiveCompare($1)
}

func installCollation(_ pool: DatabasePool) {
    pool.add(collation: standardCollation)
    pool.add(collation: unicodeCollation)
}
```

groue:
This PR fixes #561.

cc @benasher44

### DIFF
--- a/GRDB/Core/DatabaseCollation.swift
+++ b/GRDB/Core/DatabaseCollation.swift
@@ -50,6 +50,6 @@ extension DatabaseCollation: Hashable {
     /// :nodoc:
     public static func == (lhs: DatabaseCollation, rhs: DatabaseCollation) -> Bool {
         // See https://www.sqlite.org/c3ref/create_collation.html
-        return sqlite3_stricmp(lhs.name, lhs.name) == 0
+        return sqlite3_stricmp(lhs.name, rhs.name) == 0
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@ All notable changes to this project will be documented in this file.
 
 GRDB adheres to [Semantic Versioning](https://semver.org/), with one expection: APIs flagged [**:fire: EXPERIMENTAL**](README.md#what-are-experimental-features). Those are unstable, and may break between any two minor releases of the library.
 
-<!--
+
 [Next Release](#next-release)
--->
+
 
 #### 4.x Releases
 
@@ -51,9 +51,11 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one expection: 
 
 - [0.110.0](#01100), ...
 
-<!--
+
 ## Next Release
--->
+
+- [#562](https://github.com/groue/GRDB.swift/pull/537): Fix crash when using more than one DatabaseCollation
+
 
 ## 4.1.1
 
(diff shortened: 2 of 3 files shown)


GRDB/Core/DatabaseCollation.swift
```diff
@@ -50,6 +50,6 @@ extension DatabaseCollation: Hashable {
     /// :nodoc:
     public static func == (lhs: DatabaseCollation, rhs: DatabaseCollation) -> Bool {
         // See https://www.sqlite.org/c3ref/create_collation.html
-        return sqlite3_stricmp(lhs.name, lhs.name) == 0
+        return sqlite3_stricmp(lhs.name, rhs.name) == 0
```
benasher44: Why not also hash on the name, instead of 0?

groue: Because we have to honor the Hashable/Equatable rule: if a equals b, then a and b must have the same hash.

SQLite only defines collation equality in https://www.sqlite.org/c3ref/create_collation.html:

> Collation names that compare equal according to [sqlite3_strnicmp()](https://www.sqlite.org/c3ref/stricmp.html) are considered to be the same name.

I don't know exactly how this C function is implemented. I also know that native Swift String comparisons handles unicode in a way that most certainly does not match the SQLite comparison.

That's why DatabaseCollation implementation of `==` uses the C SQLite function.

And the easy way to assert the Hashable/Equatable rule is to have all values have the same hash :-)

groue: A classic misinterpretation of Set and Dictionary is that they consider two values or keys that have the same hash to be the same. This is not true! Hashing is only an optimization. When all values have the same hash, sets and dictionaries are not efficient. But the invariant remain: values and keys are uniqued according to Equatable, not Hashable.

In our specific case, we don't have to expect many operations on collations using GRDB apis. So we don't care if all collations have the same hash.

benasher44:
Thanks!

PR #568, opened 2019-07-12
Fix Swift version in projects/targets

zdnk:
For Carthage and manual installation users, update Swift version in the targets to Swift 5

### Pull Request Checklist

- [x] Changes are tested.


groue:
Hello @zdnk, I can't accept this PR, I'm sorry.

This library attempts at not introducing breaking changes in major versions, and this PR would break GRDB 4. The documented requirements are Xcode 10+ and Swift 4.2+. And one of the documented installation methods is the "manual installation" where one drags the GRDB project inside one's project. So the GRDB projects *must* use Swift 4.2.

Now you'll also read in the documentation and various issues that Carthage is *not* a supported installation method. It's important to understand, as a Carthage user, that you're in charge of most of your setup. Libraries that adhere to semantic versioning just can't reasonably deal with all the various needs of Carthage users - Swift version, compiler version, prebuilt binaries, etc etc. By choosing Carthage, you chose to live on the edge: you have to maintain your own GRDB fork.

Issue #583, opened 2019-07-31
Question: Snapshot cursor iteration fails


Kirow:
### What did you do?

I've made playground to check snapshot cursor iteration while editing database and got issue I can not explain or understand. It could be SQLite limitation or GRDB issue.

playground setup:
```swift
import GRDB
import PlaygroundSupport

let url = playgroundSharedDataDirectory.appendingPathComponent("GRDB")
try! FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
try? FileManager.default.removeItem(at: url.appendingPathComponent("database.sqlite-wal"))
try? FileManager.default.removeItem(at: url.appendingPathComponent("database.sqlite-shm"))
try? FileManager.default.removeItem(at: url.appendingPathComponent("database.sqlite"))
let databasePath = url.appendingPathComponent("database.sqlite").path
```

database models:
```swift
public struct User: Codable, TableRecord, FetchableRecord, MutablePersistableRecord {
    public static let databaseTableName: String = "user"
    
    public var id: Int64?
    public var username: String
    public var isFlagged: Bool
    
    init(id: Int64? = nil, username: String, isFlagged: Bool = false) {
        self.id = id
        self.username = username
        self.isFlagged = isFlagged
    }
    
    public mutating func didInsert(with rowID: Int64, for column: String?) {
        id = rowID
    }
}

public struct FlagUser: Codable, TableRecord, FetchableRecord, MutablePersistableRecord {
    public static let databaseTableName: String = "flagUser"
    
    public var username: String
}
```

GRDB setup:
```swift
var configuration = Configuration()
configuration.trace = { print($0) }

var migrator = DatabaseMigrator()

migrator.registerMigration("initial") { database in
    try database.create(table: User.databaseTableName) { definition in
        definition.column("id", .integer).primaryKey(autoincrement: true)
        definition.column("username", .text).notNull()
        definition.column("isFlagged", .boolean).notNull().defaults(to: false)
    }
    
    try database.create(table: FlagUser.databaseTableName) { definition in
        definition.column("username", .text).notNull()
    }
}

migrator.registerMigration("data") { database in
    [Int](0...50).forEach {
        var user = User(username: "User\($0)")
        try! user.insert(database)
    }
    
    [Int](40...60).forEach {
        var flag = FlagUser(username: "User\($0)")
        try! flag.insert(database)
    }
}

let pool = try! DatabasePool(path: databasePath, configuration: configuration)
try migrator.migrate(pool)
[...]

PR #585, opened 2019-08-01
Fix use of SQLite authorizers

groue:
This pull request fixes an SQLite error 516 ([SQLITE_ABORT_ROLLBACK](https://www.sqlite.org/rescode.html#abort_rollback)) that can happen during the iteration of some database cursors, if other database statements are compiled during iteration steps. See #583 for a detailed description.

The issue does not affect all cursors, but only some of them. It actually looks like an issue in SQLite itself (see [bug report](http://sqlite.1065341.n5.nabble.com/Issue-report-sqlite3-set-authorizer-triggers-error-4-516-SQLITE-ABORT-ROLLBACK-during-statement-itern-td107972.html)).


groue:
After D. Richard Hipp has found [where the issue lies](http://sqlite.1065341.n5.nabble.com/Issue-report-sqlite3-set-authorizer-triggers-error-4-516-SQLITE-ABORT-ROLLBACK-during-statement-itern-tp107972p107973.html), I better understand that the way GRDB uses SQLite authorizers errs on the side of the misuse. A better fix than this PR, which would not introduce any breaking change, would be to refactor GRDB around the use of a single authorizer per connection.

groue:
Forced-puushed with an all new technique that avoids any breaking change.

PR #599, opened 2019-08-21
Don't keep the sqlCipher passphrase in memory longer than necessary

michaelkirk-signal:
Pull Request Checklist
- [x] This pull request is submitted against the development branch.
- [x] Changes are tested.

Helps minimize the time the passphrase is stored in application memory

### DIFF
--- a/GRDB/Core/Configuration.swift
+++ b/GRDB/Core/Configuration.swift
@@ -79,11 +79,29 @@ public struct Configuration {
     // MARK: - Encryption
     
     #if SQLITE_HAS_CODEC
-    /// The passphrase for the encrypted database.
+    /// A block which returns the passphrase for the encrypted database.
+    ///
+    /// If encrypting, you should use either this _or_ `passphrase`.
+    /// Using both is an error.
     ///
     /// Default: nil
-    public var passphrase: String?
-    
+    public var passphraseBlock: (() throws -> String)? {
+        didSet {
+            assert(passphraseBlock == nil || passphrase == nil, "cannot use both passphrase *and* passphraseBlock")
+        }
+    }
+
+    /// The passphrase for the encrypted database.
+    ///
+    /// If encrypting, you should use either this _or_ `passphraseBlock`.
+    /// Using both is an error.
+    ///
+    /// Default: nil
+    public var passphrase: String? {
+        didSet {
+            assert(passphraseBlock == nil || passphrase == nil, "cannot use both passphrase *and* passphraseBlock")
+        }
+    }
     #endif
     
     /// If set, allows custom configuration to be run every time
--- a/GRDB/Core/Database.swift
+++ b/GRDB/Core/Database.swift
@@ -213,7 +213,7 @@ extension Database {
         
         #if SQLITE_HAS_CODEC
         try validateSQLCipher()
-        if let passphrase = configuration.passphrase {
+        if let passphrase = try configuration.passphraseBlock?() ?? configuration.passphrase {
             try setCipherPassphrase(passphrase)
         }
         #endif
--- a/Tests/GRDBTests/EncryptionTests.swift
+++ b/Tests/GRDBTests/EncryptionTests.swift
@@ -23,6 +23,27 @@ class EncryptionTests: GRDBTestCase {
         }
     }
 
+    func testDatabaseQueueWithPassphraseToDatabaseQueueWithPassphraseBlock() throws {
+        do {
+            dbConfiguration.passphrase = "secret"
+            dbConfiguration.passphraseBlock = nil
+            let dbQueue = try makeDatabaseQueue(filename: "test.sqlite")
+            try dbQueue.inDatabase { db in
+                try db.execute(sql: "CREATE TABLE data (value INTEGER)")
+                try db.execute(sql: "INSERT INTO data (value) VALUES (1)")
+            }
+        }
+
+        do {
+            dbConfiguration.passphrase = nil
+            dbConfiguration.passphraseBlock = { return "secret" }
+            let dbQueue = try makeDatabaseQueue(filename: "test.sqlite")
+            try dbQueue.inDatabase { db in
+                XCTAssertEqual(try Int.fetchOne(db, sql: "SELECT COUNT(*) FROM data")!, 1)
+            }
+        }
+    }
+
     func testDatabaseQueueWithPassphraseToDatabaseQueueWithoutPassphrase() throws {
         do {
             dbConfiguration.passphrase = "secret"


groue:
Hello @michaelkirk-signal,

I don't understand the purpose of this pull request.

groue:
I'm closing this pull request, @michaelkirk-signal.

It is not motivated, and, frankly, would have never been merged. Immutability and protection from side effects have benefits. Your request is no different from adding an `httpBodyBlock` to URLRequest, or a `dateStyleBlock` to DateFormatter. Looks odd? Yes it is. If you need a block in order to defer the definition of the passphrase, defer the definition of the Configuration instead.

Thanks!

groue:
Thanks @michaelkirk-signa, it's much more clear indeed 👍 

So we're entering the territory of best practices. Something that is not only useful for your applications, but for all GRDB users.

Let's make it right for everybody, then.

Shouldn't we deprecate the long-lived property, and if possible have it call the new API? Shouldn't we also find a better name than `passphraseBlock`, which uses vocabulary from the Objective-C era? And update the documentation?

groue:
>  It's about doing our part to keep the passphrase out of application memory when we can.

I would also appreciate that we did something that effectively *clears* that memory after it has been used, if possible. I'm not sure that Swift string deallocation does this. I'm not the security expert in this pull request, but your perspective on this would be appreciated as well.

groue:
> So now whenever we access that passphrase, we're respecting the restrictions we've placed on the passphrase: "only accessible after first unlock", "only accessible when unlocked", etc., rather than only enforcing that restriction once at launch, which defeats the purpose of the various key protection classes.

On this topic, shouldn't we also close existing connections whenever the passphrase becomes unavailable?

groue:
Hi again @michaelkirk-signal,

After a little more thoughts, please let me rephrase your goals, explain why this pull request will not be merged, and propose an alternative plan.

We want to support two use cases:

1. Some users want to apply the best security practices about the passphrase bytes.
2. Some users want the database to be accessible if and only if the passphrase is.

---

**This pull request does not allow to apply the best security practices about the passphrase bytes.**

You expressed concerns about the lifetime of the passphrase in memory. This is a valid concern.

But your suggested change does not solve it, because the Swift String object provides no guarantee about the lifetime of the string bytes in memory. For example, you won't find anywhere that Swift String securely wipes its backing storage when it is deallocated. Some users will not be satisfied by your suggested change.

In order to let demanding users apply their best security practices, it is important that GRDB does not interfere at all. We must provide raw access to the `sqlite3_key` function. And I think this is already possible with current APIs:

```swift
// How to open an encrypted database when you're a security expert:
configuration.prepareDatabase = { db in
    ... // Here, carefully load passphrase bytes
    let code = sqlite3_key(db.sqliteConnection, /* passphrase bytes */)
    ... // Here, carefully dispose passphrase bytes
    guard code == SQLITE_OK else {
        throw DatabaseError(resultCode: code, message: db.lastErrorMessage)
    }
}
let dbQueue = try DatabaseQueue(path:..., configuration: configuration)
```

I did not test this code, but I'm pretty sure it would work, when I look at the [connection setup code](https://github.com/groue/GRDB.swift/blob/v4.3.0/GRDB/Core/Database.swift#L203-L227).

I'll gladly accept a pull request which adds tests for this advanced technique, so that all GRDB users are guaranteed it will work in the long run.

---

**This pull request does not allow the database to be accessible if and only if the passphrase is.**

You claimed that you want to honor the various `kSecAttrAccessible` [Accessibility Values](https://developer.apple.com/documentation/security/keychain_services/keychain_items/item_attribute_keys_and_values#1679100). And your suggested change do indeed, at least for the passphrase: it makes it impossible to open a new database connection when the passphrase is not accessible.

However existing database connections are still accessible. Practically speaking:

1. After a database queue or pool has been opened, write accesses will always be permitted, even after the passphrase has turned not available.
2. After a database queue has been opened, read accesses will always be permitted as well.
3. After a database pool has been opened, some reads will succeed, and some reads will fail (those that will fails are parallel reads that want to open a new connection).

Don't points 1 and 2 defeat your purpose? Even if you claim they don't, I'll easily find users who claim they do.

Point 3 looks like a nasty bug to me, concurrency-based, and thus difficult to reproduce.

So here is my suggestion:

- Check if it is possible for a GRDB user to be notified that a passphrase did turn unavailable.
- Provide GRDB users a way to close all connections.

Given this, users will just have to listen to the notification and close all connections: the database will no longer be more accessible than the passphrase. Well, not *exactly*, but the precise details of the introduced guarantee are left as an exercice for the author of a pull request.

I can already provide the signature and documentation of the method that close all connections:

```swift
extension DatabaseQueue {
    /// Blocks the current thread until the SQLite database connection has been closed.
    /// Future database accesses will reopen the connection.
    ///
    /// For example:
    ///
    ///     let dbQueue: DatabaseQueue = ...
    ///     // Close the database connection
    ///     try dbQueue.closeConnection()
    ///     // Reopens the connection
    ///     try dbQueue.write { db ... }
    func closeConnection() throws
}

extension DatabasePool {
    /// Blocks the current thread until the SQLite database connections have been closed.
    /// Future database accesses will reopen connection(s).
    ///
    /// For example:
    ///
    ///     let dbPool: DatabasePool = ...
    ///     // Close the database connections
    ///     try dbPool.closeConnections()
    ///     // Reopens a writer connection
    ///     try dbPool.write { db in ... }
    ///     // Reopens a read-only connection
    ///     try dbPool.read { db in ... }
    func closeConnections() throws
}
```

---

Final word, let me explain how I came to the conclusions and suggestions above.

It's because I had much difficulties figuring out how your pull requests could be *documented*. I could not find any way to describe the behavior and expected guarantees that users of `passphraseBlock` could expect.

Please make the effort of writing documentation, especially when the change you suggest is not trivial, or enters the territory of a difficult domain. Security is very difficult. Writing documentation helps spotting oddities, inconsistencies, traps, etc. If you can't explain an API to another user - not a passive user but an active one that questions your choices - then maybe it's not a good API. But you don't know until you have tried to explain it, first to your inner self, maybe to your coworkers, and surely to me.

Thank you. I'd like to know what you think of my suggestions.

groue:
> In order to let demanding users apply their best security practices, it is important that GRDB does not interfere at all. We must provide raw access to the `sqlite3_key` function. And I think this is already possible with current APIs:
> 
> ```swift
> // How to open an encrypted database when you're a security expert:
> var configuration = Configuration()
> configuration.prepareDatabase = { db in
>     ... // Here, carefully load passphrase bytes
>     let code = sqlite3_key(db.sqliteConnection, /* passphrase bytes */)
>     ... // Here, carefully dispose passphrase bytes
>     guard code == SQLITE_OK else {
>         throw DatabaseError(resultCode: code, message: db.lastErrorMessage)
>     }
> }
> let dbQueue = try DatabaseQueue(path:..., configuration: configuration)
> ```
> 
> I did not test this code, but I'm pretty sure it would work, when I look at the [connection setup code](https://github.com/groue/GRDB.swift/blob/v4.3.0/GRDB/Core/Database.swift#L203-L227).
> 
> I'll gladly accept a pull request which adds tests for this advanced technique, so that all GRDB users are guaranteed it will work in the long run.

As a convenience, I'll accept that not-so-demanding user can provide a passphrase String:

```swift
configuration.prepareDatabase = { db in
    let passphrase = fetchPassphraseFromEnclave()
    try db.decrypt(with: passphrase)
}
let dbQueue = try DatabaseQueue(path:..., configuration: configuration)
```

The `Database.decrypt(with:)` method is just the existing [`setCipherPassphrase(_:)`](https://github.com/groue/GRDB.swift/blob/v4.3.0/GRDB/Core/Database.swift#L389) method, renamed, and made public.

This will behave just the same as your `passphraseBlock`, but better packaged.

However, this new API **must** come with the `closeConnection()` methods described above. Without them, we'd introduce an opportunity for the nasty concurrency bug I described. Of course, I may be mistaken in my analysis, so I remain curious about your motivated opinion.

groue:
> [...] users will just have to listen to the notification [that a passphrase did turn unavailable] and close all connections: the database will no longer be more accessible than the passphrase.

Closing connections in database queues and pools can be done with new methods, as suggested.

However, it is impossible to close the connection of a [DatabaseSnapshot](https://github.com/groue/GRDB.swift/blob/master/README.md#database-snapshots), because closing its connection would kill the state of [snapshot isolation](https://sqlite.org/isolation.html) that provides the guarantee of an unchanging database content.

Therefore, if a user wants passphrase accessibility to be reflected on the database, database snapshots must be deallocated when the passphrase turns unavailable. This will have to be documented, because only users can deallocate their snapshots.

michaelkirk-signal:
> I would also appreciate that we did something that effectively clears that memory after it has been used, if possible. I'm not sure that Swift string deallocation does this. I'm not the security expert in this pull request: your perspective on this would be appreciated as well.

It's safe to assume that Swift string deallocation does *not* do anything to zero at memory.

Some enhancements in CryptoKit with iOS13 offers behavior like this for keys retrieved from the enclave, but I need to do more research to see how this applies for generic "passphrase" data stored in the enclave. It's not immediately obvious to me how it does.

> On this topic, shouldn't we also close existing connections whenever the passphrase becomes unavailable?

This is a substantially harder nut to crack. I've read and agree with your analysis. I don't immediately have an answer for how to handle this in a reasonable default way, but will continue to think about it.

Since it seems like if you were to proceed with this, you'd likely not go the route of the `passphraseBlock`,  so I'll close this PR. Thanks for your thoughts.

groue:
Thank, Michael

> This is a substantially harder nut to crack. I've read and agree with your analysis. I don't immediately have an answer for how to handle this in way that would be a reasonable default for all users, but will continue to think about it.

Thank you again :-)

I had my own progression on the topic since my last comments. "Closing a database", currently, is equivalent to the deallocation a whole database queue/pool/snaphot. Maybe we don't need to invent a new way to close a database.

We could:

1. Deprecate `Configuration.passphrase`
2. Test and ship `Database.decrypt(with:)`
3. Document that it MUST be used from `Configuration.prepareDatabase`, and nowhere else.
4. Explain that:

    Errors thrown from `prepareDatabase` are rethrown when you create a database queue, pool, or snapshot. They are also rethrown, whenever GRDB opens a new SQLite connection, by database access methods (read, write) and high-level APIs such as database observation. For example, new connections can be created when a database pool performs several parallel reads. Generally speaking, SQLite connections are managed by GRDB, which may close and open them at any time (*).
    
    Practically speaking, there are consequences if you fetch the passphrase from the system keychain. After the passphrase turns inaccessible, you will face database errors whenever GRDB wants to open a new connection. On the other side, connections that are already open will remain usable. Some database accesses will fail, and some database accesses will succeed (**).

    You may want to destroy your database queue, pool and snapshots instances whenever the passphrase turns inaccessible, and reopen them after the passphrase is accessible again (***).

    (*) This sentence is important, because I want future GRDB versions to be able to freely change their policy whenever relevant.

    (**) This is what I call a bug in previous comments. It's less a bug if we describe it up front, and provide a workaround.

    (***) Is there any system notification that we could suggest using?

michaelkirk-signal:
> (***) Is there any system notification that we could suggest using?

UIApplicationProtectedDataWillBecomeUnavailable seems promising: 

https://developer.apple.com/documentation/uikit/uiapplication/1623058-protecteddatawillbecomeunavailab

> Upon receiving this notification, clients should release any references to protected files. This notification does not contain a userInfo dictionary.

There is subtly here:
> clients should release any references to __protected__ files.

How exactly do we interpret "protected"?   We have to talk about two related things,  *passphrase* (enclave) protection and *file* protection.

For our purposes, enclave restrictions can fit into two buckets

> **kSecAttrAccessibleWhenUnlocked**
> The data in the keychain item can be accessed only while the device is unlocked by the user.
>
> **kSecAttrAccessibleAfterFirstUnlock**
> The data in the keychain item cannot be accessed after a restart until the device has been > unlocked once by the user.

There are more cases to handle for fileProtectionLevels.

## switch fileProtectionLevel:

### case NSFileProtectionComplete
In the trivial case that the user has specified NSFileProtectionComplete for the db and key material, then it seems reasonable to close the database immediately.

### case NSFileProtectionCompleteUntilFirstUserAuthentication
If they *were* opened, they must have previously authenticated. Leave them open, even if the device is locked. 

### case NSFileProtectionNone
Leave them open, even if the device is locked.

### case NSFileProtectionCompleteUnlessOpen:
This one is tricky. If taken literally, it seems like we shouldn't do anything. Open ones will remain open. It's a little strange, because opening/closing connections is a private detail of GRDB, it's hard to be confident what the developer really wants to have happen here. Maybe it should be officially discouraged in the documentation as not making sense for database files.

Additionally, while developing, note that UIApplicationProtectedDataWillBecomeUnavailable won't be called when data protection isn't available - namely on simulators or devices without a passcode.

So there are two things the user needs to configure in lockstep:

1. The enclave restrictions on the db passphrase
2. The fileProtectionLevel of the files on disk

If a user wants their db available after first auth, we'd want them to do something like:
passphrase: kSecAttrAccessibleAfterFirstUnlock
dbDir: NSFileProtectionCompleteUntilFirstUserAuthentication

If a user never wants their db available when locked, we'd want them to do something like:
passphrase: kSecAttrAccessibleWhenUnlocked
dbDir: NSFileProtectionComplete

Note that it's possible for the user to think they're doing well, then shoot themselves in the foot:
passphrase: kSecAttrAccessibleWhenUnlocked
dbDir: NSFileProtectionNone

We could try to do something clever which offers a high level "security configuration" API to make it harder to shoot yourself in the foot, but there are a lot of legitimate combinations, and ultimately the user could modify things with the underlying OS API's anyway.  So I think the risk of introducing a new security concept outweighs the benefit.

The user who understands fileProtectionLevel and enclave passphrase storage can already do the right thing. Documented recipes like those outlined above are the best hope for the rest, while never obscuring the underlying iOS systems.

...

Going back to the beginning of my post:
> There is subtly here:
>> clients should release any references to __protected__ files.

We'd use `UIApplicationProtectedDataWillBecomeUnavailable` as a trigger to *check* on the database, and close it if appropriate based on its file protection level.

Best practice would be to dynamically fetch the passphrase from the enclave in `configureDatabase` block. Using a securely zeroing type should be recommended in the documentation. The specifics of that need more design.

groue:
> We'd use UIApplicationProtectedDataWillBecomeUnavailable as a trigger to check on the database, and close it if appropriate based on its file protection level.

Yes. As you say, users will have to think about what they have to do.

Our role is not to provide a ready-made high-level API that performs the necessary work. There are just too many distinct real-world situations. It's more important to let users do what they need to do - and document where their attention should focus.

> Best practice would be to dynamically fetch the passphrase from the enclave in configureDatabase block. Using a securely zeroing type should be recommended in the documentation. The specifics of that need more design.

Here as well, we don't have to provide a ready-made high-level API. GRDB is a database library, it would be a mistake to introduce security-sensitve methods inside it. We don't have the necessary skills in the contributors of the lib. We could not safely maintain the security-sensitive code over the years. The strategy that makes it possible to talk directly to SQLCipher, to the `sqlite3_key()` function, is the sweet spot that moves the security responsibility on to the application, leaving GRDB out of this dangerous game. And we provide a String-based API for convenience, because not all users have the *skills* to write secure code.

groue:
The [Encryption](https://github.com/groue/GRDB.swift/blob/c2e68d95febf515f78ef69713200c1afab9b3a4a/README.md#encryption) documentation chapter has been rewritten.

Folks, I think I have addressed questions and concerns, and that the pull request is ready to be merged.

Please chime in quickly if you see something wrong, or this will ship as is in GRDB 4.4.

PR #602, opened 2019-09-01
Don't keep the sqlCipher passphrase in memory longer than necessary

groue:
This PR is a replacement for #599. cc @michaelkirk-signal

The way to set the passphrase of an encrypted SQLCipher database has changed. The old way is still supported, but deprecated:

```swift
// Deprecated
var config = Configuration()
config.passphrase = "secret"
let dbQueue = try DatabaseQueue(path: ..., configuration: config)
```

One now sets the passphrase in the database preparation function:

```swift
// New
var config = Configuration()
config.prepareDatabase = { db in
    try db.usePassphrase("secret")
}
let dbQueue = try DatabaseQueue(path: ..., configuration: config)
```

This new technique allows the most demanding users to manage the security of the passphrase at the lowest possible level, the `sqlite3_key` C function:

```swift
// Some security experts will appreciate this
var config = Configuration()
config.prepareDatabase = { db in
    ... // Carefully load passphrase bytes
    let code = sqlite3_key(db.sqliteConnection, /* passphrase bytes */)
    ... // Carefully dispose passphrase bytes
    guard code == SQLITE_OK else {
        throw DatabaseError(resultCode: code, message: db.lastErrorMessage)
    }
}
let dbQueue = try DatabaseQueue(path: ..., configuration: config)
```

This pull request is not complete. The `change(passphrase:)` method has to be deprecated, too:

```swift
try dbQueue.change(passphrase: "newSecret")
```

It has to be deprecated because it stores the new password in memory, for the remaining lifetime of the database queue or pool. And it does not allow demanding users to change the passphrase at the lowest possible level.

I'm facing a few difficulties there. Details will come.


groue:
How to run SQLCipher tests:

- Clone GRDB
- `cd Tests/CocoaPods/SQLCipher4`
- `pod update`
- Open Tests/CocoaPods/SQLCipher4/GRDBTests.xcworkspace
- Run the GRDBTests test scheme

groue:
> The change(passphrase:) method has to be deprecated, too. [...] I'm facing a few difficulties there.

#603 aims at solving those difficulties.

Changing the passphrase of a database queue should soon look like:

```swift
// Deprecated
try dbQueue.change(passphrase: "newSecret")

// New
    try db.changePassphrase("newSecret")
}
```

And for a database pool:

```swift
// Deprecated
try dbPool.change(passphrase: "newSecret")

// New: use a barrier so that no read can happen
// during the modification of the passphrase:
try dbPool.barrierWriteWithoutTransaction { db in
    try db.changePassphrase("newSecret")
    // All readers should now use the new passphrase
    dbPool.invalidateReadOnlyConnections()
}
```

Those new techniques allow the most demanding users to manage the security of the passphrase at the lowest possible level, with the raw sqlite3_rekey() function. The Database.changePassphrase(_:) method is a convenience for not-so-demanding users who are OK with using Swift strings.

However, applications that need to change the passphrase of encrypted databases will have to provide a shared storage for the passphrase.

```swift
// Initialize storage. Here it is the memory, but it could be the keychain.
var passphrase = "oldSecret"

var config = Configuration()
config.prepareDatabase = { db in
    // Read passphrase from the storage
    db.usePassphrase(passphrase)
}
let dbPool = try DatabasePool(path: ..., configuration: config)

try dbPool.barrierWriteWithoutTransaction { db in
    // Store new passphrase
    passphrase = "newSecret"
    try db.changePassphrase(passphrase)
    dbPool.invalidateReadOnlyConnections()
}
```

@michaelkirk-signal, what do you think?

Also, `barrierWriteWithoutTransaction`, introduced by #603, does not take database snapshots in account: it allows snapshots to run concurrently with the database re-keying, with consequences I have not analyzed. Maybe the re-keying fails, I don't know. Maybe concurrent snapshot reads fail, I don't know. Surely *some* future snapshot reads will fail. After all, snapshots created before the re-keying use an obsolete passphrase: application has to stop using them. As you are a contributor to https://github.com/secure-sign/securesign-ios, a repository that uses database snapshots, you are the only user I know with some experience of them. I wonder if you have an opinion on this?

groue:
@darrenclark, @hartbit, @Marus, you are users of SQLCipher (or have been). Would you mind telling if this pull request looks like a correct way forward to you?

hartbit:
If users write the following code:

```swift
var passphrase = "oldSecret"

var config = Configuration()
config.prepareDatabase = { db in
    try db.usePassphrase(passphrase)
}
```

won’t the `prepareDatabase` closure capture the `passphrase` variable, thus causing the secret to be kept in memory?

If that’s correct, then this change just transposes the problem to the client side. No?

groue:
Hi David,

> If that’s correct, then this change just transposes the problem to the client side. No?

Totally, that's precisely the goal :smiling_imp:

There are now at least three levels of security:

**Level 1**: The worst level of security is the code snippet you comment. It indeed captures the `passphrase` variable, and keeps the secret in memory. This is not much different from the current implementation of GRDB, which keeps the secret in a Configuration value, for at least as long as the database queue/pool that uses it (this is what @michaelkirk-signal wanted to solve).

So... it doesn't introduce any *new* bad behavior. But maybe we should not show this in the documentation.

**Level 2**: The second one is somehow better (but I don't really know how much). At least GRDB doesn't retain String instances:

```swift
var config = Configuration()
config.prepareDatabase = { db in
    let passphrase = try fetchPassphraseFromSystemKeychain()
    try db.usePassphrase(passphrase)
    // <- passphrase is released here
}
```

The passphrase remains in memory as long as... Swift does. I'm sure String is not designed to be secure. But it's convenient: some users will be OK with that.

And I guess this is what should be shown in documentation.

**Level 3**: And finally, if you are a very demanding user, you can manage the passphrase bytes as you want, as close as possible from the SQLCipher metal (you can't be closer):

```swift
var config = Configuration()
config.prepareDatabase = { db in
    ... // Load passphrase bytes as securely as you can
    let code = sqlite3_key(db.sqliteConnection, /* passphrase bytes */)
    ... // Dispose bytes as securely as you can
    guard let code = SQLITE_OK else { throw DatabaseError(...) }
}
```

This pull request does not provide any ready-made API at this level. Heck, if passphrase comes from some secure UITextField, then the passphrase has been floating around wrapped in an NSString, and I'm not even sure it is useful to perform any byte-juggling after that.

I guess this low-level technique will be suggested in the doc, without much detail.

In a nutshell, I aim at this sweet spot: let security experts do their best, without pretending we are security experts.

groue:
> [...] won’t the prepareDatabase closure capture the passphrase variable, thus causing the secret to be kept in memory?

Do you think, @hartbit, that we should explicitly warn against such capture? Or think about a way to prevent it in the first place?

groue:
Some news on the `change(passphrase:)` front.

It will be deprecated, but now I'm not sure it will get a replacement.

There are several reasons for that.

Some blind spot created by the deprecated `Configuration.passphrase` property prevented me from realizing that when the passphrase comes from a *failable storage* like the system keychain, things are very difficult for the application. This is because there are two passphrases to deal with: the one in the storage, managed by the app, and the one that encrypts the database, managed by SQLCipher. When one passphrase change fails, and the other succeeds, you end up with a database that is impossible to decrypt - unless some backup of the previous passphrase has been kept. Applications have to implement some kind of *journaling* of the passphrase. Applications have to be able to perform *several passphrase attempts*.

This means that providing sample code for changing database passphrase becomes close to impossible. Too simple, the sample code would have no value. And GRDB is not supposed to tell how to perform journaling of keychain items. This is the job of the application, or another library.

Next, the `sqlite3_rekey()` function that performs in-place passphrase change is [discouraged](https://github.com/ccgus/fmdb/issues/547#issuecomment-259219320) in favor of [this technique](https://discuss.zetetic.net/t/how-to-encrypt-a-plaintext-sqlite-database-to-use-sqlcipher-and-avoid-file-is-encrypted-or-is-not-a-database-errors/868).

Besides, I'm pretty sure there exist indirect strategies that allow the database passphrase to never change and still support user password updates.

All those reasons tell me that a high-level API for in-place passphrase change has little value. Too much to lose here, too little to win for the users.

hartbit:
> Do you think, @hartbit, that we should explicitly warn against such capture? Or think about a way to prevent it in the first place?

I think the safest API is to request the passphrase as a DatabaseQueue/Pool initializer argument: that makes it much safer than providing it inside a closure that captures very easily.

> It will be deprecated, but now I'm not sure it will get a replacement.

That might be problematic. In all applications I currently manage that use GRDBCipher, I require the database to be re-encrypted when the user password changes. So I’m going to be stuck using the deprecated API got a while if there isn’t a replacement.

> Besides, I'm pretty sure there exist indirect strategies that allow the database passphrase to never change and still support user password updates.

That might be a good alternative. Do you have more information on those strategies?

groue:
Thanks for your answers, @hartbit.

> I think the safest API is to request the passphrase as a DatabaseQueue/Pool initializer argument: that makes it much safer than providing it inside a closure that captures very easily.

But this would defeat the two purposes that this pull request (and I) wants to bring:

- stop keeping the passphrase in memory.
- provide access to the raw C passphrase API for the most demanding users

In some way, your suggestion could be convenient for database queues, which only use the passphrase when they open their SQLite connection. They could release the passphrase once the SQLite connection has been established. But it would not work at all for database pools, which open connections on demand. We'd need to keep the passphrase in memory.

So if you don't want the closure to capture the passphrase... Just don't capture it, maybe?

```swift
// Caution: passphrase is captured
let passphrase = try loadPassphrase()
var config = Configuration()
config.prepareDatabase = { db in
    try db.usePassphrase(passphrase)
}

// Better: passphrase is not captured
var config = Configuration()
config.prepareDatabase = { db in
    let passphrase = try loadPassphrase()
    try db.usePassphrase(passphrase)
}
```

>> `change(passphrase:)` will be deprecated, [and] now I'm not sure it will get a replacement
>
> That might be problematic. In all applications I currently manage that use GRDBCipher, I require the database to be re-encrypted when the user password changes. So I’m going to be stuck using the deprecated API got a while if there isn’t a replacement.

Would you be OK with the alternative techniques described below?

For database queues, it would give:

```swift
try dbQueue.write { db in
    try db.changePassphrase("newSecret")
}
```

For database pools, it is almost the same:

```swift
try dbPool.barrierWriteWithoutTransaction { db in
    try db.changePassphrase("newSecret")
    dbPool.invalidateReadOnlyConnections()
}
```

In whole honnesty, the case of database pool is not that easy.

With the deprecated `change(passphrase:)` method, the passphrase is kept in memory, so that future SQLite connections open for reading can use the new passphrase. And because `change(passphrase:)` keeps the passphrase in memory, it has to be deprecated.

Now that the app is responsible for the passphrase storage, this gives:

```swift
// Assume those two functions exist:
func loadPassphrase() throws -> String
func savePassphrase(_ passphrase: String) throws

// Setup the pool
var config = Configuration()
config.prepareDatabase = { db in
    let passphrase = try loadPassphrase()
    try db.usePassphrase(passphrase)
}
let dbPool = try DatabasePool(path: ..., configuration: config)

// Change the passphrase
let newPassphrase = "newSecret"
try dbPool.barrierWriteWithoutTransaction { db in
    try savePassphrase(newPassphrase)       // 1.
    try db.changePassphrase(newPassphrase)  // 2.
    dbPool.invalidateReadOnlyConnections()
}

// Keep on using the pool
try dbPool.read { db ... }
```

I see questions is the relation between 1 and 2 (which order to choose? What if one fails but not the other?).

But this "problem" has always existed, even for database queues, even before this PR. And it has always been on the shoulders of the application. It just becomes *much more visible* after this pull request. I'm not sure it is a defect.

Would you be OK with this?

>> Besides, I'm pretty sure there exist indirect strategies that allow the database passphrase to never change and still support user password updates.
>
> That might be a good alternative. Do you have more information on those strategies?

Not really. And I'm not a security expert. But the trick would be to generate a passphrase for the database, once, independent from the user password, and encrypt it. The user password is needed to decrypt the passphrase.

groue:
Oh, you may wonder why not keeping `change(passphrase:)`, if they are equivalent to:

```swift
try dbQueue.write { db in
    try db.changePassphrase("newSecret")
}

try dbPool.barrierWriteWithoutTransaction { db in
    try db.changePassphrase("newSecret")
    dbPool.invalidateReadOnlyConnections()
}
```

But they are not. At least, not for pools. For pools, you must change the passphrase for future reads *inside* the barrier, which blocks all concurrent reads:

```swift
try dbPool.barrierWriteWithoutTransaction { db in
    // Must happen inside the barrier, so that we are sure
    // no call to loadPassphrase() happens until the database
    // has been re-keyed.
    try savePassphrase(newPassphrase)
    
    try db.changePassphrase(newPassphrase)
    dbPool.invalidateReadOnlyConnections()
}
```

groue:
The [Encryption](https://github.com/groue/GRDB.swift/blob/c2e68d95febf515f78ef69713200c1afab9b3a4a/README.md#encryption) documentation chapter has been rewritten. It does not hide intrinsic difficulties, and is precise up to the point where the brain of the reader has to kick in. I think it's not bad.

Folks, as far as I know, I have addressed questions and concerns: the pull request is ready to be merged.

Please chime in quickly if you see something wrong, or this will ship as is in GRDB 4.4!

README.md
```diff
@@ -7388,105 +7387,128 @@ See [SQLCipher 4.0.0 Release](https://www.zetetic.net/blog/2018/11/30/sqlcipher-
...
-// 2. Later on, change the passphrase
 try dbQueue.write { db in
     try db.changePassphrase("newSecret")
 }
+```
 
-// 3. Keep using the database queue with the new passphrase
-try dbQueue.write { ... }
-try dbQueue.read { ... }
+When you use a [database pool](#database-pools), make sure that no concurrent read can happen during the passphrase change with the `barrierWriteWithoutTransaction` method, and that all future reads open a new database connection with the `invalidateReadOnlyConnections` method:
```
michaelkirk-signal: 🤓 language nerd here, but the "and that all future reads" is pretty far from the "make sure that" to which it applies.

Maybe this is a little clearer:

When you use a [database pool](#database-pools), make sure that no concurrent read can happen by changing the passphrase within the `barrierWriteWithoutTransaction` block. You must also ensure all future reads open a new database connection by calling the `invalidateReadOnlyConnections` method:

README.md
```diff
@@ -7388,105 +7387,128 @@ See [SQLCipher 4.0.0 Release](https://www.zetetic.net/blog/2018/11/30/sqlcipher-
...
+
+```swift
+try dbPool.barrierWriteWithoutTransaction { db in
+    try db.changePassphrase("newSecret")
+    dbPool.invalidateReadOnlyConnections()
+}
 ```
 
-For database pools, it is a little bit more complicated. This is because pools create SQLite connections on demand. Your application must provide a storage for the passphrase, ready to be queried whenever a new SQLite connection should be created. This "storage" can be, for example, the system keychain.
+> :point_up: **Note**: When an application wants to keep on using a database queue or pool after the passphrase has changed, it is responsible for providing the correct passphrase to the `usePassphrase` method called in the database preparation function.
```
michaelkirk-signal: I'm not totally clear on what this note is trying to prevent. Is it cases like this:

```
// this won't work across a password change
let passphrase = getStoredPassphrase()
prepareDatabase { db 
    try db.usePassphrase(passphrase)
}
```

```
// This could work across a password change
prepareDatabase { db 
    let passphrase = getStoredPassphrase()
    try db.usePassphrase(passphrase)
}
```

README.md
```diff
@@ -7388,105 +7387,128 @@ See [SQLCipher 4.0.0 Release](https://www.zetetic.net/blog/2018/11/30/sqlcipher-
...
-
-// Now the copy is done, and the clear-text database can be deleted.
+let dbQueue = try DatabaseQueue(path: dbPath, configuration: config)
 ```
 
-### Attempting Several Passphrases
+#### Passphrase availability vs. Database availability
 
-TODO. Context: synchronizing the passphrase in the keychain, and the passphrase used by the database is tricky: one saving operation may succeed, and the other fail. It is possible that one has to try *two* passphrases:
+When the passphrase is securely stored in the system keychain, your application can protect is using the [`kSecAttrAccessible`](https://developer.apple.com/documentation/security/ksecattraccessible) attribute.
```
michaelkirk-signal: [...] can protect ~~is~~ **it** using [...]

README.md
```diff
@@ -7388,105 +7387,128 @@ See [SQLCipher 4.0.0 Release](https://www.zetetic.net/blog/2018/11/30/sqlcipher-
...
+// May fail if passphrase has turned unavailable
+try dbPool.read { ... }
 
-- TODO: Connections can be used after passphrase has turned unavailable
-- TODO: Advanced passphrase protection (use cases for the raw C SQLCioher functions)
+// May trigger value observation failure if passphrase has turned unavailable
+try dbPool.write { ... }
+```
+
+Those eventual database pool failures are not predictable, because pools maintain a pool of long-lived SQLite connection(s). Some database accesses will use those existing connections, and succeed. Some other database accesses will fail, as soon as the pool wants to open a new connection.
```
michaelkirk-signal: > pools maintain a pool of long-lived

maybe less ambiguous:

> DatabasePool maintains a pool of long-lived SQLite connection(s)

README.md
```diff
@@ -7388,105 +7387,128 @@ See [SQLCipher 4.0.0 Release](https://www.zetetic.net/blog/2018/11/30/sqlcipher-
...
+
+For the same reason, a database queue, which also maintains a long-lived SQLite connection, will remain available even after the passphrase has turned unavailable.
+
+Application that want to prevent any database access when the passphrase is unavailable are responsible for destroying their instances of database queue or pool when the passphrase becomes unavailable.
```
michaelkirk-signal: Application**s**

README.md
```diff
@@ -7350,35 +7349,35 @@ pod 'SQLCipher', '~> 3.4'
...
+var config = Configuration()
+config.prepareDatabase = { db in
     try db.usePassphrase("secret")
 }
-let dbQueue = try DatabaseQueue(path: "...", configuration: configuration)
+let dbQueue = try DatabaseQueue(path: dbPath, configuration: config)
 ```
 
-After you have set the passphrase, you can perform other [SQLCipher configuration steps](https://www.zetetic.net/sqlcipher/sqlcipher-api/) that must happen very early in the database lifetime. For example:
+It is also in the database preparation function that you perform other [SQLCipher configuration steps](https://www.zetetic.net/sqlcipher/sqlcipher-api/) that must happen early in the lifetime of a SQLCipher connection. For example:
```
michaelkirk-signal: Maybe less ambiguous:

> It is also in ~~the database preparation~~ `prepareDatabase` that you perform [...]

michaelkirk-signal:
Overall the API, inherent limitations, and documentation make sense to me. 

I don't have any real world use with the changePassphrase API, so I can't speak to it much. At a high level the API and limitations make sense to me.

I added some nit-picky feedback to the documentation in https://github.com/groue/GRDB.swift/pull/602/commits/c2e68d95febf515f78ef69713200c1afab9b3a4a. Feel free to take-it-or-leave-it.

Thanks for seeing this through!

README.md
groue: Thanks Michael, your way to say it is so much better 👍  cc7d19d92567c6c71ba80ba474db7df5fc72afe6

README.md
groue: Yes, this is it. Do you think we should provide guidance here, as below?

> :point_up: **Note**: When an application wants to keep on using a database queue or pool after the passphrase has changed, it is responsible for providing the correct passphrase to the `usePassphrase` method called in the database preparation function. Consider:
>
> ```swift
> // WRONG: this won't work across a passphrase change
> let passphrase = try getPassphrase()
> var config = Configuration()
> config.prepareDatabase = { db in
>     try db.usePassphrase(passphrase)
> }
>
> // CORRECT: get the latest passphrase when it is needed
> var config = Configuration()
> config.prepareDatabase = { db in
>     let passphrase = try getPassphrase()
>     try db.usePassphrase(passphrase)
> }
> ```

README.md
groue: 👍

README.md
groue: You're right, this is very bad English. I did attempt a full rewrite of the paragraph: cc7d19d92567c6c71ba80ba474db7df5fc72afe6

README.md
groue: Thanks again

README.md
groue: 👍

groue:
> Overall the API, inherent limitations, and documentation make sense to me.

Thanks @michaelkirk-signal! After all, you are the one who kicked all this brainstorm, so satisfying your initial request is the least I can do :-)

> I don't have any real world use with the changePassphrase API, so I can't speak to it much. At a high level the API and limitations make sense to me.

Yes, same for me. I'm just wondering if people can deal properly with keychain error AND/OR database error that may happen when they change the password. I can wait a couple more days before I merge the PR and hope for the best.

> I added some nit-picky feedback to the documentation in c2e68d9. Feel free to take-it-or-leave-it.

Your help is very welcome, Michael :-)

> Thanks for seeing this through!

Thank you! We have a much more secure GRDB now!

README.md
groue: OK let's do it: a1c96f909333e81909e1cfd06b335c42cc1147ab

michaelkirk-signal: Yes I think that's better. 👍

hartbit:
I understand better why ˋprepareDatabase` is necessary. And with that better understanding, I agree that your current solution seems like the best.

> But they are not. At least, not for pools. For pools, you must change the passphrase for future reads inside the barrier, which blocks all concurrent reads:

Those code snippets are complex enough (especially for pools) that I think it might make sense to wrap the passphrase change in a convenience function, that is obviously named differently than `change(passphrase:)`:

```swift
func triggerPassphraseChange(to passphrase: String) {
    try write { db in
        try db.changePassphrase(passphrase)
    }
}

func triggerPassphraseChange(to passphrase: String) {
    try barrierWriteWithoutTransaction { db in
        try db.changePassphrase(passphrase)
        self.invalidateReadOnlyConnections()
    }
}
ˋˋˋ

groue:
> Those code snippets are complex enough (especially for pools) that I think it might make sense to wrap the passphrase change in a convenience function, that is obviously named differently than change(passphrase:):

I wish it were possible. But https://github.com/groue/GRDB.swift/pull/602#issuecomment-527581820

Do you yourself use an encrypted pool, and change its passphrase? If yes... Then how do you manage the change of passphrase, in *both* the database and the keychain (assuming you use the keychain)? How do you handle partial errors (passphrase change in keychain is OK, but not in the db, or the opposite)?

This is the rather complex use case that we don't want to hide inside a convenience method. It would not be convenient: it would be naive. It's OK if `barrierWriteWithoutTransaction` looks like a magic incantation to some users. It won't stay magic for long after they understand that they must not allow keychain and database to get out of sync.

... And on top of that, those String-based convenience methods prevent access to the raw C SQLCipher APIs, which is undesirable in the context of this PR. The only String-based convenience methods we want to allow are on the `Database` class, because it exposes its raw SQLite connection handle and allows demanding users to avoid unsecure Swift strings.

groue:
🚀 Shipped in version 4.4.0! Thank you all :-)

PR #650, opened 2019-11-14
Update code example in README.md

SebastianOsinski:
Fixes code sample for inserting struct record with implicit rowId
### Pull Request Checklist

- [x] ~~Inline documentation has been updated.~~ (not applicable)
- [x] ~~Changes are tested.~~ (not applicable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3269,7 +3269,8 @@ When SQLite won't let you provide an explicit primary key (as in [full-text](Doc
     You will then be able to track your record ids, update them, or check for their existence:
     
     ```swift
-    let event = Event(message: "foo", date: Date())
+    // Event is a struct, so it needs to be a variable to allow mutation (via didInsert(with:for) method) on insertion:
+    var event = Event(message: "foo", date: Date())
     
     // Insertion sets the record id:
     try event.insert(db)


groue:
Hello @SebastianOsinski,

Thanks for your interest and this pull request, but it won't be accepted.

Yes, what you write is 100% correct. You understand what is happening and why `var` is required.

Yes, I am able to understand why you spent some time in order to figure it out, and why you were confused.

Yes, I agree there are opportunities for documentation improvements.

But your pull request teaches about Swift struct mutability in a chapter that is telling about the hidden rowid column.

**Documentation should generally avoid teaching several things at the same time.** We want to help readers *focus*. Focus is better when *readers are ready for information*. Readers are ready for information when documentation is *telling a story* that readers are interested into. Documentation stories are better when they are *logical and linear*. You can't plug a sentence about Swift struct mutability in a story about the rowid column. You can't expect readers to care about this information at this time. You may even confuse some other readers.

On the topic of record mutation, the place where the doc could be improved is in the chapter about PersistableRecord protocol. Maybe you'll find a way to improve the story that is told here. Make sure you do not teach about Swift: when one makes the `let/var` mistake, one gets an error from the Swift compiler, and it is not our job to tell users how to fix this error. What we could do is helping users understand that structs with an auto-incremented id are mutated when they get inserted.

Issue #652, opened 2019-11-18
databaseMigrator.appliedMigrations throws error for new users


michaelkirk-signal:
### What did you do?

As our schema changes, we register new migrations with our DatabaseMigrator. Existing users will run the new migrations incrementally as they are run.

As an optimization for new users, we want to skip the incremental migrations, and only import the final resultant schema.

We need to:
1. Detect if the user is a new user.
2. Import the final resultant schema, which we track outside of GRDB.
3. Mark all the currently existing migrations as "run", so they don't get "re-run" next launch.

For [1], to detect if the user is new, we thought to inspect `databaseMigrator.appliedMigrations`. However an error is thrown when running `try migrator.appliedMigrations(in: pool)` for new users.

### What did you expect to happen?

I expected to run `try migrator.appliedMigrations(in: pool)` for a new user and get an empty result, to infer that no migrations had been run. 

This would mean we could run the optimized version of the migrator which imports the final schema, rather than running the incremental migrations.

### What happened instead?

An error is thrown because `appliedMigrations` consults the `grdb_migrations` table, which does not necessarily exist. Furthermore, the method which sets up the migration table, is private:

```
private func setupMigrations(_ db: Database) throws {
    try db.execute(sql: "CREATE TABLE IF NOT EXISTS grdb_migrations (identifier TEXT NOT NULL PRIMARY KEY)")
}
```

It seems reasonable that this method is private, but we're wondering if there's a public stable API to achieve what we want without hardcoding assumptions about where the migrations will live.

### Environment

**GRDB flavor(s):** SQLCipher
**GRDB version:** 4.3.0
**Installation method:** CocoaPods
**Xcode version:** 10.3
**Swift version:** 5.0
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 10.14

### Demo Project

I can provide one if it's not clear from the above.

groue:
Hello @michaelkirk-signal,

`appliedMigrations` should unquestionably return an empty array if the migrations table has not been created yet. Would you like to set up a pull request for this fix?

PR #653, opened 2019-11-18
Fix: `appliedMigrations` throws error on new install

michaelkirk-signal:
Before setup, the grdb_migrations table doesn't exist, so calling
`appliedMigrations` would throw an error.

To avoid this, if the table to track migrations doesn't exist yet, we can be
sure no migrations have been run, and we retun the empty set.

FIXES https://github.com/groue/GRDB.swift/issues/652

### Pull Request Checklist

- [n/a] Inline documentation has been updated.
- [n/a] README.md or another dedicated guide has been updated.
- [x] Changes are tested.

### DIFF
--- a/GRDB/Migration/DatabaseMigrator.swift
+++ b/GRDB/Migration/DatabaseMigrator.swift
@@ -214,14 +214,6 @@ public struct DatabaseMigrator {
     /// Returns the set of applied migration identifiers.
     public func appliedMigrations(in reader: DatabaseReader) throws -> Set<String> {
         return try reader.read { db in
-            guard let doesMigrationsTableExist = try Bool.fetchOne(db, sql: "SELECT EXISTS (SELECT 1 FROM sqlite_master WHERE type='table' AND name='grdb_migrations')") else {
-                fatalError("grdb_migrations check unexpectedly returned nil")
-            }
-
-            guard doesMigrationsTableExist else {
-                return Set()
-            }
-
             return try appliedIdentifiers(db)
         }
     }
@@ -241,6 +233,12 @@ public struct DatabaseMigrator {
     }
     
     private func appliedIdentifiers(_ db: Database) throws -> Set<String> {
+        let tableExists = try Bool.fetchOne(db, sql: """
+            SELECT EXISTS (SELECT 1 FROM sqlite_master WHERE type='table' AND name='grdb_migrations')
+            """)!
+        guard tableExists else {
+            return []
+        }
         return try Set(String.fetchAll(db, sql: "SELECT identifier FROM grdb_migrations"))
     }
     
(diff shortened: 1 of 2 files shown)


GRDB/Migration/DatabaseMigrator.swift
```diff
@@ -213,7 +213,17 @@ public struct DatabaseMigrator {
     
     /// Returns the set of applied migration identifiers.
     public func appliedMigrations(in reader: DatabaseReader) throws -> Set<String> {
-        return try reader.read { try appliedIdentifiers($0) }
+        return try reader.read { db in
+            guard let doesMigrationsTableExist = try Bool.fetchOne(db, sql: "SELECT EXISTS (SELECT 1 FROM sqlite_master WHERE type='table' AND name='grdb_migrations')") else {
```
groue: I can't help thinking that `appliedMigrations(in:)` is fixed, but the private `appliedIdentifiers` method is still sensitive to the lack of the grdb_migrations tables, for no good reason.

Maybe a better fix would be to fix the root `appliedIdentifiers` method itself.

```swift
public func appliedMigrations(in reader: DatabaseReader) throws -> Set<String> {
    return try reader.read(appliedIdentifiers)
}

private func appliedIdentifiers(_ db: Database) throws -> Set<String> {
    let tableExist = try Bool.fetchOne(db, sql: """
        SELECT EXISTS (SELECT 1 FROM sqlite_master WHERE type='table' AND name='grdb_migrations')
        """)!
    guard tableExist else {
        return []
    }
    return try Set(String.fetchAll(db, sql: "SELECT identifier FROM grdb_migrations"))
}
```

Also note how the force unwrap of the Bool is just enough: SQLite will never return zero row or NULL.

Tests/GRDBTests/DatabaseMigratorTests.swift
```diff
@@ -236,6 +236,31 @@ class DatabaseMigratorTests : GRDBTestCase {
             }
         }
     }
+
+    func testAppliedMigrations() throws {
```
groue: Amazing, there was no plain test of this method yet 🤣

Thanks 👍

GRDB/Migration/DatabaseMigrator.swift
michaelkirk-signal: I've pushed up a new commit which does this.

groue:
> error: unexpected service error: The Xcode build system has crashed. Please close and reopen your workspace.

Travis is sooo flaky. Let's wait a little more…

groue:
Yes! I'll ship 4.6.2 shortly. Thank you Michael!

groue:
Done! CocoaPods will be updated when `pod trunk push` has ended.

PR #660, opened 2019-12-01
Database Lock Prevention

groue:
**This pull request makes it possible to prevent acquisition of locks on a database.**

It aims at providing the necessary ground-level work for the prevention of `0xdead10cc` exception in iOS apps that store their database in the container on an [App Group](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_application-groups).

https://developer.apple.com/library/archive/technotes/tn2151/_index.html:

> The exception code `0xdead10cc` indicates that an application has been terminated by the OS because it held on to a file lock or sqlite database lock during suspension. If your application is performing operations on a locked file or sqlite database at suspension time, it must request additional background execution time to complete those operations and relinquish the lock before suspending.

We can thus prevent the `0xdead10cc` exception if we can release and prevent locks between the notification that the additional background execution time is about to expire, and the end of the suspended state.

In this pull request, we only address the exclusive lock prevention itself, not its scheduling.

```swift
dbQueue.startPreventingLock() // or dbPool
dbQueue.stopPreventingLock()  // or dbPool
```

Those methods are public because they are defined on the DatabaseReader protocol. But they are not supposed to be used by the GRDB user. They won't be documented in the main README. They will be used by a future pull request which will target `0xdead10cc` specifically.

Between `startPreventingLock()` and `stopPreventingLock()`, all database accesses may throw a DatabaseError of code `SQLITE_INTERRUPT`, or `SQLITE_ABORT`, except reads in WAL mode.


groue:
> I could witness that SHARED locks do not trigger `0xdead10cc` (by performing a long read).

:scream: I was wrong. My initial test did indeed perform a long running query, by calling a custom function which only returns after one minute. But it did not access the database content, and SQLite did not create the SHARED lock.

A better test reveals that accessing the database content in a long running query does trigger the `0xdead10cc`.

Do we don't really need exclusive lock prevention. What we need is to prevent any kind of database access. Back to the drawing board!

groue:
> A better test reveals that accessing the database content in a long running query does trigger 0xdead10cc.

...except in WAL mode.

PR #663, opened 2019-12-08
[On Hold] Sharing a Datatase in an App Group Container

groue:
This pull request aims at helping applications sharing their databases in App Group containers.

It is composed of a set of new APIs, and most importantly a new guide: [Sharing a Datatase in an App Group Container](https://github.com/groue/GRDB.swift/blob/dev/dead10cc/Documentation/AppGroupContainers.md).

We have several goals:

1. All users understand that this guide is the guide that will help them fixing their errors, if they happen to jump into App Group containers before reading it.
2. All users understand that sharing a database is an App Group container comes with new kinds of errors, and understand those errors.
3. All users understand that they can copy-paste the sample code (hence the sample code must be correct, and focused on the problem to be solved).
4. Expert users understand and agree with the recommended setup.

---

I would appreciate if any knowledgeable SQLite/GRDB user would check if the [guide](https://github.com/groue/GRDB.swift/blob/dev/dead10cc/Documentation/AppGroupContainers.md) sounds sensible (@michaelkirk-signal, perhaps?)

---

**Help wanted**: this pull request is ON HOLD until we learn a way to reproduce the `0xdead10cc` crash in an iOS application extension.


groue:
Closing now that #668 has been merged in

Issue #500, opened 2019-03-10
FetchedRecordsController collection view and multiple sections


simme:
Hey,

So, I've read and understood that the `FetchedRecordsController` is basically only tested with table views and that there's no current support for multiple sections. And also that the performance might be so-so on bigger data sets.

I'm writing an app that could benefit from all of the above sooner rather than later and was wondering if you put any thoughts into how it could work and how it would be implemented. I _might_ have the possibility of helping out and contributing to something like this — depends a little on how my spare time can be allocated. Not impossible that I could actually get my work to sponsor some time for this too.

I had one thought, that might be helpful or misguided :P What if: instead of fetching full records the initial query only fetches row ids for each record? Could this help with up-front performance? The data source prefetch APIs of UIKit might help with fetching items from the data store. Might make other things like diffing harder though..

CoreData uses a key path to determine the sectioning of items. So I guess one query that groups items on one of the record's properties and then uses that to render the sections could be one solution.

"The problem" with all ideas I can come up with is that it "burdens" the developer with constructing a query that does what the controller expects. But maybe that's not too much to ask? Maybe it could be formalized in some way using the query interface or an initializer that builds the query based on input parameters.

I'm just spewing out ideas. I haven't spent _too_ much time thinking about this or looking at the internals of the current `FetchedRecordsController` to really come up with anything useful at this point. Which is why I'm asking if you, @groue, have done any thinking already? :)

groue:
Welcome back, @simme!

> So, I've read and understood that the FetchedRecordsController is basically only tested with table views and that there's no current support for multiple sections. And also that the performance might be so-so on bigger data sets.

Yes, this describes pretty well the FetchedRecordsController situation. Collection views are supported, though, because they accept the same diffs as table views.

One of the earliest GRDB observation tool, FetchedRecordsController shows it direct inspiration from Core Data. It hasn't been touched in years. Its inner implementation is not very easy to analyse and change. It looks like it lacks some love.

But it is a great high level tool, because diffing is difficult. It is central in the [demo app](https://github.com/groue/GRDB.swift/blob/master/DemoApps/GRDBDemoiOS/README.md).

> I'm writing an app that could benefit from all of the above sooner rather than later and was wondering if you put any thoughts into how it could work and how it would be implemented. I might have the possibility of helping out and contributing to something like this — depends a little on how my spare time can be allocated. Not impossible that I could actually get my work to sponsor some time for this too.

[...]

PR #682, opened 2020-01-15
Adding Multiple Section Support to FetchedRecordsController

gjeck:
### Summary:
This is an exploration into adding multiple section support to `FetchedRecordsController`. I've closely followed the suggestions described in #500 (these were very helpful). Though #635 has mentioned that `FetchedRecordsController` is being sunset, I would hope that any discussion here would possibly show a path forward for possibly keeping the feature around. 

Essentially, the internals of `FetchedRecordsController` have been rewritten to utilize [ValueObservation](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservation) and the awesome [DifferenceKit](https://github.com/ra1028/DifferenceKit) library has been embedded to support diffing. I don't know if embedding the dependency is the correct path forward and am open to suggestions for alternatives.

I haven't gone too far into the implementation. Tests are pretty lacking. I didn't want to go too far before getting any feedback though, so that's why I've posted this now. If it turns out that this isn't something desirable for GRDB to have, perhaps it will at least serve as an implementation example for something else.

### Usage:
The main alteration to the API of `FetchedRecordsController` is an addition of a `sectionColumn` property that allows results to be grouped based on a row column. E.g:
```
let controller = try FetchedRecordsController(dbQueue, request: Person.all(), sectionColumn: Column("name"))
```

### For Review:
The change-set is quite large for this, but most of it is just embedding the DifferenceKit library. The files I think that are most important for review are `FetchedRecordsController.swift`, `PlayersViewController.swift` and `FetchedRecordsControllerTests.swift`.

### Media:
multiple-sections

### Pull Request Checklist

- [ ] Changes are tested.

### DIFF
--- a/Documentation/DemoApps/GRDBDemoiOS/GRDBDemoiOS/Player.swift
+++ b/Documentation/DemoApps/GRDBDemoiOS/GRDBDemoiOS/Player.swift
@@ -17,7 +17,7 @@ extension Player: Hashable { }
 // See https://github.com/groue/GRDB.swift/blob/master/README.md#records
 extension Player: Codable, FetchableRecord, MutablePersistableRecord {
     // Define database columns from CodingKeys
-    private enum Columns {
+    enum Columns {
         static let id = Column(CodingKeys.id)
         static let name = Column(CodingKeys.name)
         static let score = Column(CodingKeys.score)
--- a/Documentation/DemoApps/GRDBDemoiOS/GRDBDemoiOS/PlayersViewController.swift
+++ b/Documentation/DemoApps/GRDBDemoiOS/GRDBDemoiOS/PlayersViewController.swift
@@ -9,8 +9,8 @@ class PlayersViewController: UITableViewController {
     }
     
     @IBOutlet private weak var newPlayerButtonItem: UIBarButtonItem!
-    private var players: [Player] = []
-    private var playersObserver: TransactionObserver?
+    private var playerData: [ArraySection<FetchedRecordsSectionInfo<Player>, Item<Player>>] = []
+    private var playersFetchController: FetchedRecordsController<Player>?
     private var playerCountObserver: TransactionObserver?
     private var playerOrdering: PlayerOrdering = .byScore {
         didSet {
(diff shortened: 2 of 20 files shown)


groue:
Hello @gjeck,

> This is an exploration into adding multiple section support to FetchedRecordsController. I've closely followed the suggestions described in #500 (these were very helpful). Though #635 has mentioned that FetchedRecordsController is being sunset, I would hope that any discussion here would possibly show a path forward for possibly keeping the feature around.

Thank you for that.

Yes, I prefer leaving FetchedRecordsController alone, for two main reasons:

1. It looks like NSFetchedResultsController, and I don't want to force Core Data-inspired users to migrate their existing and running code to ValueObservation, or its RxGRDB and GRDBCombine wrappers. Keeping FetchedRecordsController alive *does not harm* the rest of the library. It is based on foundations that are quite stable now, so it does not require much maintenance. Why not leaving happy users alone? GRDB is mature now, and the stability of its API is an asset.

2. FetchedRecordsController is a *very bad fit* for the concurrency model of GRDB. Getting inspiration from Core Data was unwise. The problem appears when you want to fetch values that are related to the observed request. How do you make sure they are *consistent*?

    The `fetchAlongside` FetchedRecordsController parameter aims at "solving" this issue, but it is a plaster on a wooden leg.  This class is not a tool which allows the user to gracefully enlarge the observation scope.

    ValueObservation is a much better fit for the concurrency model of GRDB, it can deal with as complex fetching strategies as needed, and should be fostered in place of FetchedRecordsController.

All in all, I have to thank you. A few years ago, this pull request would have been welcomed with open arms. But those days are gone.

---

Here is my suggestion:

Unless I'm mistaken, FetchedRecordsController is fully based on public APIs, which means that you can copy it in your own app, and extend it with DifferenceKit and support of sections. It is simpler than maintaining a fork of GRDB.

You can even *ship it* as an external library, just like RxGRDB and GRDBCombine.

I'll be happy to soothe any pain point you would meet in this process.

gjeck:
I completely agree with your feedback. As I got further into the implementation it seemed like this would make way more sense as an external library. I'll close this issue for now and will reach out once I've got a library up if you think it would be worth forwarding to any future users 😄

groue:
I'm glad we agree :-) And sure, ping me when you have something new!

PR #683, opened 2020-01-16
String concatenation

groue:
This pull request lets you concatenate strings with the query interface:

```swift
// SELECT firstName || ' ' || lastName FROM player
Player.select([
    Column("firstName"), 
    " ".databaseValue, 
    Column("lastName")]
    .joined(operator: .concat))
```

The `joined(operator:)` method now accepts `.concat`, `.and`, `.or`, `.add`, and `.multiply`.


tarun-personatech:
@groue Wonderfull. Thank you so much.

groue:
Shipped in v4.9.0

PR #731, opened 2020-03-11
Hide ValueObservation implementation details

groue:
ValueObservation used to expose an underling protocol, `ValueReducer`, shipped as an experimental interface.

It appears that it has no real use in applications, and that value reducers can be removed from the public API.

Making them an implementation detail was also an opportunity to simplify the implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 - [#724](https://github.com/groue/GRDB.swift/pull/724): SE-0242: Synthesize default values for the memberwise initializer
 - [#728](https://github.com/groue/GRDB.swift/pull/728): Make ValueObservation error handling mandatory
 - [#729](https://github.com/groue/GRDB.swift/pull/729): ValueObservation always emits an initial value
+- [#731](https://github.com/groue/GRDB.swift/pull/731): Hide ValueObservation implementation details
 
 
 ## 4.11.0
--- a/GRDB/Core/DatabaseReader.swift
+++ b/GRDB/Core/DatabaseReader.swift
@@ -251,7 +251,7 @@ public protocol DatabaseReader: AnyObject {
     /// during observation
     /// - parameter onChange: a closure that is provided fresh values
     /// - returns: a TransactionObserver
-    func add<Reducer: ValueReducer>(
+    func add<Reducer: _ValueReducer>(
         observation: ValueObservation<Reducer>,
         onError: @escaping (Error) -> Void,
         onChange: @escaping (Reducer.Value) -> Void)
@@ -367,7 +367,7 @@ public final class AnyDatabaseReader: DatabaseReader {
     // MARK: - Value Observation
     
     /// :nodoc:
-    public func add<Reducer: ValueReducer>(
+    public func add<Reducer: _ValueReducer>(
         observation: ValueObservation<Reducer>,
         onError: @escaping (Error) -> Void,
         onChange: @escaping (Reducer.Value) -> Void)
--- a/GRDB/Core/DatabaseSnapshot.swift
+++ b/GRDB/Core/DatabaseSnapshot.swift
@@ -116,7 +116,7 @@ extension DatabaseSnapshot {
     
     // MARK: - Value Observation
     
-    public func add<Reducer: ValueReducer>(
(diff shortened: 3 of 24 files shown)


gjeck:
Is there an alternative API provided that can replace this? I'm currently using `AnyValueReducer` to [support multi-section value observation](https://github.com/groue/GRDB.swift/pull/682/files#diff-fa2b12f71f327bbd1a119b96f5cc20a2R107) for a sectioned based `UICollectionViewController`.

groue:
That's a valid question, @gjeck! 

Did you think of a way to achieve your goals without ValueReducer, and did you face a difficulty?

groue:
BTW, GRDB5 will have features removed, but it won't ship blindly. There will be at least one beta or release candidate that will allow to gather feedback.

gjeck:
So I have tried to resolve this by using `request.observationForAll` and using the `onChange` block, but I have run into issues that I will try and outline here.

Essentially, I have written an extension on `QueryInterfaceRequest where T: FetchableRecord` that supports multiple sections. Note that the constraint is only on `FetchableRecord` and not `Hashable` and `Equatable`, which is required to support `NSDiffableDataSourceSnapshot`. So, with `AnyValueReducer` I had access to the `Row` items, which are `Hashable` and `Equatable`, allowing a wrapper class that is initialized with the `Row` and lazily creates a `lazy var record = T(row: self.row)`. This wrapper can then power a `NSDiffableDataSourceSnapshot`. Without the custom reducer, I'd only get `[T]` in the `onChange` block.

Without this API it seems I would have to force my extension to provide tighter constraints, e.g. `extension QueryInterfaceRequest where T: FetchableRecord, T: Hashable, T: Equatable`. This isn't too big of a deal, but is a bit of a bummer unless I'm missing some other way to satisfy this problem.

Let me know if this doesn't make sense and I can try and post some more details.

gjeck:
Upon further inspection, it looks like I can just use:
```
ValueObservation.tracking(request, fetch: { ... })
```
Sorry for misunderstanding!

groue:
@gjeck, thank you for providing details: now we can talk.

First, comparing rows instead of records is smart. Indeed, this allows to remove requirements from the record type, because the database already contains the necessary information :+1:

Next, I'm happy `ValueObservation.tracking(_:fetch:)` had you find a solution without ValueReducer.

Last, I'll talk optimization. To this end, I'll compare three strategies which end up with the same results, but don't have the same runtime behavior:

```swift
let request = /* some request */

// 1. ValueObservation.tracking(value:)
let observation = ValueObservation.tracking(value: { db -> MyValue in
    let dbValues = try request.fetchAll(db)
    return makeMyValue(from: dbValues)
})


// 2. ValueObservation.tracking(_:fetch:)
let observation = ValueObservation.tracking(request, fetch: { db -> MyValue in
    let dbValues = try request.fetchAll(db)
    return makeMyValue(from: dbValues)
})

// 3. ValueObservation.tracking(_:fetch:).map(_:)
let observation = ValueObservation
    .tracking(request, fetch: request.fetchAll)
    .map { makeMyValue(from: $0) }
```

All those observations emit exactly the same sequence of MyValue.

Strategy 2 adds an optimization for DatabasePool. After a transaction has impacted the tracked request, the fetch is performed in a concurrent reader. The advantage is that the unique, serialized, writer dispatch queue is not blocked by the fetch. Other threads can write in the database while the fetch is executing. This reduces *writer contention*, and this what helps a "Save" button in your UI complete immediately, instead of waiting for the writer dispatch queue to be available.

Strategy 3 adds another optimization: the `makeMyValue` function is no longer called in a database dispatch queue. The advantage is that the database reader connection turns available for other reads from other threads earlier. This reduces *reader contention*.

Strategy 3 is supported by the now private ValueReducer protocol:

```swift
protocol ValueReducer {
    associatedtype Fetched
    associatedtype Value
    
    /// Runs in a database dispatch queue
    func fetch(_ db: Database) throws -> Fetched
    
    /// Does not run in a database dispatch queue
    mutating func value(_ fetched: Fetched) -> Value?
}
```

But you can achieve it with the public `map` operator, because `map` does not execute its closure argument in a database dispatch queue.

---

All in all, thank you very much for your vigilance. This conversation will certainly feed the future Migration Guide :+1:

gjeck:
Rad! Thank you for the follow-up and extensive details @groue 😄

Issue #601, opened 2019-08-23
ValueObservation won't start until concurrent write transaction has ended


zdnk:
Hello, I have encountered this issue today and not sure if there is any setting to fix this.

Basically I have observation:
```swift
return ProductCategory.children(of: id).rx.observeAll(in: writer, startImmediately: true, observeOn: ConcurrentDispatchQueueScheduler(qos: .userInitiated))
```

And then there is a write transaction that takes several seconds to finish (syncing database from server dump).

It seems that the observation does not emit any events as long the transaction is running which seems odd since I am using `DatabasePool`:
```swift
            var config = Configuration()
            config.foreignKeysEnabled = true
            config.label = "products-db"
            config.maximumReaderCount = 100
            let pool = try! DatabasePool(path: path, configuration: config)
            pool.setupMemoryManagement(in: UIApplication.shared)
```

I am also using the regular `read` and `write` methods on `DatabasePool`

Would you be able to provide any guidance on this issue please?

zdnk:
I just double-checked with brekapoints and it really is not running my read statements while there is write transaction :(

zdnk:
One note: the observable creation is actually wrapped in `pool.read {}`, but since reads are not serialized according to the documentation, that should not be the issue, right?

groue:
Hello @zdnk,

> It seems that the observation does not emit any events as long the transaction is running.

So you say that you'd like the observation to be notified of partial changes that happen during the transaction, even before the transaction has been completed.

Indeed this is not how ValueObservation works. Its [documentation](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservation) states:

> Changes are only notified after they have been committed in the database.

And this won't change. ValueObservation applies one of the most important design decisions in GRDB, which is that most applications want to use the database as a **single source of truth**. And what is the database truth? Data which is saved on disk. This is why only changes which are saved to disk *after* the transaction has been successfully committed are notified by ValueObservation.

This principle greatly simplifies the architecture of most applications. For example, clients of a ValueObservation only care about the content of the database. They don't have to worry about transactions that *fail*. Or uncommitted transactions that temporarily put the database in an *inconsistent state*. Your server synchronizer only has to care about its transaction, and nothing else. All components are independent and only have to care about their own business. This is quite a relief!

> [this] seems odd since I am using DatabasePool

[...]

PR #733, opened 2020-03-12
Fetch the current value of a ValueObservation

groue:
This pull request introduces the `ValueObservation.fetchValue(_:)` method.

The intent is to soothe a pain point of ValueObservation: a long running background transaction delays the notification of the first value of a ValueObservation, even when the application is using a DatabasePool. There are very good reasons for this unfortunate behavior, which is unlikely to change as long as ValueObservation comes with the hard guarantee that *absolutely all* changes are notified. See #601 by @zdnk for more information.

`ValueObservation.fetchValue(_:)`, as a regular fetching method, is not concerned by long running background transactions. It should help users prepend the current value to the regular  elements of a ValueObservation, and avoid unwanted delays.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 - [#729](https://github.com/groue/GRDB.swift/pull/729): ValueObservation always emits an initial value
 - [#731](https://github.com/groue/GRDB.swift/pull/731): Hide ValueObservation implementation details
 - [#732](https://github.com/groue/GRDB.swift/pull/732): Remove ValueObservation.compactMap
+- [#733](https://github.com/groue/GRDB.swift/pull/733): Fetch the current value of a ValueObservation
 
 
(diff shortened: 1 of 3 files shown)


groue:
> There are very good reasons for this unfortunate behavior, which is unlikely to change as long as ValueObservation comes with the hard guarantee that absolutely *all* changes are notified.

This guarantee is the crux of the issue, and this pull request is not a good answer, because it keeps on pushing all the hard work on to the users.

We need to drop this guarantee.

First, most users don't need *all* changes. They want their UI to be eventually updated with the latest value. They don't care if some values are lost, they care of not losing the last one. If I count from 1 to 10 very very fast, what is important is that the view eventually displays 10.

Next, users that need *all* changes can still use DatabaseRegionObservation.

Last, as soon as we drop the guarantee that ValueObservation notifies *all* changes, we open the path to new enhancements:

- Fix #601 without any special setup.
- Have ValueObservation refreshed when other processes modify the database.

groue:
Follow-up: #736

PR #748, opened 2020-04-04
Rename QueryInterfaceRequest to Request

groue:
I just could not stand any longer this awful long name 😅


groue:
This pull request was reverted. `Request` is not specific enough. Apps don't just perform database requests. They also perform HTTP requests, etc.

Issue #764, opened 2020-04-17
DateComponents best practises?


mtancock:
### What did you do?

Good question...

I'm storing information which includes dates, however the dates don't need to be precise - I'm only interested in YMD. I read in the readme that GRDB supports reading/writing to the db as dateComponents, but having a hard time visualising how I should set this up.

I'm currently creating my records as structs e.g

```swift
struct Activity: Codable, FetchableRecord, PersistableRecord {
    var id: String?
    var name: String
    var date: <What goes here?>
}
```

My migration looks like this:

```swift
try database.create(table: "activity") { table in
    table.column("id", .text).notNull().primaryKey()
    table.column("name", .text).notNull()
    table.column("date", .date).notNull()
}
```

I've tried using date directly, and at one point I had three separate fields for YMD as Ints, and then recreated them after fetching from the database but that seems annoying. Then I read the Readme (should have started there...) and see that there's support for `DateComponents` via `DatabaseDateComponents` but in terms of defining my record, saving back to the db etc. , I'm not sure how I should be using it. Any tips or suggestions where I should be looking to figure this out?

Thanks for all your work on this library. It's great.

### Environment

**GRDB flavor(s):** Regular GRDB
**GRDB version:** 4.13
**Installation method:** Cocoapods
**Xcode version:** 11.3.1
**Swift version:** 5ish
**Platform(s) running GRDB:** iOS/macOS
**macOS version running Xcode:** 10.15.4

groue:
Hello and thank you, @mtancock 

Let's start from your app:

> I'm storing information which includes dates, however the dates don't need to be precise - I'm only interested in YMD

It's indeed frequent to only care about YMD. Not all date-related contexts such as calendar (gregorian, japanese, locale-dependent, etc.) or time zones (local, utc, etc.) are always relevant. For example, YMD is a perfect storage format for birth dates (although it may be tactful to make the year optional).

**Your first question should be about the type of those dates**, the type that you want your application to deal with. It's too early to think about the database. Just think about the domain you are modelling, and the constraint you want your type to express.

- `Foundation.Date` is an *absolute point in time*, and does not look like a great fit.
- `Foundation.DateComponents` is better, but it forces your app to deal with its optional `year`, `month`, and `day` properties. Is it a big deal? I don't know, it all depends on your application constraints.
[...]

PR #766, opened 2020-04-17
Add Codable support to DatabaseDateComponents

mtancock:
Add `Codable` conformance to `DatabaseDateComponents` as per discussion in #764 

Embarrassingly, after \*cough\* years of development, this is my first open source PR so apologies for anything I've done wrong. 

What if any tests would be needed for this? I had a look but couldn't figure out what to test as there doesn't appear to be anything much equivalent in the library that I could steal from.

It also didn't seem like there was anything in the Readme to update? But happy to add something if you point me to what needs updating.

Of course if this is embarrassingly bad, let's just delete it and pretend it never happened?

### Pull Request Checklist

- [x] Changes are tested.

### DIFF
--- a/GRDB/Core/Support/Foundation/DatabaseDateComponents.swift
+++ b/GRDB/Core/Support/Foundation/DatabaseDateComponents.swift
@@ -11,7 +11,7 @@ import SQLite3
 public struct DatabaseDateComponents: DatabaseValueConvertible, StatementColumnConvertible, Codable {
     
     /// The available formats for reading and storing date components.
-    public enum Format: String, Codable {
+    public enum Format: String {
         
         /// The format "yyyy-MM-dd".
         case YMD = "yyyy-MM-dd"
@@ -162,19 +162,19 @@ public struct DatabaseDateComponents: DatabaseValueConvertible, StatementColumnC
 
     // MARK: - Codable adoption
 
-    /// CodingKeys for codable conformance.
-    enum CodingKeys: String, CodingKey {
-        case date
-    }
 
     /// Creates a new instance by decoding from the given decoder.
     ///
     /// - parameters:
     ///     - decoder: The decoder to read data from.
     public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        let stringValue = try container.decode(String.self, forKey: .date)
-        self = DatabaseDateComponents.fromDatabaseValue(stringValue.databaseValue)!
+        let container = try decoder.singleValueContainer()
+        let stringValue = try container.decode(String.self)
+        guard let decodedValue = DatabaseDateComponents.fromDatabaseValue(stringValue.databaseValue) else {
+            throw DecodingError.dataCorruptedError(in: container,
+                                                   debugDescription: "Unable to initialise databaseDateComponent")
+        }
+        self = decodedValue
     }
 
     /// Encodes this value into the given encoder.
@@ -182,7 +182,7 @@ public struct DatabaseDateComponents: DatabaseValueConvertible, StatementColumnC
     /// - parameters:
     ///     - encoder: The encoder to write data to.
     public func encode(to encoder: Encoder) throws {
-        var container = encoder.container(keyedBy: CodingKeys.self)
-        try container.encode(String.fromDatabaseValue(databaseValue)!, forKey: .date)
+        var container = encoder.singleValueContainer()
+        try container.encode(String.fromDatabaseValue(databaseValue)!)
     }
 }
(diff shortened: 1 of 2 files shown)


groue:
Hello, @mtancock, I'm happy you could not resist ;-)

So now, `DatabaseDateComponents` is Codable, and the code below now compiles:

```swift
struct Activity: Codable, FetchableRecord, PersistableRecord {
    var id: String
    var name: String
    var date: DatabaseDateComponents
}
```

Funnily, This Codable conformance *is not used* by GRDB, at runtime, when this type roundtrips a database. Indeed, we don't see the two coding keys, `dateComponents` and `format`, anywhere in the database content! We just see regular `YYYY-MM-DD` formatted strings, depending on the format property of the DatabaseDateComponents value.

The reason is the DatabaseValueConvertible protocol. It is this protocol which allowed DatabaseDateComponents to go into the database, even before this pull request was opened. And this protocol has priority over Codable, when a type conforms to both. This allows `Date` to encode itself in a `YYYY-MM-DD HH:MM.SS.SSS` string, instead of the numerical timestamp used by the Codable implementation of Date.

So can we do whatever we want in the Codable implementation of `DatabaseDateComponents`, since, in the context of Codable records, it is never used at runtime, and only used at compile time?

Not really, because now we have a public Codable conformance, so people may use it for other purposes. The `Activity` type above may end up in JSON, for example. In the proposed implementation, we get:

```json
// Proposed
{
  "id": "12345",
  "name": "Lockdown",
  "date": {
    "format": "yyyy-MM-dd",
    "dateComponents": {
      "hour" : 10,
      "minute" : 0,
      "year" : 2020,
      "month" : 4,
      "day" : 19
    }
  }
}
```

(Yeah, nothing prevents such extra `hour` and `minute` to sneak in).

This is not wrong, because it does the job. But I won't ship it as is, because I would not want to use it in my projects.

I suggest an alternative:

```json
// Alternative
{
  "id": "12345",
  "name": "Lockdown",
  "date": "2020-04-19"
}
```

The conversion of DatabaseDateComponents to and from a String is provided by `DatabaseValueConvertible`, as below:

```swift
let encoded = DatabaseDateComponents(
    DateComponents(year: 2020, month: 4, day: 19, hour: 10, minute: 00),
    format: .YMD)

let string = String.fromDatabaseValue(encoded.databaseValue)! // can't fail
print("string: \(string)")

if let decoded = DatabaseDateComponents.fromDatabaseValue(string.databaseValue) {
    print("decoded: \(decoded)")
} else {
    print("decoding error!")
}
```

Would you be OK with amending your pull request in this direction?

groue:
> What if any tests would be needed for this? I had a look but couldn't figure out what to test as there doesn't appear to be anything much equivalent in the library that I could steal from.

Yes we need at least one test. FoundationDateComponentsTests.swift is a good recipient. There is no "test encoder" in the stdlib or Foundation, so just perform a JSON encoding and check the resulting output: this will be quite enough.

> It also didn't seem like there was anything in the Readme to update? But happy to add something if you point me to what needs updating.

It's OK, there's nothing to add in the README. As we've seen together in #764, this conformance is perfectly expectable, and does not need any explicit mention.

> Of course if this is embarrassingly bad, let's just delete it and pretend it never happened?

I'd much prefer that we complete it :-) You are really helping, and not only on the coding side of things!

mtancock:
Sure - I'll look at those changes later on today. Thanks for the gentle review!

mtancock:
I've updated. I'm not sure what I should be doing (rather than force unwrapping) in the initialiser as there's no error to re-throw.

groue:
> I've updated. I'm not sure what I should be doing (rather than force unwrapping) in the initialiser as there's no error to re-throw.

You have to throw a DecodingError: https://developer.apple.com/documentation/swift/decodingerror

And you need a single-value-container: https://developer.apple.com/documentation/swift/encoder/2894221-singlevaluecontainer

mtancock:
I did suggest you might want to delete it. ;-)  Sorry I misread your comments. Update incoming

groue:
Hey, thank you very much @mtancock! Now we have a useful Codable implementation :-) I'll ship it shortly

groue:
@mtancock, welcome to the GRDB contributors! Would you like to be granted a push access to the repository? I'll invite you right away. You don't have to be shy, because you know how things work in this repo, now. You won't be bothered unless you want to. And it's good for the library that I'm not the only one able to push here.

mtancock:
Thanks. I feel completely unqualified but always happy to help out any way that I’m able to.

groue:
Invitation sent :-) You know, I was just as unqualified when this started 😅 All it takes is a desire to enhance our common daily tasks, and a little taste in order to remember that other people may use our code in *unexpected ways*, and that this is the beauty of our job :-)

groue:
Shipped in [v4.14.0](https://github.com/groue/GRDB.swift/tree/v4.14.0) 🤝

PR #767, opened 2020-04-18
Association to first or last element in a to-many Association

groue:
This pull request introduces the `first` and `last` properties on to-many associations.

They let you feed, for example, a list of all authors along with their last book:

```swift
// Setup
extension Author {
    static let books = hasMany(Book.self)
}

// Usage
struct AuthorInfo: FetchableRecord, Decodable {
    var author: Author
    var lastBook: Book
}
let infos = try dbQueue.read { db in
    try Author
        .including(required: Author.books.order(Column("year")).last)
        .asRequest(of: AuthorInfo.self)
        .fetchAll(db)
}
for info in infos {
    print("Last book of \(info.author.name) is \(info.lastBook.title)")
}
```


groue:
Question on the SQLite forums: [Subqueries and "no such column" error on references to other tables](https://sqlite.org/forum/forumpost/bae3d3ba2e)

groue:
This pr is not ready yet. The use case is definitely relevant, easy to define and understand, and there exists no straight-forward way to address it with the current apis.

But there are still too many implementation challenges. Basically, we need to define the missing foreign key in some kind of temporary tables. The above link to SQLite forums contain some drafts of SQL queries that provide an acceptable runtime complexity, and offer room for optimization when unique indexes exist.

The problem I don't feel like solving today is that query interface requests let the user use TableAlias in order to build bridges between the various associations used in a request. Those bridges make it difficult to properly isolate the temporary tables needed for this feature. The chances of generating invalid SQL is high. And it is hard to prevent them with a "not implemented" fatal error (the favorite GRDB way to avoid SQL syntax errors when a feature is not implemented).

So... We'll see later.

PR #803, opened 2020-06-29
Remove FetchedRecordsController

groue:
`FetchedRecordsController` was introduced more than four years ago (!) in version 0.54.0, thanks to @pakko972. Later it was sunsetted, in version 4.5.0, after ValueObservation had become the preferred observation tool.

With all due credits to [NSFetchedResultsController](https://developer.apple.com/documentation/coredata/nsfetchedresultscontroller), it is now time to say farewell.

So long, `FetchedRecordsController`!


pakko972:
Farewell old friend 👋

Issue #816, opened 2020-07-25
Discussion: naïve query interface filters on "boolean" columns


mallman:
Suppose we create a database `test` with the following schema:

```sql
CREATE TABLE version (
  id integer PRIMARY KEY NOT NULL,
  isInTrash boolean NOT NULL
)
```

Let's assume we have a GRDB `Version` record that simply follows the table definition. Let's consider three query interface requests:

```swift
let requestA = Version.select(Column("id")).filter(Column("isInTrash"))
let requestB = Version.select(Column("id")).filter(Column("isInTrash") == true)
let requestC = Version.select(Column("id")).filter(Column("isInTrash") == 1)
```

These requests generate the following SQL queries, respectively:

```sql
Query A: SELECT "id" FROM "version" WHERE "isInTrash"
Query B: SELECT "id" FROM "version" WHERE "isInTrash"
Query C: SELECT "id" FROM "version" WHERE "isInTrash" = 1
```

The SQL for `requestA` makes perfect sense. What surprises me is that the SQL for `requestB` is not the same as the SQL for `requestC`. In examining the source code for GRDB, I see why this is the case. I don't know if you would consider this a bug. However, I'd like to suggest this is not the best experience for the GRDB user. Let me elaborate.

I believe there is a common circumstance where a user who writes a query interface request like `requestB` will benefit greatly from using the SQL generated by `requestC`.

Suppose that every value in the `isInTrash` column is either 0 or 1. Then queries B and C return identical result sets.

Now suppose we create an index on the `isInTrash` column:

```sql
CREATE INDEX version_on_isInTrash ON version (isInTrash)
```

Let's examine the query plans for queries B and C.

```
sqlite> explain query plan select id from version where isInTrash;
QUERY PLAN
`--SCAN TABLE version USING COVERING INDEX version_on_isInTrash
sqlite> explain query plan select id from version where isInTrash = 1;
QUERY PLAN
`--SEARCH TABLE version USING COVERING INDEX version_on_isInTrash (isInTrash=?)
```

[...]

PR #817, opened 2020-07-26
Allow query planner to optimize some boolean tests

groue:
This pull request changes the SQL generated by boolean tests:

```swift
// GRDB 4: SELECT * FROM player WHERE isActive
// GRDB 5: SELECT * FROM player WHERE isActive = 1
Player.filter(Column("isActive") == true)

// GRDB 4: SELECT * FROM player WHERE NOT isActive
// GRDB 5: SELECT * FROM player WHERE isActive = 0
Player.filter(Column("isActive") == false)

// GRDB 4 & 5: SELECT * FROM player WHERE isActive
Player.filter(Column("isActive"))

// GRDB 4 & 5: SELECT * FROM player WHERE NOT isActive
Player.filter(!Column("isActive"))
```

The goal is allow users to trigger SQLite query planner optimizations when boolean columns and expressions are indexed (see #816 for the feature request).

However, this change is breaking for database boolean values that are not `0`, `1`, or `NULL`.


groue:
@mallman, do we agree this is what you had in mind?

groue:
> @mallman, do we agree this is what you had in mind?

Precisely speaking, *all* explicit Boolean checks are now fully expressed in SQL, with the sole exception of the `isEmpty` aggregate:

```swift
// SELECT ... HAVING COUNT(player.id) > 0
Team.having(Team.players.isEmpty == false)

// SELECT ... HAVING COUNT(player.id) = 0
Team.having(Team.players.isEmpty == true)

// SELECT ... HAVING COUNT(player.id) = 0
Team.having(Team.players.isEmpty)
```

mallman:
Looks good to me. I'm especially impressed by the support for indexed expressions. I didn't know SQLite supported that feature.

Thank you @groue!

groue:
You're welcome! Yeah, I sometimes check the SQLite release notes, so I learn about one trick or two ;-) Next beta will ship shortly. Thanks for having improved the library for you and other demanding users. And thanks for submitting your request while it was possible to introduce a breaking change 👍

groue:
Shipped in beta 10!

Issue #846, opened 2020-10-05
Non-public API usage on Mac Catalyst


harryworld:
This may likely an issue from App Store, but I'm curious if anyone experienced this and has suggestion on what to do next.

### What did you do?

Submit Mac Catalyst app with the use of GRDB.swift

Besides, I can also confirm that both iOS app and AppKit Mac app are working

### What did you expect to happen?

The binary can be submitted and processed successfully

### What happened instead?

Binary is invalid with the following email message

ITMS-90338: Non-public API usage - The app references non-public symbols in Contents/Frameworks/GRDB.framework/Versions/A/GRDB: _sqlite3_snapshot_cmp, _sqlite3_snapshot_free, _sqlite3_snapshot_get. If method names in your source code match the private Apple APIs listed above, altering your method names will help prevent this app from being flagged in future submissions. In addition, note that one or more of the above APIs may be located in a static library that was included with your app. If so, they must be removed. For further information, visit the Technical Support Information at http://developer.apple.com/support/technical/

### Environment

**GRDB flavor(s):** GRDB default
**GRDB version:** 5.0.1
**Installation method:** CocoaPods
**Xcode version:** 12.0 official
**Swift version:** 5.1
**Platform(s) running GRDB:** Mac Catalyst
**macOS version running Xcode:** 10.15.6

### Demo Project

https://github.com/harryworld/TestGRDB

This is really basic repo, by just adding GRDB via Cocoapods and doing nothing else

groue:
Hello @harryworld,

Thanks for reporting this issue. GRDB indeed uses SQLite [snapshots](https://www.sqlite.org/c3ref/snapshot.html), in a way that is questionable. Those apis are not exposed in the SQLite headers that ship with the various SDKs, but it happens that a library or an application can link against them.

I assumed this was a oversight in the headers.

Your issue reveals that it is actually *verboten*, and that this needs to be fixed at the GRDB level.

A fix will be released in the next few days.

See also #845 for a related issue.

harryworld:
Thanks for the quick response @groue 
So when was snapshot added? I would like to try a previous version prior to the addition of this.

groue:
You can go back to the latest GRDB 4 version. I can be quite fun to reverse the [migration guide to GRDB 5](https://github.com/groue/GRDB.swift/blob/master/Documentation/GRDB5MigrationGuide.md) ;-)

harryworld:
[...]

Issue #845, opened 2020-10-02
Unable to build with custom sqlite and GRDB5 due to snapshot functions


benasher44:
### What did you do?

Attempt to upgrade from GRDB 4 to 5 and build.

### What did you expect to happen?

After some hacking (after all, we do have a custom sqlite setup), build with GRDB 5.

### What happened instead?

I'm not able to build due to issues related to the snapshots functionality, which I will explain.

GRDB introduced features that use the sqlite snapshots functions, which can be enabled with the `SQLITE_ENABLE_SNAPSHOT` preprocessor macro. We don't enable this macro in our sqlite build. However, it appears the sqlite amalgamation comes with the main struct type defined for snapshots in its header, even if you don't have this macro enabled, which leads to the first issue we're seeing: compiler complains of duplicate definition of the `sqlite3_snapshot` struct. This is easy enough to work around (see notes in the sample project)— we can do some hacking to namespace the struct defined in GRDB.

Once we do that though, we get a linker error because the final GRDB binary (not compiled with `SQLITE_ENABLE_SNAPSHOT`) references `sqlite3_snapshot_get`, `grdb_snapshot_free`, and `grdb_snapshot_free`. By not using `SQLITE_ENABLE_SNAPSHOT` in the sqlite build, those functions are compiled out of the sqlite binary. But, GRDB still references them in the final binary, so the linker errors.

#### Possible Solutions

1. We could enable `SQLITE_ENABLE_SNAPSHOT` on our end, but we'd prefer not to do that. I haven't studied the surface area of those APIs, or how they would impact our sqlite usage + required testing.
2. Some kind of patch to GRDB to also compile out the referenced functions when `SQLITE_ENABLE_SNAPSHOT` is not enabled?
3. Some other hacking on our end, which I haven't considered? Open to suggestions

Thanks in advance for your help and consideration!

### Environment

**GRDB flavor(s):** (GRDB, SQLCipher, Custom SQLite build?): GRDB + custom sqlite build/hackery in our Podfile
**GRDB version:** 5.0.1
**Installation method:** (CocoaPods, SPM, manual?) CocoaPods
**Xcode version:** 11.7
**Swift version:** 5.2
**Platform(s) running GRDB:** (iOS, macOS, watchOS?) iOS
**macOS version running Xcode:** 10.15.6

### Demo Project

https://github.com/benasher44/GRDB5SnapshotTest

In the Podfile, there are lines to un-comment to attempt to namespace the already-defined struct.

[...]

PR #847, opened 2020-10-06
Remove dependency on SQLITE_ENABLE_SNAPSHOT

groue:
This pull request addresses #845 and #846


harryworld:
Thanks for the quick work.
So from now on, GRDB has `SQLITE_ENABLE_SNAPSHOT` turned off by default?

groue:
Yes @harryworld. `SQLITE_ENABLE_SNAPSHOT` is still used for ValueObservation optimization, but only when this flag has  explicitly been made available by a custom SQLite build.

groue:
I asked Apple to expose it publicly in FB9793771 (Expose SQLite APIs enabled by the SQLITE_ENABLE_SNAPSHOT compilation option)

PR #849, opened 2020-10-08
Fix grammar mistake AssociationsBasics.md

MarshalGeazipp:
I've just found this small grammar mistake and decided to fix it 😃 .


groue:
Thank you very much, @MarshalGeazipp! This kind of help is greatly appreciated 👍

PR #890, opened 2020-12-12
Row Values

groue:
This pull request brings [experimental](https://github.com/groue/GRDB.swift/blob/master/README.md#what-are-experimental-features) support for [row values](https://www.sqlite.org/rowvalue.html).

Row values let you write code like:

```swift
// WITH cte AS (...)
// SELECT * FROM player WHERE (a, b) IN cte
let cte = CommonTableExpression(...)
let request = Player
    .with(cte)
    .filter(cte.contains(RowValue2(Column("a"), Column("b"))))

// WITH cte AS (...)
// UPDATE player SET (a, b) = (SELECT * FROM cte)
let cte = CommonTableExpression(...)
try Player
    .with(cte)
    .updateAll(db, RowValue2(Column("a"), Column("b")).set(to: cte.all()))
```

The pull request comes with support for pairs and triples (`RowValue2` and `RowValue3`).

The API is experimental because it is not very satisfying. Ideally we'd represent SQL row values as Swift tuples:

```swift
// NOT IN THIS PULL REQUEST
let cte = CommonTableExpression(...)
let request = Player
    .with(cte)
    .filter(cte.contains((Column("a"), Column("b"))))

let cte = CommonTableExpression(...)
try Player
    .with(cte)
    .updateAll(db, (Column("a"), Column("b")).set(to: cte.all()))
```

Unfortunately, the Swift language does not let us define the `set(to:)` method on tuples (Swift tuples can't be extended).

Conclusion: a *nice* support for row values will have us redesign the assignment API. Until this happens, this pull request will do the job, even if it does not look good.

PS: also note the new `cte.contains(stuff)` api, which generates `stuff IN <cte table name>`.

cc @rpoelstra

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 - **Fixed**: [#881](https://github.com/groue/GRDB.swift/pull/881) by [@felixscheinost](https://github.com/felixscheinost) and [#883](https://github.com/groue/GRDB.swift/pull/883) by [@professordeng](https://github.com/professordeng): Documentation improvements
 - **Fixed**: [#885](https://github.com/groue/GRDB.swift/pull/885): Robustness of eager loading of to-many associations based on compound foreign keys
 - **Documentation update**: A new guide: [Common Table Expressions](https://github.com/groue/GRDB.swift/blob/master/Documentation/CommonTableExpressions.md)
-- **Documentation update**: The [Foreign Keys](https://github.com/groue/GRDB.swift/blob/master/Documentation/AssociationsBasics.md#foreign-keys) chapter of the Associations Guide clarifies the behavior of SQLite and GRDB regarding the presence of NULL if compound foreign keys.
+- **Documentation update**: The [Foreign Keys](https://github.com/groue/GRDB.swift/blob/master/Documentation/AssociationsBasics.md#foreign-keys) chapter of the Associations Guide clarifies the behavior of SQLite and GRDB regarding the presence of NULL in compound foreign keys.
 
 
 ## 5.2.0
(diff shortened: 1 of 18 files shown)


rpoelstra:
Is it also possible to perform a small calculation in the `.set(to:)` as I do in the example query?

groue:
> Is it also possible to perform a small calculation in the `.set(to:)` as I do in the example query?

`set(to:)` accepts an expression.

Expressions accept computations.

You can also pass a request to `set(to:)`. Requests can always be used where a regular expression is accepted: the request is then a subquery. Requests support customized selection, filtering, ordering, limit, etc.

So yes I think your example is supported, even if I did not check it explicitly.

rpoelstra:
You're absolutely right (of course). It works like a charm. Thanks for this all!

groue:
Thanks for suggesting valuable ideas!

groue:
@rpoelstra, I'm afraid I'll have to postpone the merging of this PR.

Row values don't work well with all versions of SQLite. Assignment, in particular, are not always OK:

```
On iOS 10.3.1: SQLite error 1: 2 columns assigned 1 values
WITH "cte" AS (SELECT 1, 2) UPDATE "t" SET ("a", "b") = (SELECT * FROM "cte")
```

Unfortunately, I can't express those constraints as Swift availability checks. I think it's a problem in the Swift language itself (did I miss something?):

```swift
// Protocol for assignable values (columns & row values made of columns)
public protocol SQLAssignable: SQLExpressible { }
extension SQLAssignable {
    public func set(to value: SQLExpressible?) -> Assignment { ... }
}

// Row values are not always available
@available(OSX 10.13, iOS 10.3.1, tvOS 10.3.1, watchOS 4, *)
public struct RowValue2<A: SQLExpressible, B: SQLExpressible>: SQLSpecificExpressible { ... }

// Assignment to row values are even less available
@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 5.0, *)
extension RowValue2: SQLAssignable where A: ColumnExpression, B: ColumnExpression { }

// Oddly, this code compiles on iOS 10.3.1, when it should only be available from iOS 12:
RowValue2(...).set(to:)
```

groue:
We'll revisit row values later.

PR #918, opened 2021-02-03
Added direct values initializer

Jasperav:
This enables me to write code that directly sets values, like so:

`selectStatement.setUncheckedArguments(StatementArguments(values: [firstName.databaseValue]))`

This improves speed. 

I am close to finishing an ORM based on this library which uses Codable records. Instead of GRDB checking properties at runtime, the properties are checked at compile time (by using code generation).

This was the only thing I needed to add for the ORM. I will mention you when I put it on Github to verify if the generated code is correct if you want.

### DIFF
--- a/GRDB/Core/Statement.swift
+++ b/GRDB/Core/Statement.swift
@@ -649,6 +649,11 @@ public struct StatementArguments: CustomStringConvertible, Equatable,
     /// Creates empty StatementArguments.
     public init() {
     }
+
+    /// Creates a new StatementArguments with the given values
+    public init(values: [DatabaseValue]) {
+        self.values = values
+    }
     
     // MARK: Positional Arguments
     


groue:
Hello @Jasperav, welcome to GRDB!

You can already initialize `StatementArguments` with an array of `DatabaseValue`:

```swift
// Already ok
StatementArguments(["foo".databaseValue, 1.databaseValue])
```

> This improves speed.

- Do you have any evidence for that?
- Would any improvement here be measurable against SQLite I/O?
- If so, can't we instead improve the existing public initializers, for the benefits of all GRDB users, and not only you?

Mind that `StatementArguments` has a private implementation, which can change at any time in the future. It is thus not desirable to expose an "optimized" initializer unless we *commit* for the long run on such optimization. Such commitment is not a light decision, and needs a precise analysis of how `StatementArguments` is used and where performance is lost.

> I am close to finishing an ORM based on this library which uses Codable records.

That's great! I'll be happy to discover your ideas :-)

Jasperav:
Thanks for the quick feedback!

I just released the v0.1.0 version of GRDB-ORM which can be found here: https://github.com/Jasperav/GRDB-ORM. Please checkout https://github.com/Jasperav/GRDB-ORM/blob/master/Parser/generated/DbUser.swift for the generated code.  

About these PR: The code performance is 30% faster with the initializer in this PR, than the initializer you are proposing. I know this sounds weird and such a difference is too much, so I would like you to test this out yourself. Please follow this steps to check out the difference yourself:

1. Run `testGenerated` a few times in https://github.com/Jasperav/GRDB-ORM/blob/master/GRDBPerformance/GRDBPerformanceTests/Insert.swift
2. Change the initializer here: https://github.com/Jasperav/GRDB-ORM/blob/master/GRDBPerformance/GRDBPerformance/Generated/DbUser.swift#L107 from `values: values` to just `values`
3. Repeat step 1 and see the big difference in performance. Maybe it's just my laptop, but I keep getting that large performance difference. With initializer in this PR, the time is around 0.330 (average run) and with the `values: values` initializer, it takes surely 0.400 for an average run.

groue:
> About these PR: The code performance is 30% faster with the initializer in this PR, than the initializer you are proposing.

Thanks for the detailed feedback, I'll look at it.

> I don't know about improving existing public initializers. I think the current initializers are clear enough. However, developing GRDB-ORM I came across functions that may decrease performance:
> 
> * When calling `Row.fetchAll`, this is invoked:
>   `return try Array(fetchCursor(statement, arguments: arguments, adapter: adapter).map { $0.copy() })`
>   Why the `.map { $0.copy() }`? For every row, a loop is done and above that, the row is copied. That looks expensive.
> * When calling `Row.fetchOne`, this is invoked:
>   `        // Keep cursor alive until we can copy the fetched row return try withExtendedLifetime(cursor) { try cursor.next().map { $0.copy() } }`
>   This looks like the same issue (the `$0.copy()`). I also don't understand the `withExtendedLifetime` (but maybe that's due to the lack of Swift from my side).

To understand this, you need to know how GRDB implement the [fetching methods](https://github.com/groue/GRDB.swift/blob/master/README.md#fetching-methods):

- `fetchCursor`
- `fetchAll`
- `fetchSet`
- `fetchOne`

All those methods are implemented differently, depending on the type of fetched values:

- Raw rows `Row`.
- Optimized values such as `Int`, `String`, `Date` that adopt the `StatementColumnConvertible` (barely public, not mentioned in the documentation).
- Non-optimized values such as `NSNumber` that I did not bother to optimize, and user types that adopt the `DatabaseValueConvertible` protocol.
- Records that adopt `FetchableRecord`.

**Let's start with cursors: `fetchCursor`**

The root fetching method is `fetchCursor`. Cursors MUST be iterated when the SQLite connection is available. This allows cursors to directly access the database values when an SQLite statement is iterated: https://www.sqlite.org/c3ref/column_blob.html

For example, take this cursor of `Int`:

```swift
// 🔥
let cursor = try Int.fetchCursor(db, sql: "SELECT score FROM player")
while let int = try cursor.next() {
    print(int)
}
```

It is a [FastDatabaseValueCursor](https://github.com/groue/GRDB.swift/blob/v5.3.0/GRDB/Core/StatementColumnConvertible.swift#L56) which is designed to trigger the optimized [Int decoding](https://github.com/groue/GRDB.swift/blob/v5.3.0/GRDB/Core/Support/StandardLibrary/StandardLibrary.swift#L99). In release configuration, it is compiled so that `sqlite3_step` and `sqlite3_column_int64` are called in a tight loop (spending a few extra cycles in order to check for NULL and overflow).

Now, take this cursor of `Row`:

```swift
// 🔥
let cursor = try Row.fetchCursor(db, sql: "SELECT id FROM player")
while let row = try cursor.next() {
    let id: Int = row[0]
} 
```

It is a [RowCursor](https://github.com/groue/GRDB.swift/blob/v5.3.0/GRDB/Core/Row.swift#L771) which *always return the same row*. This `Row` instance has a reference to an SQLite statement, and only happens to give distinct values on each iteration because `cursor.next()` calls `sqlite3_step`. The `let id: Int = row[0]` assignment triggers the same optimization as above, and we again have a tight loop of `sqlite3_step` and `sqlite3_column_int64`.

Finally, take this cursor of `Player`:

```swift
struct Player: FetchableRecord {
    var id: Int
    var name: String
    
    init(row: Row) {
        id = row["id"]
        name = row["name"]
    }
}

// 🐢 (slowed by column lookup)
let cursor = try Player.fetchCursor(db, sql: "SELECT * FROM player")
while let player = try cursor.next() {
    ...
} 
```

It is a [RecordCursor](https://github.com/groue/GRDB.swift/blob/v5.3.0/GRDB/Record/FetchableRecord.swift#L495) which, just as a row cursor, reuses a single instance of `Row` on each iteration step, and makes it possible for `Player.init(row:)` to, again, access the low-level `sqlite3_column_int64` and other fast extraction functions.

The `Player` record above spends time looking for the positions of "id" and "name" columns. Even if `Row` caches those positions, this takes a toll. This toll is the price to pay in order to accept columns in any order. With `Decodable`, this is even worse (you noticed it).

But it is possible to build [optimized records](https://github.com/groue/GRDB.swift/blob/master/README.md#examples-of-record-definitions) that lock the order or columns, and can decode with numeric positions without any String lookup. Those optimized GRDB records are twice as slow as raw calls to C SQLite functions, with a sweet Swift API, as well as error, null and overflow checking. Seriously, this is bloody fast, and this rocks:

```swift
struct OptimizedPlayer: FetchableRecord, TableRecord {
    var id: Int
    var name: String
    
    static let databaseTableName = "player"
    static let databaseSelection: [SQLSelectable] = [Column("id"), Column("name")]
    init(row: Row) {
        id = row[0]
        name = row[1]
    }
}

// 🔥
let cursor = try OptimizedPlayer.fetchCursor(db)
while let player = try cursor.next() {
    ...
} 
```

**Let's continue with arrays: `fetchAll`**

Arrays are usually built straight from cursors. They can thus profit from the raw access to the low-level `sqlite3_column_int64` and other fast extraction functions:

```swift
// 🔥
let ints = try Int.fetchAll(db, sql: "SELECT score FROM player")
let ints = try Array(Int.fetchCursor(db, sql: "SELECT score FROM player"))

// 🐢 (slowed by column lookup)
let players = try Player.fetchAll(db, sql: "SELECT * FROM player")
let players = try Array(Player.fetchCursor(db, sql: "SELECT * FROM player"))

// 🔥
let players = try OptimizedPlayer.fetchAll(db)
let players = try Array(OptimizedPlayer.fetchCursor(db))
```

There is one exception, though: arrays of raw `Row`. Do you remember that cursors above reuse a single instance of `Row`? Well we can't use the same technique when the user fetches an array of rows. The array will be iterated long after the SQLite statement has been fully iterated. And GRDB guarantees that arrays can be used on any thread.

That's why row arrays contain *row copies*, not based on the current state of an SQLite statement, but from `DatabaseValue`:

```swift
// 🐢 (slowed because rows are copied)
let rows = try Row.fetchAll(db, sql: "SELECT * FROM player")
let rows = try Array(Row.fetchCursor(db, sql: "SELECT * FROM player").map { $0.copy() })
```

The `fetchSet` and `fetchOne` methods have the same behavior (`Row` are copied).

From row copies, it is *impossible* to access the low-level `sqlite3_column_int64` and other fast extraction functions. The database context has been lost:

```swift
// Detached rows:
let row = try db.fetchOne(db, sql: "SELECT * FROM player")
let row = ["id": 1, "name": "Arthur"]
```

**An anti-pattern**

What happens when you decode values or records from detached rows? You prevent the raw access to the low-level `sqlite3_column_int64` and other fast extraction functions. And you spill memory by creating useless row instances:

```swift
// 🔥
let ints = try Int.fetchAll(db, sql: "SELECT score FROM player")

// 🐢
let rows = try Row.fetchAll(db, sql: "SELECT score FROM player")
let ints = try rows.map { $0[0] as Int }

// 🔥
let players = try OptimizedPlayer.fetchAll(db)

// 🐢
let rows = try Row.fetchAll(db, sql: "SELECT id, name FROM player")
let players = try rows.map { OptimizedPlayer(row: row) }
```


It is thus critical, when you aim at sheer fetching performance, that you:

- *never* prevent access to the fast extraction functions.
- *never* copy rows.
- avoid string lookup as much as possible.
- avoid Decodable as much as possible.
- read the documentation about [cursors](https://github.com/groue/GRDB.swift/blob/master/README.md#cursors).
- use the GRDB cursor that best fits your need.

groue:
> About these PR: The code performance is 30% faster with the initializer in this PR, than the initializer you are proposing.

Well done! Your generated code brings a big performance gain 👍 !

Thanks for your test project. Like you, I see that your change in the `StatementArguments` initializer brings an improvement. When I run test in the Release configuration, the improvement is not quite as big (see https://github.com/groue/GRDB-ORM/commits/genuine). In the below screenshot, `genInsert` spends 5.7% of its time in the regular `StatementArguments` initializer:

Capture d’écran 2021-02-07 à 20 12 53

Please consider again this above paragraph of mine:

> Mind that `StatementArguments` has a private implementation, which can change at any time in the future. It is thus not desirable to expose an "optimized" initializer unless we commit for the long run on such optimization.

By this I mean two things. First, I'd rather see performance improvements in GRDB itself, instead of exposing an implementation detail in the public API. Second, whenever GRDB wants to bring support to [sqlite3_bind_pointer or sqlite3_bind_value](https://www.sqlite.org/c3ref/bind_blob.html), the implementation of `StatementArguments` will change, and your library will lose its fragile advantage.

GRDB-ORM is already plenty fast for the specific use cases it generates code for! To make it even faster, even for requests that do not profit from its code generation, maybe efforts should be focused on GRDB itself.

Jasperav:
@groue Thanks for the very detailed heads-up about fetching through GRDB! 

I see the code changes you made on your own branch. The code looks cleaner with those changes 👍 . I will implement those changes by the code generator.

I added the 'Discussion' feature on GRDB-ORM: https://github.com/Jasperav/GRDB-ORM/discussions. I can ping you there when I implemented the code changes from your branch.

For now, I close the this PR since it won't get merged by the reasons you mentioned.

Issue #437, opened 2018-10-26
Feature Request: Failable initializers


cjarvi:
Hi there.  I'd like to request an enhancement to GRDB for your consideration: Failable initializers, specifically in the FetchableRecord protocol.  Perhaps something like init?(row:Row)

The use case for this is when storing/retrieving blob objects there's a chance for the Data() conversion to fail, especially when using custom de/serialization.  If this failure occurs, I'd like to be able to fail gracefully and not create the object.  

I attempted to roll my own, but found that I would need to override quite a few things to make it work...particularly anything that relies on the Record protocol.  It just didn't seem worth it at the time

groue:
Hello @cjarvi,

You're not the first requesting a failable or a throwing initializer. This is quite a valid request! But please follow me:

When a type can't decode a Row, it is often the sign of a database that contains bizarre values. Those databases are called "untrusted databases". They exist, I've met them! And they are not warmly welcomed in GRDB, as you have noticed.

Supporting untrusted database would come with problems that would weigh on the shoulders of all users, including those who deal with trusted databases. Not only a crippled API with more `try` and `if let` than ever, but lost opportunities for optimizations that are only possible when decoding is safe (such as deferred decoding, a technique used by some database observation tools).

A local database is not some JSON loaded from a remote server. The database is right there on your device: it is *your thing*.

This is why GRDB fosters trusted databases, and will continue doing so for the foreseeable future, for the benefit of all.

Now let's bring some possible solutions to your issue :-)

----

### First option: Keep the database in a clean state.

> The use case for this is when storing/retrieving blob objects there's a chance for the Data() conversion to fail, especially when using custom de/serialization

In your particular cases, it looks like [migrations](https://github.com/groue/GRDB.swift/blob/master/README.md#migrations) may well be your friends.

Before your record types can even think about fetching anything, give the database a good massage with a migration that decides what should be done with those invalid blobs: repair, destroy, mark them as invalid in a dedicated column. But don't let the migration quit until the database is in a pristine and controlled state.

After that, your records will be able to adopt FetchableRecord without a second thought, and this will greatly simplify your application logic.

----

### Second option: Let record types decode invalid rows

Keep the database in a dirty state, but make your records able to deal with it: have them tell if they could decode their database row, or not. Here is a very simplified example:

```swift
struct DatabaseUser: FetchableRecord {
    enum Content {
        case invalid
        case valid(name: String)
    }
    var content: Content
    
    init(row: Row) {
[...]

PR #920, opened 2021-02-06
Avoid decoding crashes

### LINKED ISSUE #320: How to handle failure in RowConvertible `init(row: Row)`?
# What did you do?

> Trying to make my object conform to RowConvertible.

## What did you expect to happen?

> Have a way to throw or return nil in case I cannot create my instance from a row.

## What happened instead?

> Can't find a way to propagate error from `init(row: Row)`. I think this can happen in cases like the database is corrupted. Marking `init(row: Row)` with `throws` result in `does not conform to protocol` error.

## Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 2.8.0
**Xcode version:** Version 9.2 (9C40b)
**Swift version:** 4
**macOS version running Xcode:** 10.13.3 (17D102)

## Demo Project

> Please link to or upload a project we can download that reproduces the issue.

groue:
This pull request makes it possible to catch database decoding errors. It is now possible to avoid crashes when dealing with untrusted, or uncontrolled, databases.

It contains a breaking change on the public `StatementColumnConvertible` protocol:

```diff
 public protocol StatementColumnConvertible {
-    init(sqliteStatement: SQLiteStatement, index: Int32)
+    init?(sqliteStatement: SQLiteStatement, index: Int32)
 }
```

However, since `StatementColumnConvertible` is not mentioned in the documentation, and thus barely public, chances are very low any user is affected by this change.

Fixes #320, #437


groue:
Nope, this won't do. The `Record` subclass would introduce too many breaking changes if its initializer would throw. Let's wait for GRDB 6!

PR #932, opened 2021-03-04
Fix crash in ReceiveValuesOnSubscription

jameshurst:
I managed to encounter a crash here in my application. Its probably better to just break here instead of crashing the application.

### DIFF
--- a/GRDB/Utils/ReceiveValuesOn.swift
+++ b/GRDB/Utils/ReceiveValuesOn.swift
@@ -120,7 +120,7 @@ where
                 }
                 
             case .waitingForRequest, .subscribed, .finished:
-                break
+                subscription.cancel()
             }
         }
     }
--- a/GRDB/Utils/ReceiveValuesOn.swift
+++ b/GRDB/Utils/ReceiveValuesOn.swift
@@ -120,7 +120,7 @@ where
                 }
                 
             case .waitingForRequest, .subscribed, .finished:
-                preconditionFailure()
+                break
             }
         }
     }


groue:
Hello @jameshurst,

I'm sorry you did not provide any detail or stack trace about this crash. *Something* is wacky here, and I'm entitled to say that the crash you have witnessed is deserved, and that you should look after its root cause.

Anyway: `ReceiveValuesOnSubscription` was introduced as a workaround for the bugs of the early Combine releases. I'll see if I can just get rid of it, now that almost nobody runs iOS 13.0.

jameshurst:
Unfortunately I was unable to create a sample application that reliably reproduced this crash. I did still encounter this crash once though while trying to create a reproducible case. The case in which this happened was when I had many pending database requests that would get cancelled, probably before a connection to the pool could be acquired.

I've seen similar errors before while implementing my own custom Combine subscriptions where Apple provided Combine operators would sometimes call methods in an unexpected order. Looking at how OpenCombine's RecieveOn publisher works (https://github.com/OpenCombine/OpenCombine/blob/master/Sources/OpenCombine/Publishers/Publishers.ReceiveOn.swift), when they receive a subscription while in an unexpected state, they immediately cancel the subscription, which would be favourable to crashing.

groue:
> I did still encounter this crash once though while trying to create a reproducible case.

Would you please share *something*, please?

Your suggestion to hide a problem without sufficient rationale is not acceptable, I'm sorry. If the problem is in GRDB, then something has to be done in order to prevent this situation (instead of working around it). If the problem is in the app, then the app has to be fixed (and it is not GRDB business). If the problem is in Combine, then we may have to work around it (maybe with your patch), but we'd also open a issue on https://feedbackassistant.apple.com, and properly comment this workaround in the GRDB code. None of those three cases are represented by this pull request only.

jameshurst:
Since you feel strongly about this I'll close this PR and continue to investigate and see if I can determine the root cause of the issue.

groue:
All right, @jameshurst! I'll be happy to keep on discussing the topic, and helping you on the way, when you feel like it. Mind I'll need something to chew on (sample code and stack trace) ;-)

jameshurst:
I've put together a sample application that is able to reliably reproduce the crash. After you run the app and let it perform fetches for a bit it will eventually crash. Looking at the logs it seems like a race condition in which `cancel()` gets called before `receive(subscription: Subscription)`. Note that the publisher chain doesn't really make sense and is just a way to reproduce this issue.

[GRDBBug.zip](https://github.com/groue/GRDB.swift/files/6093496/GRDBBug.zip)

groue:
Thank you very much for the sample project, @jameshurst. I can witness the precondition failure. 👍

groue:
OK @jameshurst, you have found a GRDB bug.

When `ReceiveValuesOnSubscription` receives a request, it subscribes to its upstream publisher, and waits for an upstream subscription. Meanwhile, it can get canceled. In this case, the awaited upstream subscription should indeed be cancelled.

> Anyway: ReceiveValuesOnSubscription was introduced as a workaround for the bugs of the early Combine releases. I'll see if I can just get rid of it, now that almost nobody runs iOS 13.0.

That sounds too risky.

I'll open a different pull request, because this one does not follow the [guidelines](https://raw.githubusercontent.com/groue/GRDB.swift/master/.github/PULL_REQUEST_TEMPLATE.md) that were shown to you when you initially opened it (particularly, PRs should be submitted against the development branch).

Thanks for the report and the time you spent helping identifying the problem! 👍

groue:
Closing in favor of #934

Issue #931, opened 2021-03-04
SQLITE_NOTFOUND (12) error when following the recommend "Sharing a Database"


RuiAAPeres:
### What did you do?

Setting up the AppContainer following this guide: https://github.com/groue/GRDB.swift/blob/master/Documentation/SharingADatabase.md

### What did you expect to happen?

I would expect the setup to successfully work.

### What happened instead?

There's a an error thrown here:

```swift
try v1Migrator.migrate(dbPool)
```

The `v1Migrator` looks like the following:

```swift
  var migrator = DatabaseMigrator()
  migrator.eraseDatabaseOnSchemaChange = true

  migrator.registerMigration("v1", migrate: { db in
    try db.create(table: "table0") { table in

    }

    try db.create(table: "table1") { table in
    }
  })

  return migrator
```

Calling the initial method `openSharedDatabase ` looks like this:

```swift
    do {
      let pathAppGroup = FileManager()
        .containerURL(forSecurityApplicationGroupIdentifier: groupIdentifier)!
        .appendingPathComponent("database.sqlite")

      let pool = try openSharedDatabase(at: pathAppGroup)
      self.pool = pool
    } catch {
      fatalError(error.localizedDescription)
    }
```

The actual error thrown:

```
▿ SQLite error 12: unknown operation
  ▿ extendedResultCode : 12 (unknown operation)
    - rawValue : 12
  ▿ message : Optional<String>
    - some : "unknown operation"
  - sql : nil
  - arguments : nil
```

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 5.3.0
**Installation method:** SPM
**Xcode version:** Version 12.4 (12D4e)
**Swift version:** 5.3.2
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:**

RuiAAPeres:
Removing `migrator.eraseDatabaseOnSchemaChange = true` no longer throws the error. But ideally I would like to continue iterating on the DB, so cleaning up would be ideal.

groue:
Hello @RuiAAPeres,

Thank you for the reports. I'm glad you found a workaround until the issue has been understood.

RuiAAPeres:
@groue do you see a workaround for this situation,  while keeping `migrator.eraseDatabaseOnSchemaChange = true`?

groue:
[...]

PR #933, opened 2021-03-06
Fix DatabaseMigrator.eraseDatabaseOnSchemaChange in the context of shared databases

groue:
The `DatabaseMigrator.eraseDatabaseOnSchemaChange` flag creates a temporary database in order to look for potential changes in database migrations, and nuke the database if needed.

[Temporary on-disk databases](https://www.sqlite.org/c3ref/open.html) created with an empty path [do not support the WAL mode](https://sqlite.org/forum/forumpost/f4f9d58515). They can make the migration fail when the database configuration relies on the availability of the WAL mode. This is the case, for example, in the sample code provided in the [Sharing a Database](https://github.com/groue/GRDB.swift/blob/master/Documentation/SharingADatabase.md) Guide: `SQLITE_FCNTL_PERSIST_WAL` is not available for such temporary databases, as revealed by #931.

This pull request fixes this problem, and #931, by having `DatabaseMigrator` create a "regular" temporary database that supports the WAL mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 
 #### 5.x Releases
 
+- [Development Branch](#development-branch)
 - `5.5.x` Releases - [5.5.0](#550)
 - `5.4.x` Releases - [5.4.0](#540)
 - `5.3.x` Releases - [5.3.0](#530)
@@ -70,6 +71,10 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 
 - [0.110.0](#01100), ...
 
+
+## Development Branch
+
+- **Fixed**: [#933](https://github.com/groue/GRDB.swift/pull/933): Fix DatabaseMigrator.eraseDatabaseOnSchemaChange in the context of shared databases
 
 ## 5.5.0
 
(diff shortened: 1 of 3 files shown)


groue:
@RuiAAPeres, in https://twitter.com/peres/status/1367773340560134145 you write:

> This crash only manifested on release builds, not debug.

This does not match my experience and understanding. I could witness the SQLITE_NOTFOUND error in debug builds as well. I do hope this pull request fixes your issue, but I may have missed something. Would you please test your app against this pull request, in release configuration, before we consider the issue is solved, and I can ship a new GRDB version? That would be very helpful of you.

RuiAAPeres:
@groue 

> This crash only manifested on release builds, not debug.

To be completely fair, there's a chance no migration was run when I was testing and the problem only surfaced when TF testers tried the app for the first time.

groue:
Thank you @RuiAAPeres. I stop wondering about this issue :-)

This pull request will shortly ship. Meanwhile, a comment in #931 contains the workaround you need.

RuiAAPeres:
Do you want me to test the fix that's in development? Or should I wait for the official release?

groue:
You can test with the development branch, sure. It would be just as if you were a beta tester, and I will never prevent you from doing so 😉 

But I'm pretty confident that this PR has solved your issue. With it, you can remove the workaround, and keep both `eraseDatabaseOnSchemaChange`, and your initial shared database preparation code.

I'd feel better if you would put `eraseDatabaseOnSchemaChange = true` inside `#if DEBUG`, though. I'd really not want this flag to destroy your user data if something unexpected has the migrator spot a schema difference where it should not. In development releases, this is harmless. In production, this is... I prefer not thinking about it.

PR #934, opened 2021-03-06
Fix a crash in the ValueObservation publisher

groue:
This pull request fixes #932 as suggested by @jameshurst.


groue:
@jameshurst, the fix has shipped in v5.6.0!

PR #935, opened 2021-03-06
make SQLOrdering helper functions public

danramteke:
### Pull Request Checklist

I would like to use these helper functions on `SQLOrdering`. However, they are `internal` and cannot be used by clients of the library.

- [x] Changes are tested.

### DIFF
--- a/GRDB/QueryInterface/SQL/SQLOrdering.swift
+++ b/GRDB/QueryInterface/SQL/SQLOrdering.swift
@@ -29,27 +29,27 @@ public struct SQLOrdering {
         case literal(SQLLiteral)
     }
     
-    static func expression(_ expression: SQLExpression) -> SQLOrdering {
+    public static func expression(_ expression: SQLExpression) -> SQLOrdering {
         self.init(impl: .expression(expression))
     }
     
-    static func asc(_ expression: SQLExpression) -> SQLOrdering {
+    public static func asc(_ expression: SQLExpression) -> SQLOrdering {
         self.init(impl: .asc(expression))
     }
     
-    static func desc(_ expression: SQLExpression) -> SQLOrdering {
+    public static func desc(_ expression: SQLExpression) -> SQLOrdering {
         self.init(impl: .desc(expression))
     }
     
-    static func ascNullsLast(_ expression: SQLExpression) -> SQLOrdering {
+    public static func ascNullsLast(_ expression: SQLExpression) -> SQLOrdering {
         self.init(impl: .ascNullsLast(expression))
     }
     
-    static func descNullsFirst(_ expression: SQLExpression) -> SQLOrdering {
+    public static func descNullsFirst(_ expression: SQLExpression) -> SQLOrdering {
         self.init(impl: .descNullsFirst(expression))
     }
     
-    static func literal(_ sqlLiteral: SQLLiteral) -> SQLOrdering {
+    public static func literal(_ sqlLiteral: SQLLiteral) -> SQLOrdering {
         self.init(impl: .literal(sqlLiteral))
     }
 }


groue:
Hello @danramteke,

Those methods are not public because `SQLOrdering`, along with `SQLExpression`, `SQLSelection` and a few other types belong to the inner layer of the query interface that is only exposed through protocols: `SQLOrderingTerm`, `SQLExpressible`, `SQLSpecificExpressible`, `SQLSelectable`, etc.

One does not build an `SQLExpression` or an `SQLOrdering` directly. Instead, one uses ready-made types such as `Column`, and feeds methods that accept protocol values:

```swift
Player
    // feed select(_ selection: SQLSelectable...)
    .select(Column("name"))
    // feed filter(_ predicate: SQLSpecificExpressible)
    .filter(Column("score") >= 1000)
    // feed order(_ orderings: SQLOrderingTerm...)
    .order(Column("name").desc)
```

That's the intent.

In particular, the `asc`, `desc` etc properties are exposed on `SQLSpecificExpressible`, the protocol for "genuine" SQL expressions (excluding expressions like `42` or `Date()` that only conform to `SQLExpressible`):

```swift
Column("name").desc
(Column("score") + Column("bonus")).asc
```

Do you happen to be in a situation where you can not access those public `asc`, `desc` properties? Can we discuss this? I'd rather help you refactor bits of your code, or modify GRDB at the "proper" level, rather than breaking the intent described above by exposing public `SQLOrdering` constructors.

danramteke:
Hmm good points! And thanks for the explanation. I've been using this library for a long time now over at https://underway.nyc

Yeah my main reason I need to use literals for columns that belong to a different table (that are JOINed). If I don't use a literal, it seems that GRDB thinks the column belongs to the main table in the `SELECT` statement. Therefore, I use `SQLLiteral` to say things like `\"ServiceAlertInformedEntities\".\"route\"`.

So two questions then:

A) How do I create an `SQLOrderingTerm` from an `SQLLiteral`?
B) How can I refer to a column on a different table without using `SQLLiteral`?


And here is sample code:

A `ServiceAlert` has many `ServiceAlertInformedEntities`. In [GTFS](https://developers.google.com/transit/gtfs), an "Informed Entity" contains information about which entities are affected by this service alert (stations, subways, etc) :

```
  private func routeOrderingIfNeeded(_ orderByRoute: Bool) -> SQLOrdering? {
    if #available(iOS 14, *) {
      return orderByRoute
        ? SQLOrdering.literal("\"ServiceAlertInformedEntities\".\"route\" ASC NULLS LAST")
//        ? ServiceAlert.InformedEntity.Table.CodingKeys.route.ascNullsLast
        : nil
    } else {
      return orderByRoute
        ? SQLOrdering.literal("\"ServiceAlertInformedEntities\".\"route\" ASC")
//        ? ServiceAlert.InformedEntity.Table.CodingKeys.route.asc
        : nil
    }
  }
```


If I try to use `Column` like in your example,  i get an error `no such column: ServiceAlerts.ServiceAlertInformedEntities.route`
```
  private func includeJoinsExpr(_ orderByRoute: Bool) -> SQLOrderingTerm? {
    let c: Column = Column("ServiceAlertInformedEntities.route")

    if #available(iOS 14, *) {
      return c.ascNullsLast
    } else {
      return c.asc
    }
  }
```

danramteke:
You used to be able to use `SQLExpression` in `.order()`. It seems recently there's been some API changes around this.

groue:
Thanks for being a faithful GRDB user :-)

> You used to be able to use `SQLExpression` in `.order()`. It seems recently there's been some API changes around this.

You can. `SQLExpression` conforms to `SQLOrderingTerm` (expressions can be used as an ordering term: http://www.sqlite.org/syntax/ordering-term.html).

> Yeah my main reason I need to use literals for columns that belong to a different table (that are JOINed). If I don't use a literal, it seems that GRDB thinks the column belongs to the main table in the SELECT statement.

All right. It is correct that literals are not processed (there is no SQL parser in GRDB), and that they are a way to inject raw SQL in requests.

Your fixed sample code is:

```swift
private func routeOrderingIfNeeded(_ orderByRoute: Bool) -> SQLOrdering? {
    guard orderByRoute else { return nil }
    // Use a literal in order to force the table.
    // Build an expression so that `asc` and `ascNullsLast` are available.
    let route = SQLLiteral("ServiceAlertInformedEntities.route").sqlExpression
    if #available(iOS 14, *) {
        return route.ascNullsLast
    } else {
        return route.asc
    }
}
```

Note how the method return an `SQLOrdering`. Why? Because most GRDB apis *accept* protocol values, but *return* values of type `SQLExpression`, `SQLOrdering`, etc. Let's just do the same.

---

I'll remind below how columns are dispatched to their tables, so that you can think, when you have time, at a way to avoid literals:

1. Columns are dispatched to their contextual table:

    ```swift
    // In a request of Player, columns are associated to the player table.
    let playersRequest = Player.order(Column("score"))

    // In an association to Team, columns are associated to the team table.
    let teamAssociation = Player.team.order(Column("name"))
    
    // SELECT * FROM player JOIN team ON team.id = player.teamID
    // ORDER BY player.score, team.name
    let joinedRequest = playersRequest.including(required: teamAssociation)
    ```

2. When one wants to dispatch a column to another table, one uses a [TableAlias](https://github.com/groue/GRDB.swift/blob/master/Documentation/AssociationsBasics.md#table-aliases).

    This helps us reversing the ordering terms in the above joined request:

    ```swift
    // SELECT * FROM player JOIN team ON team.id = player.teamID
    // ORDER BY team.name, player.score
    let teamAlias = TableAlias()
    let joinedRequest = Player
        .including(Player.team.aliased(teamAlias))
        .order(
            teamAlias[Column("name")], // Escape the rule 1
            Column("score")) // Rule 1: a player column
    ```

---

Eventually, I don't think this pull request is necessary or reveals a hole in available public APIs.

danramteke:
Thanks so much for the explanation!

PR #940, opened 2021-03-13
Prevent ValueObservation from notifying duplicate initial values when no concurrent write is observed

### LINKED ISSUE #937: ValueObservation notifies duplicate initial value when started from a DatabasePool
### What did you do?
I just updated from GRDB 4 + GRDBCombine to GRDB 5. 
It was surprisingly painless and the only code change I had to make was to replace the `fetchOnSubscription()` calls with `scheduling: .immediate` when creating the Combine publishers. 

I expected identical notification behavior as before. That is, one initial value notification right away, and then one notification for each observed value change after that. 

### What happened instead?

The initial value is being sent twice. 

Dealing with the duplicate notification is trivial. 
However the fact that the SQL query is executed twice can be a potentially significant performance regression.
(In my application this is an issue on the watch)

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 5.5.0
**Xcode version:** 12.5 beta 3
**Swift version:** 5
**Platform(s) running GRDB:** iOS and watchOS
**macOS version running Xcode:**

groue:
When one starts a `ValueObservation` in a `DatabasePool`, one would get a duplicate notification of the initial value.

Now this duplicate notification is avoided when we can prove that no transaction was committed, in the pool's writer connection, between the fetch of the initial value, and the first access to the writer connection (where the observation can really start by hooking on the SQLite commits).

A double notification will still happen if some writer commit is attempted during the ValueObservation start-up (even if the observed value is not impacted by the changes, and even if the commit fails).

As [documented](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservation-usage), ValueObservation users must be ready for duplicate notifications. This pull request produces *less* duplicate notifications, but does not remove all of them.

Fix #937

### DIFF
--- a/GRDB/Core/DatabasePool.swift
+++ b/GRDB/Core/DatabasePool.swift
@@ -873,6 +873,13 @@ extension DatabasePool: DatabaseReader {
         // tracked database region (we have no way to know).
         //
         // 3. Install the transaction observer.
+        //
+        // When `SQLITE_ENABLE_SNAPSHOT` is not available, we perform a similar
+        // but weaker check. We compare the number of transaction commits
+        // performed between this method, and the first access to the writer
+        // database. Whenever this number has changed, we perform the second
+        // fetch, even if the actual changes are unrelated to the tracked
+        // database region (we have no way to know).
         
         #if SQLITE_ENABLE_SNAPSHOT
         if scheduler.immediateInitialValue() {
@@ -906,11 +913,16 @@ extension DatabasePool: DatabaseReader {
                 }
         }
         #else
+        // Grab `attemptedCommitCount` *before* reader transaction has started,
+        // so that it describes a state of the database from before the
+        // beginning of snapshot isolation. We are guaranteed to read an initial
+        // value from this state, or from a later state.
+        let attemptedCommitCount = self.writer.attemptedCommitCount
         if scheduler.immediateInitialValue() {
             do {
                 let initialValue = try read(observer.fetchInitialValue)
                 onChange(initialValue)
-                addObserver(observer: observer)
+                addObserver(observer: observer, from: attemptedCommitCount)
             } catch {
                 observer.complete()
                 observation.events.didFail?(error)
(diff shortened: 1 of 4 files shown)


pocketpixels:
Thank you for tackling this. This solution looks good to me.

Just to make sure this is 100% guaranteed to be correct (I wouldn't care for my application, but I assume you probably would for general use): 
Can we completely rule out that a commit that happens right around the time of reading the `attemptedCommitCount` pre-fetch could sneak through undetected? 
In particular: 
The `sqlite3_commit_hook` gets called right before the write transaction gets committed, right? 
Let's say in the commit hook the `attemptedCommitCount` just got incremented but the hook/callback has not returned yet. 
Immediately after the count is read from the reader connection. 
Could it be possible that the initial fetch now happens and reads a version of the data that does not yet reflect the write transaction? Or does SQLite guarantee that his can never happen?

groue:
> Just to make sure this is 100% guaranteed to be correct (I wouldn't care for my application, but I assume you probably would for general use):
> Can we completely rule out that a commit that happens right around the time of reading the `attemptedCommitCount` pre-fetch could sneak through undetected?

Thanks for the help 💯 

> In particular:
> The `sqlite3_commit_hook` gets called right before the write transaction gets committed, right?

Correct.

> Let's say in the commit hook the `attemptedCommitCount` just got incremented but the hook/callback has not returned yet.

OK. Let's say `attemptedCommitCount` is 1, but database is still at 0. I see you coming...

> Immediately after the count is read from the reader connection.
> Could it be possible that the initial fetch now happens and reads a version of the data that does not yet reflect the write transaction? Or does SQLite guarantee that his can never happen?

You're absolutely right. The read can fetch from state 0 (when commit is not performed yet), and when we later enter the writer queue, the commit has been performed, we compare 1 to 1 and skip the second fetch 🐛 

All right, back to the drawing board! Thank you very much again!!

Issue #945, opened 2021-03-23
Tables without rowid causes "no such column"-logs


simme:
### What did you do?

Created a table without rowid.
```swift
db.create(table: "product_categories", withoutRowID: true)
```

### What did you expect to happen?

Nothing special really.

### What happened instead?

Everytime I access the database through my `FetchableRecord`s I get these log entries.
`no such column: rowid in "SELECT rowid FROM "list_items""`

The reason my tabels are without rowid is because I use UUIDs to identify everything. Perhaps this is the wrong approach?

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:**  5.6.0
**Installation method:** SPM
**Xcode version:** 12.3
**Swift version:** 5.3
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 11.2.3

### Demo Project

If necessary for further debugging I can certainly provide this! :)

groue:
Hello @simme,

I know GRDB has some blind spots about "without rowid" tables, but it is not intended that "everytime [one] accesses the database" an error happens.

Can you be more specific (and tell precisely about errors, instead of logging errors away)?

simme:
Hey!

Sorry if I was unclear. There's no error per se. Just the log statement. Everything seems to work fine.
I'll see if I can narrow down the cause a bit. But I'm pretty sure it's related to starting a `ValueObservation`. Like so:

```swift
    ValueObservation
      .tracking { db -> [ProductCategoryModel] in
        try ProductCategoryModel.fetchAll(db)
      }
      .publisher(in: pantry.databasePool).map {
        $0.map { ProductCategory(model: $0, locale: .current) }
      }
      .replaceError(with: [])
      .eraseToAnyPublisher()
```

simme:
I'm in the middle of the last 10% of the app I'm working on at the moment, so not a ton of extra time. Since this only leads to the log message and everything seems to work there's no panic. Maybe everything is working as intended, but I'm holding it wrong :)

groue:
OK. Do you happen to reproduce with a simple: 

```
try pantry.databasePool.read { db in
    try ProductCategoryModel.fetchAll(db)
}
```

What I ask here is your help building a *minimal* reproducing case, if you have time.

simme:
That does not cause the log statement.

Sorry for the overly inclusive statement in the initial ticket. I do all my fetching through `ValueObservation`s so I was sort of right that it's `Everytime I access my database`.

[...]

Issue #42, opened 2016-04-25
Crash during deinit of DatabaseQueue


ptseng:
This commit introduces deinit for DatabaseQueue.
https://github.com/groue/GRDB.swift/commit/005c4ded2408611789e01829295b10b40d1c2af3

Which gives me the following crash on deinit:

```
*** NSForwarding: warning: object 0x7f96e526a550 of class 'GRDB.DatabaseQueue' does not implement methodSignatureForSelector: -- trouble ahead
Unrecognized selector -[GRDB.DatabaseQueue FXNotifications_observers:]
```

Perhaps DatabaseQueue should also inherit from NSObject?

groue:
Hello @ptseng. I can not reproduce your issue. Do you have more detail?

ptseng:
Podfile:

```
platform :ios, '8.0'
use_frameworks!

pod 'GRDB.swift'
pod 'FXNotifications'
```

ViewController.swift:

```
import UIKit
import FXNotifications
import GRDB

class ViewController: UIViewController {
    
    var queue = [Int:DatabaseQueue]()

    override func viewDidLoad() {
        super.viewDidLoad()
        queue[1] = DatabaseQueue()
        queue[2] = DatabaseQueue()
        queue[1] = nil
        queue[2] = nil
    }
}
```

Making DatabaseQueue a subclass of NSObject seems to fix the crash bug.

groue:
Wow.

I did spend a few minutes setting up a new project with your Podfile, and pout your code in a plain ViewController, nothing else. And I get a crash... related to FXNotifications.

I suggest that you open an issue in FXNotifications instead. GRDB has no problem.

ptseng:
@groue I will post this issue with FXNotifications.  However, I think this bug was introduced with the new (26 days) use of deinit() in DatabaseQueue.  Please reference and consider reopening this issue: 

http://stackoverflow.com/questions/24415662/object-x-of-class-y-does-not-implement-methodsignatureforselector-in-swift

ptseng:
https://github.com/nicklockwood/FXNotifications/issues/9

groue:
@ptseng You're totally right, this crash was introduced by the `NSNotificationCenter.defaultCenter().removeObserver(self)` line added in DatabaseQueue.deinit in commit https://github.com/groue/GRDB.swift/commit/005c4ded2408611789e01829295b10b40d1c2af3.

[...]

PR #946, opened 2021-03-23
Prohibit observation of tables created WITHOUT ROWID

simme:
As discussed in #945 this PR adds a fatal error whenever a developer attempts to observe a table that lacks a rowid. Since SQLite does not report changes to these tables, notifying the developer of this makes sense.

The method on `DatabaseRegion` that performs the check loops all table regions and checks if those tables have a `rowid` and returns the names of tables that don't. First I had a variant that returned a `bool` as soon as a table without `rowid` was found. Thinking we don't need to check more than that. But it struck me that for valid regions the loop will always check every table anyway. So it felt better to instead gather up the names of the offending tables, if more than one, to provide better error reporting to the user.

I am not 100% confident that the single test case covers all bases.

I have not updated the README as I think maybe the documentation as is probably is enough together with this change.

And of course, let me know if you're not happy with the solution. You probably know if there are better ways to do it :)

### Pull Request Checklist

- [x] Changes are tested. **Two other tests in the suite failed, but they seemed unrelated to this change**

### DIFF
--- a/GRDB/Core/DatabaseRegion.swift
+++ b/GRDB/Core/DatabaseRegion.swift
@@ -169,6 +169,20 @@ public struct DatabaseRegion: CustomStringConvertible, Equatable {
         }
         return DatabaseRegion(tableRegions: filteredRegions)
     }
+
+    /// Returns an array containing the names of any tables in the region that
+    /// were created with WITHOUT ROWID.
+    func tablesWithoutRowid(_ db: Database) throws -> [String] {
+        guard let tableRegions = tableRegions else { return [] }
+        var tables: [String] = []
+        for (table, _) in tableRegions {
+            let pkInfo = try db.primaryKey(table)
+            if !pkInfo.tableHasRowID {
+                tables.append(table)
+            }
+        }
+        return tables
+    }
 }
 
 extension DatabaseRegion {
--- a/GRDB/ValueObservation/ValueObserver.swift
+++ b/GRDB/ValueObservation/ValueObserver.swift
@@ -221,13 +221,15 @@ extension ValueObserver {
 
         let withoutRowid = try observedRegion?.tablesWithoutRowid(db) ?? []
         if !withoutRowid.isEmpty {
+            // https://www.sqlite.org/withoutrowid.html
+            // > The sqlite3_update_hook() interface does not fire callbacks for
+            // > changes to a WITHOUT ROWID table.
+            //
+            // This prevents us from observing those tables, and there is no
+            // workaround. It is thus a programmer error:
             fatalError("""
-                The sqlite3_update_hook() interface does not fire callbacks for changes to a WITHOUT ROWID table.
-                This observation contains the following tables that cannot be observed:
-                    \(withoutRowid.joined(separator: ", "))
-
-                See https://www.sqlite.org/withoutrowid.html for more information.
-            """)
+                Tracking WITHOUT ROWID table is not supported: \(withoutRowid.joined(separator: ", "))
+                """)
         }
 
         return result
--- a/GRDB/ValueObservation/ValueObserver.swift
+++ b/GRDB/ValueObservation/ValueObserver.swift
@@ -218,7 +218,18 @@ extension ValueObserver {
         // TransactionObserver protocol. By removing them from the observed
         // region, we optimize our TransactionObserver conformance.
         observedRegion = try region.ignoringViews(db).ignoringInternalSQLiteTables()
-        
+
+        let withoutRowid = try observedRegion?.tablesWithoutRowid(db) ?? []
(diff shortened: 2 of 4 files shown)


groue:
Thank you @simme!

> So it felt better to instead gather up the names of the offending tables, if more than one, to provide better error reporting to the user.

I very much agree!

> I am not 100% confident that the single test case covers all bases.

It looks quite good to me.

> I have not updated the README as I think maybe the documentation as is probably is enough together with this change.

Yes. I can take care of it.

> Two other tests in the suite failed, but they seemed unrelated to this change

They do not fail without the new check in `ValueObserver`, so they are (indirectly) related ;-) Do you feel like investigating? I'll provide help if needed. And if you don't have time, I'll complete this excellent pull request.

groue:
Would you please allow me to push to your branch?

groue:
Thanks ;-)

simme:
> They do not fail without the new check in ValueObserver, so they are (indirectly) related ;-) Do you feel like investigating? I'll provide help if needed. And if you don't have time, I'll complete this excellent pull request.

Oh. Could've sworn I ran the tests without my changes and got the same. Will double check.

> Would you please allow me to push to your branch?

✅

---

I'll be able to look more on this tomorrow.

simme:
I've tried a few different things. My conclusion is that the `testWriteObservation()` test fails with `caught error: "SQLite error 1: no such table: temp"`. Commenting these lines in my new method makes the test pass.
```swift
let pkInfo = try db.primaryKey(table)
if !pkInfo.tableHasRowID {
  tables.append(table)
}
```.

I think I'm lacking a little bit of insight into how SQLite works in this case.

This is the observation in the test:
```swift
var observation = ValueObservation.trackingConstantRegion { db -> Int in
    XCTAssert(db.isInsideTransaction, "expected a wrapping transaction")
    try db.execute(sql: "CREATE TEMPORARY TABLE temp AS SELECT * FROM t")
    let result = try Int.fetchOne(db, sql: "SELECT COUNT(*) FROM temp")!
    try db.execute(sql: "DROP TABLE temp")
    return result
}
```

Could it be something with the accessing of the database when checking the primary keys on a table that somehow makes the temporary table disappear?

groue:
This test makes sure users can define observations that write (#42). I don't remember why I put a temporary table there. But I'm happy I did :-)

When I run this test with a breakpoint on all Swift errors, I can see that since the observed region contains both the `temp` and `t` tables, the new code attempts at making sure none of them is WITHOUT ROWID, and loads the primary key for the "temp" table. `Database.primaryKey(_:)` calls `Database.columns(in:)`, which does not support temp tables: it throws "no such table" instead.

Capture d’écran 2021-03-24 à 10 01 33

One can wonder if it is useful to track [temporary tables](https://www.sqlite.org/tempfiles.html#temp_databases) as all:

> Tables created using the "CREATE TEMP TABLE" syntax are only visible to the database connection in which the "CREATE TEMP TABLE" statement is originally evaluated.

Since `DatabasePool` has ValueObservation fetch from one unspecified reader connection from its pool, it's impossible to guarantee that it will see a temp table...

But `DatabaseQueue` and `ValueObservation` with the `requiresWriteAccess` can do it (since both have the observation fetch from the single writer connection)...

A possible fix is to have `Database.primaryKey(_:)` able to run on temp tables. According to https://sqlite-users.sqlite.narkive.com/QOXQJwBh/accessing-temporary-tables, we need to prefix the table name with 'temp' when we run the `table_info` and `table_xinfo` pragmas.

I can't make up my mind right now. I'll come back. Your ideas are of course welcomed.

simme:
If it's fairly straight forward to detect a temporary table and if so update `Database.primaryKey(_:)` to work on them that seems like the correct fix. But the whole use case do seem very edge-casey, might be a balance where the extra overhead and maintenance cost is not worth it?

That said, I don't know what's right 🤷🏼‍♂️

groue:
Yes, I was having a look, and starting to think that cleaning up the way GRDB handles [schemas](https://sqlite.org/lang_naming.html) first (the default "main", the built-in "temp", and custom ones with attached databases) should be done first.

groue:
@simme, here is a temporary conclusion:

- we do not merge this pull request today
- I update the doc so that users understand that WITHOUT ROWID tables are not tracked

And we'll revive this pull request in the future, if we're ready and still think it's worth it.

Is it OK for you?

simme:
Yep, absolutely!
My initial issue was fixed by not using WITHOUT ROWID, which was a pre-mature optimization that I thought would simplify things. Let me know if there's anything else I can do to be of assistance :)

groue:
I'm happy we could explore a little bit a new topic! Thanks for the short but enjoyable journey, @simme. See you soon, happy GRDB!

Issue #944, opened 2021-03-21
Possible memory issue


MartinP7r:
First of all thanks for this awesome library.
I have an issue 
### What did you do?

Trying to persist/update a sizable amount of records (~180,000). 

### What did you expect to happen?

Records being persisted/updated.

### What happened instead?

Device crash at memory usage of 2+ GB

### Environment

**GRDB flavor(s):**
`GRDB.swift` with FTS5 via cocoapods
```ruby
pod 'GRDB.swift', :git => 'https://github.com/groue/GRDB.swift.git'
  
  post_install do |installer|
    installer.pods_project.targets.select { |target| target.name == "GRDB.swift" }.each do |target|
      target.build_configurations.each do |config|
        config.build_settings['OTHER_SWIFT_FLAGS'] = "$(inherited) -D SQLITE_ENABLE_FTS5"
      end
    end
  end
```
**GRDB version:** 5.2.0 (I've updated to 5.6.0 and will run below instruments again)
**Installation method:** pods, see above
**Xcode version:** 12.2
**Swift version:** 5.2
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 10.15.7

### Demo Project

I'm sorry I can't upload a demo project at the moment, but if you can spare the time and give me your two cents on my approach below, that'd be already really helpful I hope.

Basically all I'm doing is 

```swift
func persist(items: [T]) {
    l.info("persisting items for Type \(T.self)")
    do {
        for item in items {
            try db.dbQueue.write { db in
                try item.save(db)
            }
            db.dbQueue.releaseMemory()
        }
    } catch {
        l.error(error)
    }
}
```
for an array of ~180,000 items in one table.

The items are being parsed from an xml file (100mb) beforehand which is already quite hefty and they take up about 800mb of memory.

I've been wondering if I should just persist the data in smaller chunks. Keeping all the parsed structs in memory seems like it's probably just too much. So splitting the xml data up in smaller sizes and going from there is probably my only option?

groue:
Hello @MartinP7r,

In my experience, the following loop consumes a stable amount on memory:

```swift
let dbQueue = try DatabaseQueue(path: ...)

try dbQueue.write { db in
    try db.create(table: "player") { t in
        t.autoIncrementedPrimaryKey("id")
        t.column("name", .text).notNull()
        t.column("score", .integer).notNull()
    }
}

struct Player: Encodable, PersistableRecord {
    var id: Int64
    var name: String
    var score: Int
[...]

PR #950, opened 2021-04-01
Fix memory consumption when encoding JSON columns

MartinP7r:
This PR is meant to fix the issue shown in #944 where encoding json columns leads to significant memory buildup.

What I found is that the memory buildup can be prevented by simply wrapping it into an autoreleasepool. However, I have not much experience with that mechanic, so please be the judge of whether or not this is the right way to go.

--- 

As a possible reason for the buildup I found that the a new JSONEncoder is initiated for every json column of every item being persisted: 
https://github.com/groue/GRDB.swift/blob/67892d7cdb7d16a56bd975a091e9b427d7115270/GRDB/Record/EncodableRecord.swift#L127-L137

Since `databaseJSONEncoder(for column: String)` doesn't seem to actually use its parameter `column` this could maybe actually be implemented as a lazy computed property (or maybe cache the encoder elsewhere?).

Sadly I couldn't directly see a way around this without changing the `EncodableRecord` protocol implementation since the extension doesn't allow stored properties.

Would be happy to hear your thoughts on this!

---

### Pull Request Checklist

- [ ] Changes are tested.

Existing tests seem to run fine and it doesn't seem like any additional ones are necessary.
I also can't see if there's any need to change the documentation. I would like to leave that decision up to you.

### DIFF
--- a/GRDB/Record/EncodableRecord+Encodable.swift
+++ b/GRDB/Record/EncodableRecord+Encodable.swift
@@ -148,16 +148,19 @@ private class RecordEncoder<Record: EncodableRecord>: Encoder {
                 }
             } catch is JSONRequiredError {
                 // Encode to JSON
-                let jsonData = try Record.databaseJSONEncoder(for: key.stringValue).encode(value)
-                
-                // Store JSON String in the database for easier debugging and
-                // database inspection. Thanks to SQLite weak typing, we won't
-                // have any trouble decoding this string into data when we
-                // eventually perform JSON decoding.
-                // TODO: possible optimization: avoid this conversion to string,
-                // and store raw data bytes as an SQLite string
-                let jsonString = String(data: jsonData, encoding: .utf8)!
-                persist(jsonString, forKey: key)
+                try autoreleasepool {
+
+                    let jsonData = try Record.databaseJSONEncoder(for: key.stringValue).encode(value)
+
+                    // Store JSON String in the database for easier debugging and
+                    // database inspection. Thanks to SQLite weak typing, we won't
+                    // have any trouble decoding this string into data when we
+                    // eventually perform JSON decoding.
+                    // TODO: possible optimization: avoid this conversion to string,
+                    // and store raw data bytes as an SQLite string
+                    let jsonString = String(data: jsonData, encoding: .utf8)!
+                    persist(jsonString, forKey: key)
+                }
             }
         }
     }


groue:
Hello @MartinP7r,

> What I found is that the memory buildup can be prevented by simply wrapping it into an autoreleasepool. However, I have not much experience with that mechanic, so please be the judge of whether or not this is the right way to go.

This pretty much looks like a very sensible pull request. `JSONEncoder` uses [NSJSONSerialization](https://developer.apple.com/documentation/foundation/nsjsonserialization?language=objc) in its implementation. As many Objective-C objects do, it may allocate temporary objects in an autorelease pool. So memory will eventually be released. But in a tight loop of many JSON encodings, those temporary objects may just add up.

Since your pull request wraps the only use of this JSON encoder, it is correct 👍

But is is useful? You [wrote](https://github.com/groue/GRDB.swift/issues/944#issuecomment-811811271):

> I think I found a possible cause (and maybe solution) and started a PR #950 if that's ok.

Could you please check if this pull request fixes your memory issue? 😉 

We'll address your other questions then, if this is ok for you.

MartinP7r:
> But is it useful? You wrote:
> > I think I found a possible cause (and maybe solution) and started a PR #950 if that's ok.

> Could you please check if this pull request fixes your memory issue? 😉

I'm sorry, my post was structured badly and might be a little confusing. The PR's change with `autoreleasepool` fully solves the memory issue for the example in my comment in #944 :

```swift
let dbQueue = try DatabaseQueue(path: ...)

try dbQueue.write { db in
        t.autoIncrementedPrimaryKey("id")
        t.column("team", .text).notNull()
    }
}

struct Player: Encodable, PersistableRecord {
    var id: Int64
    var team: Team
}

struct Team: Encodable {
    var id: Int64
    var name: String
}

try dbQueue.write { db in
    for i in 1..<180_000 {
        let player = Player(id: Int64(i), team: Team(id: Int64(i), name: "Team \(i)"))
        try player.save(db)
    }
}
```

It uses a constant 10~20mb of memory after the change. 
That's actually the `maybe solution` I was talking about. However I didn't want to presume, rather hear your opinion on it since I don't have much experience with `autoreleasepool()`.

The part after `As a possible reason` is just referring to the possible root cause in `databaseJSONEncoder(for column: String)` (I didn't investigate to far since `autoreleasepool()` solves my problem) and maybe other solutions (like caching) that might work if `autoreleasepool()` is not acceptable.

Sorry for making this maybe more complicated than it needs to. For me the current PR's content fixes the memory problem at least for the minimal reproducible example above.

groue:
All right, that's great, then! I merge your pull request.

> I found that the a new JSONEncoder is initiated for every json column of every item being persisted

The default implementation of `databaseJSONEncoder(for:)` could return a shared, statically allocated, JSONEncoder. But until we find an actual reason for performing such a change, I suggest we just leave this code as it is.

> Existing tests seem to run fine and it doesn't seem like any additional ones are necessary.

Yes, this is fine.

> I also can't see if there's any need to change the documentation. I would like to leave that decision up to you.

Your enhancement has the library behave as it should have been behaving right from the start, and does not need any particular notice in the documentation.

Welcome to the GRDB contributors, @MartinP7r!

Would you accept an invitation that grants you a push access to GRDB? I do intend to keep the lead of the project, but it is good for everyone that people of various levels of Swift and SQLite expertise can act on the repository when it becomes necessary: you'll be very welcome!

MartinP7r:
Thank you, @groue ! I'm glad that I was able to contribute something. 🙂

> Welcome to the GRDB contributors, @MartinP7r!
> Would you accept an invitation that grants you a push access to GRDB? I do intend to keep the lead of the project, but it is good for everyone that people of various levels of Swift and SQLite expertise can act on the repository when it becomes necessary: you'll be very welcome!

Thank you, I'd gladly help in any capacity my time and skill level allows me to!
Also, I appreciate the opportunity to learn from you and other contributors. Thanks again for your advice so far!

groue:
Invitation sent :-)

PR #967, opened 2021-04-22
Make SelectStatement conform to CustomStringConvertible

steipete:
This PR makes it easier to use `po statement` in the debugger to read SQL.

### DIFF
--- a/GRDB/Core/Statement.swift
+++ b/GRDB/Core/Statement.swift
@@ -985,3 +985,9 @@ extension StatementArguments {
         return "[" + (namedValuesDescriptions + valuesDescriptions).joined(separator: ", ") + "]"
     }
 }
+
+extension SelectStatement: CustomStringConvertible {
+    public var description: String {
+        "SQL: \(sql), Arguments: \(arguments)"
+    }
+}
--- a/GRDB/Core/Statement.swift
+++ b/GRDB/Core/Statement.swift
@@ -19,6 +19,8 @@ public class Statement {
     
     /// The SQL query
     public var sql: String {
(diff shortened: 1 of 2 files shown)


groue:
Hello @steipete!

I much welcome this idea! I have three suggestions:

- Let's extend `Statement` instead of `SelectStatement`, so that `UpdateStatement` can be described as well.

- Let's make the [`Statement.sql`](https://github.com/groue/GRDB.swift/blob/31e2b01c10243399882eaf8cc6b5c974af93d21e/GRDB/Core/Statement.swift#L21) property thread-safe, by adding `SchedulingWatchdog.preconditionValidQueue(database)` before we return the result of `sqlite3_sql()`. I should have done this earlier, but I forgot. This PR is a good opportunity.

    This will make `Statement.sql` and `Statement.description` crash if those properties are not called in the proper dispatch queue. The raw SQLite connections and statement pointers are indeed *not* thread-safe.
     
    This will have no consequence. Statements are already very keen on crashing due to dispatch queue mismatch, but this bothers nobody because nobody uses statements outside of `dbQueue.write/read { db in ... )` blocks. Using statements somewhere else is a programmer error: 💥

- Let's update the [How do I print a request as SQL?](https://github.com/groue/GRDB.swift/blob/master/README.md#how-do-i-print-a-request-as-sql) FAQ, so that we just tell users to `print(statement)`.

groue:
Hello @steipete, I have a few commits that complete your pull request. Will you please grant me push access to your branch?

groue:
Merged! Thanks @steipete!

PR #968, opened 2021-04-22
Add conditional modifier to build requests

steipete:
Consider following statement. These two new conditionals make request building FAR more convenient. Everything can be done without the need for inline variables. This follows the trend to structure ViewModifiers in SwiftUI.

```swift
public static func get(userId: Int64, tweetUserId: Int64? = nil, limit: Int = 100, db: Database) throws -> [TweetInfo] {
try TweetUserFavorite
    .filter(TweetUserFavorite.Columns.userId == userId)
    .including(required: TweetUserFavorite.tweet
                .if(tweetUserId != nil) { request in
                    request.filter(Tweet.Columns.userId == tweetUserId!)
                }
                .order(Tweet.Columns.createdAt).reversed())
    .including(required: TweetUserFavorite.user)
    .asRequest(of: TweetInfo.self)
    .limit(limit)
    .fetchAll(db)
}
```

Even more convenient, we can use `modifier` to automatically resolve the optional:

```swift
public static func get(userId: Int64, tweetUserId: Int64? = nil, limit: Int = 100, db: Database) throws -> [TweetInfo] {
try TweetUserFavorite
    .filter(TweetUserFavorite.Columns.userId == userId)
    .including(required: TweetUserFavorite.tweet
                    .modifier(let: tweetUserId) { request, tweetUserId in
                        request.filter(Tweet.Columns.userId == tweetUserId)
                    }
                .order(Tweet.Columns.createdAt).reversed())
    .including(required: TweetUserFavorite.user)
    .asRequest(of: TweetInfo.self)
    .limit(limit)
    .fetchAll(db)
}
```

Both are valid and useful, even though this particular example only needs let. Another common use for this in my codebase is sorting and optional limit-settings.

TODO: I feel that this should be mentioned in the Documentation, I hope @groue can take this over, as I'm not familiar enough yet to know where to best put it.

### DIFF
--- a/GRDB/QueryInterface/Request/RequestProtocols.swift
+++ b/GRDB/QueryInterface/Request/RequestProtocols.swift
@@ -777,3 +777,31 @@ extension DerivableRequest {
         return request.having(expression)
     }
 }
+
+extension DerivableRequest {
+    /// Conditional request modifier to conveniently build a request by chaining methods.
+    /// - Parameters:
+    ///   - condition: The condition that needs to return true for transform to be called
+    ///   - transform: Code that modifies a request
+    /// - Returns: Modified request if condition returns true, or unmodified request.
+    public func `if`(_ condition: @autoclosure () -> Bool, transform: (Self) -> Self) -> Self {
+        if condition() {
+            return transform(self)
+        } else {
+            return self
+        }
+    }
+
+    /// Conditional request modifier to conveniently build a request by chaining methods with an optional.
+    /// - Parameters:
+    ///   - item: Optional to resolve
+    ///   - content: Code that modifies a request
+    /// - Returns: Modified request if optional is not nil, or unmodified request.
+    public func modifier<Item>(`let` item: Item?, then content: (Self, Item) -> Self) -> Self {
+        if let item = item {
+            return content(self, item)
+        } else {
+            return self
+        }
+    }
+}


groue:
Hello (again!) @steipete :-)

The `if` modifier is much welcomed, as we agreed together when we last discussed about it! As I told you then, I find it useful, I find it quite legible, and I've met it in a few places already: we are not introducing a unique and weird api:

```swift
// This is becoming kind of state of the art:
let target = value
    .foo()
    .if(condition) { value in value.bar() }
    .qux()

// This replaces:
var target = value.foo()
if condition {
    target = target.bar()
}
target = target.qux()
```

People do not have to use it, people who are familiar with the pattern will enjoy it, and people who are not familiar with it will guess what it does. 👍 

---

On the other side, I find it hard to swallow `modifier(let:)`:

```swift
let item: Item? = ...
let target = value
    .foo()
    .modifier(let: item) { (value, item) in value.bar(item) }
    .qux()

// This replaces:
let item: Item? = ...
var target = value.foo()
if let item = item {
    target = target.bar(item)
}
target = target.qux()
```

I know where `modifier` [comes from](https://developer.apple.com/documentation/swiftui/viewmodifier). SwiftUI does not use this method name in order to wrap `if let`, though, and I think we're abusing and muddying both the "modifier" concept and word, here.

For example, did you consider `if(let:) { ... }` instead, which is more easy to guess for people who are not familiar with it?

I know new apis can create knee-jerk reactions, until they just "click" and become natural and self-obvious. Convince me!

--- 

> TODO: I feel that this should be mentioned in the Documentation, I hope @groue can take this over, as I'm not familiar enough yet to know where to best put it.

Yes. I'll totally have to think seriously about putting some light on `DerivableRequest`.

groue:
OK, I close this pull request. The attached code can live at application level.

Issue #972, opened 2021-05-01
Not building 


PratikThakareIos:
Please check
<img width="755" alt="Screenshot 2021-05-01 at 8 39 41 PM" src="https://user-images.githubusercontent.com/25743876/116782713-5c5be800-aabd-11eb-95e9-81b3f4bc6660.png">

### What did you do?

Trying to install new rxGRDB or grdb  using pod but project is not building

### What did you expect to happen?

project should building

### What happened instead?

throwing error

### Environment
iOS 
**GRDB flavor(s):** (GRDB, SQLCipher, Custom SQLite build?)

**GRDB version:**
**Installation method:** (CocoaPods, SPM, manual?)
**Xcode version:**
**Swift version:**
**Platform(s) running GRDB:** (iOS, macOS, watchOS?)
**macOS version running Xcode:**

groue:
@PratikThakareIos, GRDB builds well in all supported configuration. Do not expect any support if you do not provide any information that can help somebody reproduce your setup. Also, **when you open an issue in a repository that provides an issue template, fill this template**. This really is basic etiquette.

nominalista:
@groue let me add more context to this issue, as I experienced the same troubles as @PratikThakareIos.

### What did you do?

I try to use GRDB in a xcframework along with CocoaPods. I've received the same error as in original question while trying to build the project with GDRB dependency in the podfile. It occurs that the build fails only with `BUILD_LIBRARY_FOR_DISTRIBUTION` set as `true`. 

I assume that GRDB may not supporting this setting and there are several reason as in case of [gRPC](https://github.com/grpc/grpc-swift/issues/1063), but it would be great if you someone could verify this issue.

### What did you expect to happen?

To be able to build the project with `BUILD_LIBRARY_FOR_DISTRIBUTION` flag or at least find out some explanation for this problem..

### What happened instead?

Received `'let' property 'storage' may not be initialized directly, [...]` compilation error.

### Environment

**GRDB flavor(s):** GRDB.
**GRDB version:** 5.7.4
**Installation method:** CocoaPods
**Xcode version:** 12.5
**Swift version:** 5
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 11.2.3

### Demo Project

Reproduction steps:
- Create Framework project.
- Instantiate podfile with `GRDB` dependency.
- Set `BUILD_LIBRARY_FOR_DISTRIBUTION` to `true`.
- Try to build the project.

groue:
[...]

PR #979, opened 2021-05-25
Support for archiving GRDB for xcframeworks

### LINKED ISSUE #964: Support for Module Stability?
We want to use GRDB without needing to re-compile it every time. That is, we want to create a binary framework once and use it. This has proven quite challenging, since once a library is compiled using one compiler version, it can't be used by applications compiled using other compiler versions.
That is, unless module stability is enabled. I tried to compile GRDB with module stability (And library evolution, as module stability depends on it), by adding the flag `BUILD_LIBRARY_FOR_DISTRIBUTION` to `xcodebuild`, but it doesn't compile without changing the code. And I wouldn't want to diverge from the mainline. 

Is this something that you considered supporting? It would save us a lot of build time.

asifmohd:
This PR should help resolve #964 and #972.

Opening this PR as a draft PR to get feedback from library maintainers/authors.

I am not entirely sure, why removing `@inlinable` leads to a green build while building with `SKIP_INSTALL=NO` and `BUILD_LIBRARY_FOR_DISTRIBUTION=YES`

### Pull Request Checklist

- [ ] Changes are tested.
   - Facing failures on these tests on my local machine (only when BUILD_LIBRARY_FOR_DISTRIBUTION=YES is enabled in build settings for GRDBiOS) (Will dig deeper, but would appreciate any pointers)
image

### Testing Environment
Xcode  12.5, iOS 14.5 simulator, macOS 11.4

### DIFF
--- a/GRDB/Core/DatabaseValue.swift
+++ b/GRDB/Core/DatabaseValue.swift
@@ -85,8 +85,6 @@ public struct DatabaseValue: Hashable, CustomStringConvertible, DatabaseValueCon
     }
     
     // MARK: - Not Public
-    
-    @inlinable
     init(storage: Storage) {
         self.storage = storage
     }
--- a/GRDB/Core/Support/Foundation/Data.swift
+++ b/GRDB/Core/Support/Foundation/Data.swift
@@ -13,7 +13,6 @@ extension Data: DatabaseValueConvertible, StatementColumnConvertible {
     }
     
     /// Returns a value that can be stored in the database.
-    @inlinable
     public var databaseValue: DatabaseValue {
         DatabaseValue(storage: .blob(self))
     }
--- a/GRDB/Core/Support/StandardLibrary/StandardLibrary.swift
+++ b/GRDB/Core/Support/StandardLibrary/StandardLibrary.swift
@@ -235,7 +235,6 @@ extension Int64: DatabaseValueConvertible, StatementColumnConvertible {
     }
     
     /// Returns a value that can be stored in the database.
-    @inlinable
     public var databaseValue: DatabaseValue {
         DatabaseValue(storage: .int64(self))
     }
@@ -434,7 +433,6 @@ extension Double: DatabaseValueConvertible, StatementColumnConvertible {
     }
     
     /// Returns a value that can be stored in the database.
-    @inlinable
     public var databaseValue: DatabaseValue {
         DatabaseValue(storage: .double(self))
     }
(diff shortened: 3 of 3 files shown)


groue:
Hello @asifmohd, thanks for this pull request!

Would you please tell how I can check that no regression is added in the future? A command that fails if xcframeworks support is broken? Without this check, it is difficult to make sure you won't have a bad surprise.

groue:
Oops, sorry, I did not read well the changes. So you *removed* some `@inlinable`...

> I am not entirely sure, why removing @inlinable leads to a green build while building with SKIP_INSTALL=NO and BUILD_LIBRARY_FOR_DISTRIBUTION=YES

Maybe you should make an opinion.

Also, https://github.com/groue/GRDB.swift/issues/964 makes me question your need in the first place. I think you have to elaborate.

asifmohd:
Hi @groue, apologies for the delayed reply.

I have raised a [swift compiler bug](https://bugs.swift.org/browse/SR-14719), since this issue is reproducibble on a simple swift file as well.
Waiting to hear back from Apple before proceeding further and exploring why the unit tests are broken

asifmohd:
> Also, #964 makes me question your need in the first place. I think you have to elaborate.

I was trying to see if we can create an `xcframework` for GRDB in our closed source project at my workplace and use this xcframework to save a few seconds while clean building the project everytime. 
Adopting Carthage would have been a much larger undertaking and it would also complicate our codebase a bit, since we are already using cocoapods. We've tried exploring a few ways to generate binaries from source code using cocoapods as well, but we haven't found the perfect solution yet.

Xcframeworks would have been a quick way to prevent rebuilding modules from source code for a few libraries, without drastically changing our developer workflow.


> Would you please tell how I can check that no regression is added in the future? 

For this we'll mostly have to add another job on CI, to compile GRDB.swift with the `-enable-library-evolution` flag or `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` using xcodebuild. And then run the entire test suite on top of this module stable library.

As long as the module compiles, we should have module stability for that specific version. And devs who create xcframeworks will not be blocked or face any compiler issues. Though this could again break with new swift compiler releases maybe 🤔 . But the CI should atleast help us catch this earlier.

groue:
Sorry @asifmohd, I close this pull request. Your changes remove compiler errors, but consequences on applications runtime are unknown.

groue:
(Please see https://github.com/groue/GRDB.swift/issues/1027#issuecomment-896525400 for a more detailed rationale about the reasons this PR was not merged)

Issue #988, opened 2021-06-15
table.synchronize() cancels the transaction when triggers already exist


pp5x:
Hi @groue

### What did you do?

In a application first launch case (out of migrations cases), I'm initialising the DB schema with:

```.swift
// Initializing table record; if not exists
try db.create(table: "record", ifNotExists: true) { t in
    // ...
}

// Initializing a virtual table to index content from the record table.
try db.create(virtualTable: "content", ifNotExists: true, using: FTS4()) { t in
    t.synchronize(withTable: "record")
    t.column("title")
}
```

### What did you expect to happen?

I want to make sure the DB schema is up to date; without using transactions since if there is breaking change I can still trash the DB files.

### What happened instead?

There are the following cases:

1. The DB file does not exists yet: it creates the tables and the triggers. ✅ 
2. The DB was cleared with drop tables and delete triggers, the DB file may remain but there is no issue. ✅ 
3. The code is called again but we kept the data (no clear), we make sure the schema is up to date, but the triggers created from `t.synchronize` already exists and it makes the whole transaction to fail. 🔴 

Should the `t.synchronize` have a `ifNotExists` parameter? Or even better, it would follow the `ifNotExists` from the table creation.

Or am I initialising the DB the wrong way? How can I figure out the DB state to figure out if a call to `t.synchronize` must be made?

### Current workaround

What I found so far, but it is relying on GRDB's implementation details:

```.swift
let mustCreateTriggers = try !db.triggerExists("__content_bu")
try db.create(virtualTable: "content", ifNotExists: true, using: FTS4()) { t in
      if mustCreateTriggers {
           t.synchronize(withTable: "record")
      }
```

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** v5.8.0
**Installation method:** SPM
**Xcode version:** 12.5
**Swift version:** 5
**Platform(s) running GRDB:** macOS
**macOS version running Xcode:** 11.4

groue:
Hello @pp5x,

Oops: the `ifNotExist` parameter is not passed to [FTS4.database(_:didCreate:using:)](https://github.com/groue/GRDB.swift/blob/v5.8.0/GRDB/FTS/FTS4.swift#L94-L135), which creates the synchronization triggers.

It happens that I use migrations 100% of the time, and this might explain why I did not spot this issue.

> Should the t.synchronize have a `ifNotExists` parameter? Or even better, it would follow the `ifNotExists` from the table creation.

The second option looks reasonable.

[...]

PR #989, opened 2021-06-15
FTS: add support of ifNotExists in synchronize()

pp5x:
As mentioned in #988 , `synchronize()` does not take the `ifNotExists` flag from the table. Which may result in trigger creation errors, but they could already exists.

Attempt is made to extend the current protocol by adding another that add support of VirtualTableModule metadata. Which allows classes that implements it to read them when they are provided.

*Draft PR*: I am not fluent in Swift enough to figure out how to make the protocol `VirtualTableModuleWithMetadata` to dispatch and call the appropriate method the `makeTableDefinition`.

### Pull Request Checklist

- [x] Changes are tested.

### DIFF
--- a/GRDB/QueryInterface/Schema/VirtualTableModule.swift
+++ b/GRDB/QueryInterface/Schema/VirtualTableModule.swift
@@ -21,13 +21,25 @@ public protocol VirtualTableModule {
     /// The name of the module.
     var moduleName: String { get }
     
+    // TODO: remove this requirement
     /// Returns a table definition that is passed as the closure argument in the
     /// `Database.create(virtualTable:using:)` method:
     ///
     ///     try db.create(virtualTable: "item", using: module) { t in
     ///         // t is the result of makeTableDefinition()
     ///     }
+    ///
+    /// - warning: This method is DEPRECATED. Define
+    ///   `makeTableDefinition(configuration:)` instead.
     func makeTableDefinition() -> TableDefinition
+    
+    /// Returns a table definition that is passed as the closure argument in the
+    /// `Database.create(virtualTable:using:)` method:
+    ///
+    ///     try db.create(virtualTable: "item", using: module) { t in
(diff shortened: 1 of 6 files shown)


GRDB/QueryInterface/Schema/VirtualTableModule.swift
```diff
@@ -66,7 +75,17 @@ extension Database {
         let sql = chunks.joined(separator: " ")
         try execute(sql: sql)
     }
-    
+
+    private func makeTableDefinition<Module: VirtualTableModule>(_ module: Module, _ metadata: ModuleMetadata) -> Module.TableDefinition
+    {
+        module.makeTableDefinition()
+    }
```
pp5x, line 82: Problem: this method is always called. Not sure how to have the other one used instead.

groue: You're right.

I have another idea.

groue (commented):
@pp5x, is it OK if I force push on your branch in order to replace your "dirty" commit https://github.com/groue/GRDB.swift/pull/989/commits/75cd1b915daf1dfc4f49e7545a3922d422a51215?

I think I have a solution which is not breaking, does not introduce any extra protocol, and does not have the issue you mentioned.

I kept your good idea of a dedicated "metaData" structure 👍 although I renamed it "configuration".

groue:
You can see the commits I'd like to push to your branch there: https://github.com/groue/GRDB.swift/tree/dev/issue-988

(I do intend to keep your attribution on the pull request.)

pp5x:
Hi @groue 

Feel free to push whatever you need :) I guess there was not much alternative that to make a breaking change with  `makeTableDefinition`.

Thank you!

groue:
All right, @pp5x, done! If this is OK to you, I will merge this pull request 🙂

groue:
All right, let's just ship this 👍

groue:
Shipped in v5.9.0!

Issue #990, opened 2021-06-16
Support > Is it possible to use QueryInterfaceRequest builder with dynamic table/view name?


Kirow:
### What did you do?

let's say I have following data model:

```swift
struct MyRecord: FetchableRecord {
    let title: String
    let details: String
    
    enum Columns: String, CodingKey, ColumnExpression {
        case title, details
    }
    
    init(row: Row) {
        title = row[Columns.title]
        details = row[Columns.details]
    }
}
```

And I'd like to use `QueryInterfaceRequest` to build queries and load records. But in my case I don't have table name at compile time and single model should be used across multiple tables. So it seems I can not adopt `TableRecord` and as result have no access to `QueryInterfaceRequest` builder.

for example lets say I'm loading table definition from some external source:
```swift
struct Definition {
    let name: String
    let keys: [String]
}

let myDefinitions = [
    Definition(name: "table1", keys: ["key1", "key2"]),
    Definition(name: "asdasd", keys: ["whatever", "whocares"])
]
```

### What did you expect?

I'm looking for some way to create `QueryInterfaceRequest<MyRecord>` using provided `Definition` and existing `MyRecord: FetchableRecord` 

for example:

```swift
// technically it is working for me if I change access level inside GRDB
let requests: [QueryInterfaceRequest<MyRecord>] = myDefinitions.map({ def in
    let rel = SQLRelation.all(fromTable: def.name, selection: { _ in [
        Column(def.keys[0]).forKey(MyRecord.Columns.title),
        Column(def.keys[1]).forKey(MyRecord.Columns.details),
    ] })
    
    return QueryInterfaceRequest<MyRecord>(relation: rel)
})
```

### Current workaround

If it is not possible to use `QueryInterfaceRequest` in this case, then I will be using custom wrapper with raw SQL.

```swift
struct MyQueryWrapper<T: FetchableRecord> {
    let sql: (_ limit: Int, _ offset: Int) -> String
    
    func fetch(_ db: Database, limit: Int, offset: Int) throws -> [T] {
        return try T.fetchAll(db, sql: sql(limit, offset))
    }
}

let wrappers = myDefinitions.map({ def in
    return MyQueryWrapper<MyRecord> { limit, offset in
        """
            SELECT
                \(def.keys[0]) AS \(MyRecord.Columns.title.name),
                \(def.keys[1]) AS \(MyRecord.Columns.details.name)
            FROM \(def.name)
            LIMIT \(limit) OFFSET \(offset)
[...]

Issue #916, opened 2021-02-02
I need a table name that can be adjusted according to the actual situation, instead of a fixed table name. How can I do that?


forkdog:
`static var databaseTableName: String {
        return "name"
}`

groue:
Hello @forkdog,

This question was already asked in #833.

If your database schema contains a dynamic number of tables, you must explain why you need to escape the standard relational model.

If your database schema contains a fixed set of tables, but you need to access several tables through a single record type, then this is a distinct question, and we may discuss it.

Considering the terse way you asked the question, and the fact that you did not provide any explanation for your request, I don't expect any answer: I close this issue. Feel free to reopen if you have something specific to ask.

groue:
QueryInterfaceRequest builder with dynamic table/view name is addressed by the pull request #993.

Issue #833, opened 2020-09-09
Decoupling record type from table name


sepw:
I would like to use the query interface to build queries for my record type. However there isn't a 1:1 relationship between record type and underlying source/table:

- A record may come from one of 3 tables: `A`, `B` or `B`.
- The schemas of tables `A`, `B` and `C` are the same, i.e. they have the same number of columns and the columns have the same types.
- Table `A` is in the main database. Tables `B` and `C` may be in the temporary database or attached databases, either on disk or in memory.
- Records may be inserted into table `A`, `B` or `C` but fetched from view `V`, which is a UNION of `A`, `B` and `C`.

GRDB seems to assume that a record type comes from a single table, i.e. the table returned from `TableRecord. databaseTableName`. I need a way to decouple the table name from the type.

Is there a way to specify the table/view name when using the query interface? I was hoping for something like:

``` swift
_queue.read { db in
  return MyRecord.from("temp.B").fetchAll(db)
}
```

sepw:
After some investigation into `TableRecord` and `QueryInterfaceRequest` it seems it might be straightforward to implement a `from(_ table: String)` function or similar as an extension on `TableRecord`, e.g.

``` swift
public static func from(_ table: String) -> QueryInterfaceRequest<Self> {
    let relation = SQLRelation(
        source: .table(tableName: table, alias: nil),
        selectionPromise: DatabasePromise(value: databaseSelection))
    return QueryInterfaceRequest(relation: relation)
}
```

Unfortunately `SQLRelation` has `internal` access control...

groue:
Hello @sepw,

> GRDB seems to assume that a record type comes from a single table, i.e. the table returned from `TableRecord. databaseTableName`.

Exact.

> I need a way to decouple the table name from the type.

I have been waiting for this feature request :-)

> After some investigation [...] it seems it might be straightforward to implement.

Indeed. Enough years have passed now so I do feel rather confident that this should not create much problem. I'm all with you.

> `MyRecord.from("temp.B").fetchAll(db)`

This reads well, but `MyRecord.from("name").fetchAll(db)` does not. What is "name"?

`MyRecord.from(table: "name").fetchAll(db)` reads much better, but is not quite right with views.

Also, do we want to create such requests from `TableRecord`, which has its `databaseTableName` property? In your particular case, it looks like your record type does not have *one* natural source table, but several. I wouldn't like to force users to "bless" one of the tables. I would even like to force users to pick one of those table, explicitly, for each request.

[...]

PR #993, opened 2021-06-20
QueryInterfaceRequest builder with dynamic table/view name

groue:
This pull request addresses #990, #916, and #833.

A new `Table` type is introduced, which can be used just as a `TableRecord` type. You can fetch, count, build requests, associations, etc.

```swift
// "Anonymous" Row table
let table = Table("player") // Or Table<Row>("player")
let count = try table.fetchCount(db)
let rows: [Row] = try table.fetchAll(db)
let rows: [Row] = try table.filter(...).fetchAll(db)
try table.deleteAll(db)
try table.deleteOne(db, key: 1)
try table.updateAll(db, ...)

// Typed table
struct Player: FetchableRecord { ... }
let table = Table<Player>("player")
let count = try table.fetchCount(db)
let players: [Player] = try table.fetchAll(db)
let players: [Player] = try table.filter(...).fetchAll(db)
try table.deleteAll(db)
try table.deleteOne(db, id: 1) // when Player is Identifiable
try table.updateAll(db, ...)
```

@Kirow, @forkdog, @sepw, @wzio, may I ask if this looks like it would address your initial feature request? See the [documentation diff](https://github.com/groue/GRDB.swift/pull/993/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).


sepw:
Hi @groue,

The addition of `Table("name")` looks like it would solve some of my requirements. In my current project I have several full text tables, all with the same schema: an ID and a text column. I currently use a generic table row type that allows the table name to be specified at compile time using generics. It looks like I will be able to replace this with your new `Table` type.

I took a brief look at the source for `Table` that adds an extension where `RowDecoder == Row`

```swift
extension Table where RowDecoder == Row {
     /// Create a `Table` of `Row`.
     ///
     ///     let table = Table("player") // Table<Row>
     public init(_ tableName: String) {
         self.init(tableName, Row.self)
     }
 }
```

This results in two `init(_ tableName: String)` initializers. Won't this cause call-site ambiguities?

groue:
Thanks for your feedback, @sepw!

> Won't this cause call-site ambiguities?

None that I'm aware of. All snippets below compile fine:

```swift
struct Player: FetchableRecord { ... }

try dbQueue.read { db in
    do { let rows = try Table("player").fetchAll(db) }
    do { let rows: [Row] = try Table("player").fetchAll(db) }
    do { let rows = try Table("player").all().fetchAll(db) }
    do { let players = try Table<Player>("player").fetchAll(db) }
    do { let players = try Table<Player>("player").all().fetchAll(db) }
    do { let players: [Player] = try Table("player").fetchAll(db) }
    do { let players: [Player] = try Table("player").all().fetchAll(db) }
}
```

My goal with this default `Row` generic parameter is to make it possible to just declare a `Table` without qualifier when it would be inconvenient to choose a generic type. For example:

```swift
// We don't care about the actual generic parameter
let count = try Table("player").fetchCount(db)

// This avoids this kind of useless choice:
let count = try Table<Void>("player").fetchCount(db)   // ?
let count = try Table<Row>("player").fetchCount(db)    // ?
let count = try Table<Player>("player").fetchCount(db) // ?
let count = try Table<Never>("player").fetchCount(db)  // ?
let count = try Table<WHAT>("player").fetchCount(db)   // ?
```

The same pattern is already used for [Common Table Expressions](https://github.com/groue/GRDB.swift/blob/master/Documentation/CommonTableExpressions.md), and `SQLRequest`, without known downside:

```swift
// SELECT * FROM player WHERE score = (SELECT MAX(score) FROM player)
let subquery = SQLRequest("SELECT MAX(score) FROM player") // Implicit SQLRequest<Row>
let request = SQLRequest<Player>("SELECT * FROM player WHERE score = (\(subquery))")
let players = try request.fetchAll(db) // [Player]
```

I know it's an uncommon API pattern - and a rather hidden one. Not everybody will use it, or will be conscious of its existence. But I've too frequently seen myself inject arbitrary types just to make the compiler happy. Defaulting to `Row` seemed like a better option, it does not look like it hurts anybody, and I decided the experiment was worth it.

groue:
Note to self: add deletion methods (`table.deleteAll(_:)`, `.delete(_:id:)`, etc.), and batch updates as well.

sepw:
@groue, thank you for the explanation of the initializer with the `RowDecoder == Row` predicate. I wasn’t aware that this technique could be used to simulated default generic arguments. Very useful!

groue:
Yep! The pattern is:

```swift
struct Foo<T> {
    // Support
    private init(_ type: T.Type) { }
    
    // Initializer for explicit generic parameter
    init() {
        self.init(T.self)
    }
}

extension Foo where T == Default {
    // Initializer for default generic parameter
    init() {
        self.init(Default.self)
    }
}

// Explicit generic parameter
let foo = Foo<Int>() 

// Default generic parameter
let foo = Foo()
```

groue:
😠 Travis is dead

NewestUser:
Is it possible to insert data using the `Table` struct? Or even better perform a `save` like in `MutablePersistableRecord` without actually caring of the mutation result?

PS: Awesome lib, thanks!

groue:
Hello @NewestUser, `Table` does not provide any saving features. If you want this topic to be explored, please explain your use case (don't say _how_ you want to do something, instead focus on _what_ you want to do).

NewestUser:
Thanks for the reply.

I have an iOS app that supports multiple languages. Each language has it's own set of tables containing data specific for each language.
Example. `table1_bg`, `table2_bg`, `table1_en`, `table2_en`, `table1_de`, `table2_de` etc...
`table1_bg`, `table1_en` and `table1_de` all have the same schema. They differ only by the data stored in those tables and the name suffix of the table.

Depending on the language that the application is running in the appropriate table suffix will be used.

A natural response to this design decision that I have made might be: _You can add a new language column in each of the tables and add an extra where clause to your queries so that only data for the particular language is selected._

This is true, however the effort that I needed to spent in order to do this was far more than just adjusting table suffixes.
Keeping data in separate tables with the same schemas allowed me to:
- Easily add new languages without having to adapt existing repository layer. Otherwise I would have to do a DB migration that would involve creating the new tables with the extra language column, copy data from old tables into the new ones and then clearing out the old tables. Keep in mind that these kind of migrations are done on mobile devices that are outside of your control. Sometimes one needs to keep these migration in mind for months when developing the app. The reason being is that it takes months for users to upgrade to the latest version of the app. Especially when you have a large user base. So introducing backwards incompatible changes on a db level can happen. This always leads to increased uninstall rates and user loss.
- Keep the size of tables and indices low since each language has ~10k of records.

PS: Merry Christmas 🎄

groue:
> [...] however the effort that I needed to spent in order to do this was far more compared to just adjusting table suffixes.

Is it?

NewestUser:
If you account for the db migration that would last months I would say yet it was.
Either way I am not defending the design decision that I made. If I was to implement a server side db model I would never go this route. However the current state is that I have different tables with the same schema for different languages. In addition this db schema is shared by an android app and as a single developer supporting both platforms it is easier for me to keep thinks as close as possible.

For what reason `Table` does not provide saving features? Would it require restructuring a huge chunk of the code base or something else?

groue:
> PS: Merry Christmas 🎄

You too :-) Happy New Year's Eve!

> If you account for the db migration that would last months I would say yes it is.

We still miss the rationale that supports such a claim. Don't worry, I guess it is about your Android app, and I can understand that.

> For what reason `Table` does not provide saving features?

No good reason. The popular request was the ability to build SELECT queries. I agree that the current state of the `Table` struct, able to build requests, but unable to perform inserts, is problematic. It is not cool to take a road, drive for a while, and discover later that it is a dead-end.

> Would it require restructuring a huge chunk of the code base or something else?

I don't think so. We'd just need to design an API and implement it. I would welcome such a contribution, since it would fill a hole in the api.

Now my recommendation is still the same: **stick to the standard relational model**. Leaving it is begging for trouble. You just met one (because database libraries do not spend much energy helping unorthodox practices) - you are likely to meet other ones (because opting out of the relational model is exposing yourself to the troubles it is designed to avoid).

Issue #1007, opened 2021-07-10
How to auto binding struct variable and table column name


Chen-Michael:
### What did you expect to happen?

My DBA used "snake case" in create the data table (e.g., group_id),

but we are used "camel case" in design the struct (e.g., groupId),

is there a scheme that can be auto binding two way?

I know can be use "encode(to:)" to binding column,

but there will be a lot of duplicate code in each model,

could you help me solution this problem?

groue:
Hello @Chen-Michael 

it looks like you are asking for something like https://developer.apple.com/documentation/foundation/jsondecoder/keydecodingstrategy/convertfromsnakecase and its encoding counterpart.

Those are not built-in GRDB. I would be interested in a pull request.

Chen-Michael:
> Hello @Chen-Michael
> 
> it looks like you are asking for something like https://developer.apple.com/documentation/foundation/jsondecoder/keydecodingstrategy and its encoding counterpart.
> 
> Those are not built-in GRDB. I would be interested in a pull request.

https://github.com/groue/GRDB.swift/pull/1008

Please let me know if there are any parts that have not been taken into consideration

groue:
This issue describes a scenario where the library could help the user more, I agree. #1008 was not accepted, because it does not address the issue at the correct level. The correct solution is letting Codable record types customize how their coding keys are turned into column names.

Chen-Michael:
> This issue describes a scenario where the library could help the user more, I agree. #1008 was not accepted, because it does not address the issue at the correct level. The correct solution is letting Codable record types customize how their coding keys are turned into column names.

I was inexpert in Swift, can't understand the correct solution you said

Sorry for wasting your time, hope someone can help solve this problem

groue:
No worries, @Chen-Michael. You surely did not waste the time of anyone! Your idea is very good, and I too hope it eventually ships in the library. I'm right now focusing on shipping the improvements that are already developed, so I won't personally tackle your problem right now. Maybe someone else will, or I will do it later.

The current policy is to close issues that nobody is currently working on. So I will close this issue. I'll however add the "needs revisiting" flag, as a reminder that there's a good idea that is not yet addressed here.

PR #1008, opened 2021-07-11
fix issue 1007

Chen-Michael:
Auto binding struct column(camel case) to table column(snake case)

### Pull Request Checklist

- [ ] Inline documentation has been updated. <- Only update the core code without involving public methods
- [ ] README.md or another dedicated guide has been updated. <- Only update the core code without involving public methods
- [x] Changes are tested.

### DIFF
--- a/GRDB.xcodeproj/project.pbxproj
+++ b/GRDB.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		27D12D66269A905A00D548CF /* AutoBindingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 27D12D65269A905A00D548CF /* AutoBindingTests.swift */; };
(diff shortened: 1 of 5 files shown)


groue:
Thank you @Chen-Michael. I'm off for a week, and I'll look at your pull request when I'm back.

groue:
Hello @Chen-Michael. I'm sorry, but this pull request has nothing in common with what I suggested here: https://github.com/groue/GRDB.swift/issues/1007#issuecomment-877708961

In your version, one can use access a column with both "fooBar" and "foo_bar" keys, indifferently, in some happy scenarios. This creates problems:

- The `Row` type is no longer a proper collection with clearly defined columns. You made it easy for your personal situation, but you made it quite _bizarre_ for everyone else.
- If some GRDB users have tables that have two columns named, for example, `foo_bar` and `fooBar`, then your change would break their application.

I can not accept this pull request.

PR #1010, opened 2021-07-12
Add VACUUM INTO Support

tternes:
First - thank you so much for this incredible library. It has been a joy to use in my latest project, and I'm grateful for thoughtful effort that has gone into this project.

I'm experimenting with using `VACUUM INTO` as an alternative to backup, since it has the file-trimming behavior of `VACUUM` while producing a separate database file.

This PR introduces native support for `VACUUM INTO` to `DatabaseWriter`. My first inclination was to add this to `DatabaseReader` since it doesn't modify the current database, but SQLite requires that `VACUUM` be run outside of a transaction (which I don't see a way of doing currently in `DatabaseReader`). If I've overlooked something, I'd be glad to refactor as necessary.

Thanks for considering this addition. 

### Pull Request Checklist

- [X] This pull request is submitted against the `development` branch.
- [X] Inline documentation has been updated.
- [ ] README.md or another dedicated guide has been updated. *(Not applicable)*
- [X] Changes are tested.


groue:
Hello @tternes,

Thank you very much for this addition 👍 

`VACUUM INTO` was introduced in [SQLite 3.27.0](https://www.sqlite.org/releaselog/3_27_0.html).

- We need availability checks for the regular GRDB that links against the system SQLite.
- Those checks are not required for SQLite custom builds (greater than 3.27 anyway)
- Those checks can not easily be expressed for [SQLCipher (v4.1.0+)](https://github.com/sqlcipher/sqlcipher/blob/master/CHANGELOG.md#410---march-2019---410-changes), so let's just not expose this method for SQLCipher (unless you have an idea).

This should give something like:

```swift
// VACUUM INTO was shippeed in SQLite 3.27.0:
// https://www.sqlite.org/releaselog/3_27_0.html
#if GRDBCUSTOMSQLITE
func vacuum(into...) { ... }
#elseif !GRDBCIPHER
@available(OSX 10.16, iOS 14, tvOS 14, watchOS 7, *)
func vacuum(into...) { ... }
#endif
```

Also, could you please enhance the name of the parameter, and the inline documentation, so that the reader understands that this method accepts a _file path_?

tternes:
Really great feedback - thank you @groue. Branch updated with improvements as suggested.

groue:
Due to CI concerns (the setup is completely broken), you could not spot testing issues. I've solved them.

Also, I... changed my mind regarding SQLCipher. `vacuum(into:)` is now enabled, and users of old versions of SQLCipher will have a runtime error. This shouldn't be such a big problem, since those people use a _single_ version of SQLCipher, unlike people who target the system SQLite: their own tests will easily catch the eventual problem.

groue:
I think we're good! This was a good pull request, @tternes, thank you very much :-)

I have also appreciated our joined effort. Would you like to be granted a write access to the repository?

groue:
I really hope to ship GRDB 5.9 as soon as possible. I may have to skip proper CI 😬

groue:
Shipped in v5.9.0 :-)

tternes:
Congrats on the release! 🎉

Issue #983, opened 2021-06-03
5.7.4 performance regression?


gverdouw:
### What did you do?

Updated from 5.7.3 to 5.7.4

### What did you expect to happen?

No noticeable impact on my application

### What happened instead?

My app become noticeably slower in multiple areas. Further debugging seems to show lots of threads sitting in CaseInsensitiveIdentifier.hash(into:). The main place I notice this is when going through a loop doing a fetch and then a create after.

Is there any obvious gotchas with this? Perhaps I am doing something weird with my records?

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 5.7.4 and 5.8.0
**Installation method:** SPM
**Xcode version:** 12.5
**Swift version:** 5
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 11.4

gverdouw:
One common thing I notice is the records that I notice this with the most are being fetched with a lot of optional associations. For example something like: 

```swift
MainRecord
            .including(optional: MainRecord.association1)
            .including(optional: MainRecord.association2)
            .including(optional: MainRecord.association3)
            .including(optional: MainRecord.association4)
            .including(optional: MainRecord.association5)
            .including(optional: MainRecord.association6)
            .including(optional: MainRecord.association7
                .including(all: AssociatedRecord7.association))
```

There are even more `including`s in my actual query I have only included a handful here... I'm guessing this kind of thing is causing the hash to be too much?

groue:
Hello @gverdouw,

Performance regressions are an important issue. I'd appreciate if you'd provide a detailed setup, please. There are *too many ways* to use GRDB, and I'm not able to guess yours.

gverdouw:
I'll see if I can put together a minimal reproduction project but it might take a while as my app is quite complicated and I am not really sure exactly what might be causing the issue.

While I'm putting that together is there any kind of info in particular that would be useful about my setup?

groue:
For small reproducing examples, I usually go with something like:

``` swift
// A tmp db
let dbQueue = try DatabaseQueue(path: someTempPath)
try dbQueue.write { db in
    // setup minimal schema
}

// Define minimal set of records
struct Player { ... }
struct Team { ... }
struct Game { ... }

// Populate db with minimal data
try dbQueue.write { db in
    // insert minimal data needed for reproducing the issue
}

// This request used to run much faster
let results = try dbQueue.read { db in
    let request = /* some request */
    try request.fetchAll(db)
}
[...]

PR #1037, opened 2021-08-21
Support BUILD_LIBRARY_FOR_DISTRIBUTION

### LINKED ISSUE #1027: Suggestion: provide a stable ABI for GRDB
### What did you do?

I upgraded Xcode.

I expected my projects that use my binary build of the GRDB framework to compile.

### What happened instead?

Xcode complains that GRDB was built with an older version of the Swift compiler and needs to be rebuilt.

This is a common annoyance with Swift-based binary frameworks. I think the solution basically comes down to updating GRDB so that it builds with the "Build Libraries for Distribution" Xcode build option. If only it were that easy...

When I tried building GRDB with that option, I got stumped by a compiler error—probably the first of many. The fact is, writing code that allows for a stable ABI is not entirely trivial. This mode introduces some additional restrictions in Swift coding semantics. For example, ABI-stable Swift structs are more restrictive in their implementation.

In fact, I wouldn't even say that the idea of any or every Swift framework providing a stable ABI should be considered desirable for its own sake. For example, some Swift frameworks such as the gRPC integration have no intention of providing ABI stability because (I presume) it's based on code generation. However, I do think ABI stability for GRDB is worthwhile and desirable because

 * GRDB provides API stability via the semantic versioning paradigm
 * adding GRDB as a Swift source package implies compiling it every time you clean your project's build directory, and building GRDB is slow
 * other things I'm not thinking of right now

What do you think? Is this a worthwhile goal?

Thank you as always.

### Environment

**GRDB flavor(s):** GRDB with custom SQLite build
**GRDB version:** 5.8.0
**Installation method:** manual framework built with Xcode
**Xcode version:** 13.0 beta 4
**Swift version:** 5.5
**Platform(s) running GRDB:** macOS
**macOS version running Xcode:** 11.5

groue:
This pull requests has GRDB build with the `BUILD_LIBRARY_FOR_DISTRIBUTION` flag.

**It does not in any way promises ABI stability.** GRDB ABI may change between two minor releases.

Fix #1027, #979, #964

cc @asifmohd, @byohay, @mallman,  @nominalista, @PratikThakareIos: would you please check if this change fits your needs? Please report your troubles building GRDB, but also new warnings or errors in your dependent projects.

Considering your feedback might take time, because such checks are not exactly a trivial task, and August is a quiet month, this PR will be merged anyway. We'll fix remaining issues later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 
 #### 5.x Releases
 
+- [Development Branch](#development-branch)
 - `5.9.x` Releases - [5.9.0](#590)
 - `5.8.x` Releases - [5.8.0](#580)
 - `5.7.x` Releases - [5.7.0](#570) | [5.7.1](#571) | [5.7.2](#572) | [5.7.3](#573) | [5.7.4](#574)
@@ -75,6 +76,12 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 - [0.110.0](#01100), ...
 
 ---
+
+## Development Branch
+
+- **New**: [#1036](https://github.com/groue/GRDB.swift/pull/1036) by [@groue](https://github.com/groue): Restore minimal CI tests
+- **New**: [#1037](https://github.com/groue/GRDB.swift/pull/1037) by [@groue](https://github.com/groue): Support BUILD_LIBRARY_FOR_DISTRIBUTION
+
 
 ## 5.9.0
 
--- a/GRDB/Core/Cursor.swift
+++ b/GRDB/Core/Cursor.swift
@@ -19,7 +19,6 @@ extension RangeReplaceableCollection {
     ///     let strings = try Array(cursor)
     ///
     /// - parameter cursor: The cursor whose elements feed the collection.
-    @inlinable
     public init<C: Cursor>(_ cursor: C) throws where C.Element == Element {
         self.init()
         while let element = try cursor.next() {
@@ -36,7 +35,6 @@ extension RangeReplaceableCollection {
     /// - parameter cursor: The cursor whose elements feed the collection.
     /// - parameter minimumCapacity: Prepares the returned collection to store
     ///   the specified number of elements.
-    @inlinable
     public init<C: Cursor>(_ cursor: C, minimumCapacity: Int) throws where C.Element == Element {
         self.init()
         reserveCapacity(minimumCapacity)
--- a/GRDB/Core/Database+Statements.swift
+++ b/GRDB/Core/Database+Statements.swift
@@ -588,7 +588,6 @@ extension Database {
     /// Returns the authorizer that should be used during statement execution
     /// (this allows preventing the truncate optimization when there exists a
     /// transaction observer for row deletion).
-    @usableFromInline
     func statementWillExecute(_ statement: Statement) throws -> StatementAuthorizer? {
         // Two things must prevent the statement from executing: aborted
         // transactions, and database suspension.
@@ -604,7 +603,6 @@ extension Database {
     
     /// May throw a cancelled commit error, if a transaction observer cancels
     /// an empty transaction.
-    @usableFromInline
     func statementDidExecute(_ statement: Statement) throws {
         if statement.invalidatesDatabaseSchemaCache {
             clearSchemaCache()
@@ -614,7 +612,6 @@ extension Database {
     }
     
     /// Always throws an error
-    @usableFromInline
     func statementDidFail(_ statement: Statement, withResultCode resultCode: Int32) throws -> Never {
         // Failed statements can not be reused, because sqlite3_reset won't
         // be able to restore the statement to its initial state:
--- a/GRDB/Core/Database+Statements.swift
+++ b/GRDB/Core/Database+Statements.swift
@@ -588,6 +588,7 @@ extension Database {
     /// Returns the authorizer that should be used during statement execution
     /// (this allows preventing the truncate optimization when there exists a
     /// transaction observer for row deletion).
+    @usableFromInline
     func statementWillExecute(_ statement: Statement) throws -> StatementAuthorizer? {
         // Two things must prevent the statement from executing: aborted
         // transactions, and database suspension.
@@ -603,6 +604,7 @@ extension Database {
     
     /// May throw a cancelled commit error, if a transaction observer cancels
     /// an empty transaction.
+    @usableFromInline
     func statementDidExecute(_ statement: Statement) throws {
         if statement.invalidatesDatabaseSchemaCache {
             clearSchemaCache()
@@ -612,6 +614,7 @@ extension Database {
     }
     
     /// Always throws an error
+    @usableFromInline
     func statementDidFail(_ statement: Statement, withResultCode resultCode: Int32) throws -> Never {
         // Failed statements can not be reused, because sqlite3_reset won't
         // be able to restore the statement to its initial state:
--- a/GRDB/Core/Database.swift
+++ b/GRDB/Core/Database.swift
@@ -1596,6 +1596,7 @@ extension Database {
     }
     
     /// Confirms or cancels the changes performed by a transaction or savepoint.
+    @frozen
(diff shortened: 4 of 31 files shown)


asifmohd:
>  would you please check if this change fits your needs?

Able to create xcframeworks of GRDB now, without any issues using Xcode 12.5.1. Thanks a lot for picking this up and resolving this @groue 🙇

groue:
Thanks @asifmohd 😉

groue:
Shipped in v5.10.0

mallman:
@groue, do you think the removal of the inline-related annotations will affect performance? Do you have performance regression tests?

groue:
> Do you think the removal of the inline-related annotations will affect performance? Do you have performance regression tests?

The performance tests that help generating https://github.com/groue/GRDB.swift/wiki/Performance were not affected in any significant way. One of them was actually slightly improved, I don't remember which one, due to the enhanced decoding of low-level types such as Int, Double, or String. The compiler can jump straight to specific [C decoding functions](https://www.sqlite.org/c3ref/column_blob.html) without involving any protocol witness and other Swift dynamic trampolines :-)

I'm not able to spot all performance regressions, though (such as the infamous #983), so please report any oddity you can spot :-)

If you notice an unwanted compiler warning due to a missing `@frozen enum`, please report it as well.

groue:
@mallman, re-reading my previous message, I notice that maybe I did not run the performance tests with explicit `BUILD_LIBRARY_FOR_DISTRIBUTION` flag on 5.10.0. So here are some fresh figures, with 5.9.0 as a baseline. Again, I do not see anything significant difference:

<details>
    <summary>GRDB 5.9.0</summary>

|                                  |      |
|:-------------------------------- | ----:|
| **Column indexes**               |      |
| Fetch                            | 0.05 |
| Insert                           | 0.05 |
| **Column names**                 |      |
| Fetch                            | 0.10 |
| Insert                           | 0.08 |
| **Records**                      |      |
| Fetch                            | 0.11 |
| Insert                           | 0.22 |
| **Codable Records**              |      |
| Fetch                            | 0.20 |
| Insert                           | 0.22 |
| **Optimized Records**            |      |
| Fetch                            | 0.06 |
| Insert                           | 0.06 |
| **Records with change tracking** |      |
| Fetch                            | 0.22 |
| Insert                           | 0.25 |


</details>

<details>
    <summary>GRDB 5.10.0 with <tt>BUILD_LIBRARY_FOR_DISTRIBUTION</tt> set in GRDB.xcodeproj</summary>

|                                  |      |
|:-------------------------------- | ----:|
| **Column indexes**               |      |
| Fetch                            | 0.07 |
| Insert                           | 0.05 |
| **Column names**                 |      |
| Fetch                            | 0.12 |
| Insert                           | 0.09 |
| **Records**                      |      |
| Fetch                            | 0.13 |
| Insert                           | 0.17 |
| **Codable Records**              |      |
| Fetch                            | 0.23 |
| Insert                           | 0.22 |
| **Optimized Records**            |      |
| Fetch                            | 0.08 |
| Insert                           | 0.05 |
| **Records with change tracking** |      |
| Fetch                            | 0.25 |
| Insert                           | 0.19 |

</details>

Those report were generated on a iMac (Retina 5K, 27-inch, 2017), with Xcode 12.5.1, by running the following command, and editing the output:

```sh
make test_performance | Tests/parsePerformanceTests.rb | Tests/generatePerformanceReport.rb
```

mallman:
Hi @groue. The difference in performance results you posted above between 5.9 and 5.10 look significant to me, but I may be misinterpreting them. I will spend some time educating myself and report back with a more informed opinion.

mallman:
I did a quick initial run of `make test_performance` on v5.9.0 to get started and saw about a dozen errors of the form

```
BUG IN CLIENT OF libsqlite3.dylib: database integrity compromised by API violation: vnode unlinked
while in use: /private/var/folders/1d/3bm2td7j3zs2rsfl4y05b0xr0000gn/T/GRDBPerformanceTests-75C6FBBF-C488-4AA7-9F7F-9870DCEACDC5-16774-0001C0BC54CD8D2D.sqlite
```

logged to the console. Do you know what's behind these?

groue:
> The difference in performance results you posted above between 5.9 and 5.10 look significant to me, but I may be misinterpreting them.

Subsequents runs of XCTest `measure` can vary, so I'm used to look at them with a grain of salt.

See the output of a few runs below, where I marked with 🔀 the lines that exhibit variations ≥ 10%:

|                                  |      |      |      |      |      |      |
|:-------------------------------- | ----:| ----:| ----:| ----:| ----:| ----:|
| **Column indexes**               |      |      |      |      |      |      |
| 🔀 Fetch                            | 0.08 | **0.07** | 0.08 | 0.07 | **0.10** | 0.07 |
| Insert                           | 0.07 | 0.07 | 0.07 | 0.07 | 0.07 | 0.07 |
| **Column names**                 |      |      |      |      |      |      |
| Fetch                            | 0.13 | 0.13 | 0.13 | 0.13 | 0.13 | 0.13 |
| Insert                           | 0.10 | 0.10 | 0.10 | 0.10 | 0.10 | 0.10 |
| **Records**                      |      |      |      |      |      |      |
| 🔀 Fetch                            | 0.13 | 0.13 | **0.14** | 0.14 | 0.14 | **0.12** |
| 🔀 Insert                           | 0.20 | 0.21 | **0.22** | 0.22 | 0.22 | **0.19** |
| **Codable Records**              |      |      |      |      |      |      |
| 🔀 Fetch                            | **0.31** | 0.25 | 0.25 | 0.25 | 0.26 | **0.23** |
| Insert                           | 0.25 | 0.26 | **0.27** | 0.27 | 0.25 | **0.24** |
| **Optimized Records**            |      |      |      |      |      |      |
| 🔀 Fetch                            | **0.08** | 0.08 | **0.09** | 0.08 | 0.08 | 0.08 |
| 🔀 Insert                           | **0.06** | **0.07** | 0.07 | 0.07 | 0.07 | 0.06 |
| **Records with change tracking** |      |      |      |      |      |      |
| 🔀 Fetch                            | **0.33** | 0.28 | 0.29 | **0.27** | 0.28 | 0.28 |
| 🔀 Insert                           | 0.22 | **0.24** | 0.22 | 0.24 | 0.22 | **0.21** |

I used to run them multiple times and grab the best figures, so that the [performance comparison](http://github.com/groue/GRDB.swift/wiki/Performance) would be "fair". I no longer do that, and just grab one result.

> `database integrity compromised by API violation: vnode unlinked`

My understanding of this message is that the database file is deleted before the connection closes. Some tests perform cleanup too early. I also see those message sometimes, and was too lazy to remove them. I agree this is not the best experience.

groue:
> To be honest with you, I'm not quite sure my way to test BUILD_LIBRARY_FOR_DISTRIBUTION performance above has any relevance.

Well, I have some evidence that the compiler actually specializes decoding of basic scalar types, because I profiled tests with Instruments. Both 18e9b1b9aa5b557ee19bcf3af61a596548712500 and e2c9734df12076fbe579f76265f01eb090b6ae86 had an actual impact, and were necessary in order to preserve pre-5.10.0 overall performance.

barakwei:
Hi @groue - Thanks for this truly amazing work!
Since ABI stability is not guaranteed, will it be technically possible in the future to add a compiled binary XCFramework with *static* library to the releases?

groue:
Hello @barakwei, 

it is not planned to add pre-built libraries in the releases. Support for ABI stability is not a *commitment* to ABI stability. Think of this as a *convenience* for users who want to cache their own builds. That's nothing more.

barakwei:
Thanks @groue for your reply
There might have been a misunderstanding on my side. To my understanding, ABI stability is relevant where you build the library and app separately and you want to replace the library without rebuilding the app. So this is relevant only when using dynamic libraries. My understanding was that for *static* libraries this is not relevant, as in any case you need to rebuild your app to use a new version.
So in theory at least, distributing a static XCframework will not require any commitment to ABI stability. It should just be "even more convenience".

groue:
Yeah. But no. See https://github.com/groue/GRDB.swift/issues/1027

barakwei:
I totally agree that ABI stability doesn't make any sense here.
Thanks for your time and for creating and maintaining this lib!

PR #1052, opened 2021-09-08
Change init to fix Swift error

jason-gabriele:
I was getting the following error when building GRDB 5.11 through Cocoapods:

```
❌  Pods/GRDB.swift/GRDB/Core/Support/Foundation/DatabaseDateComponents.swift:87:14: 'let' property 'dateComponents' may not be initialized directly; use "self.init(...)" or "self = ..." instead

        self.dateComponents = components.dateComponents
             ^



❌  Pods/GRDB.swift/GRDB/Core/Support/Foundation/DatabaseDateComponents.swift:88:14: 'let' property 'format' may not be initialized directly; use "self.init(...)" or "self = ..." instead

        self.format = components.format
```

So I just changed it to use self.init as it suggested. Built with Xcode 12.5.1

### Pull Request Checklist

- [X] This pull request is submitted against the `development` branch.
- [N/A] Inline documentation has been updated.
- [N/A] README.md or another dedicated guide has been updated.
- [X] Changes are tested.

### DIFF
--- a/GRDB/Core/Support/Foundation/DatabaseDateComponents.swift
+++ b/GRDB/Core/Support/Foundation/DatabaseDateComponents.swift
@@ -84,8 +84,7 @@ public struct DatabaseDateComponents: DatabaseValueConvertible, StatementColumnC
         guard let components = optionalComponents else {
             return nil
         }
-        self.dateComponents = components.dateComponents
-        self.format = components.format
+        self.init(components.dateComponents, format: components.format)
     }
     
     // MARK: - DatabaseValueConvertible adoption


groue:
Hello @jason-gabriele, I'm not aware of a build error of GRDB 5.11 with Xcode 12.5.1. What precise Swift version are you using?

jason-gabriele:
I'm building in Xcode 12.5.1 and it's imported via Cocoapods. I don't think I have any strange settings applied. My swift version is marked as 'Unspecified', but I tried changing it to 'Swift 5' and it had no effect. 

`pod 'GRDB.swift/SQLCipher', :git => 'git@github.com:groue/GRDB.swift.git'`

jason-gabriele:
I was able to reproduce it here: https://github.com/jason-gabriele/GRDBTest

Thanks

groue:
Thank you @jason-gabriele, you spotted a failure that happens when `BUILD_LIBRARY_FOR_DISTRIBUTION` is set. I do not know how I missed it during the release of v5.10.0. I'll check that. Anyway, thanks a lot 👍

Issue #1074, opened 2021-10-08
Provide API for reporting incremental database backup progress


mallman:
The GRDB database [backup API](https://github.com/groue/GRDB.swift/#backup) is implemented using SQLite's [online backup API](https://sqlite.org/c3ref/backup_finish.html). At

https://github.com/groue/GRDB.swift/blob/v5.12.0/GRDB/Core/Database.swift#L1204

GRDB calls `sqlite3_backup_step(backup, -1)`. According to the SQLite documentation, this will copy all remaining source pages. Since this is the only call into the `sqlite3_backup_step` function, GRDB performs the entire backup in one call. This simplifies the backup logic at the expense of opacity to the caller. Specifically, we do not know anything about the progress being made in the backup.

By providing a positive integer `n` to `sqlite3_backup_step(backup, n)`, we can request an incremental database backup step copying up to `n` additional blocks. Additionally, SQLite provides the `sqlite3_backup_remaining` and `sqlite3_backup_pagecount` functions for calculating backup progress. The [SQLite backup documentation](https://sqlite.org/backup.html) provides an example of backing up a live database (Example 2: Online Backup of a Running Database) with a progress reporting callback. I believe we can use this example as a model for an implementation of a backup capability for GRDB that reports progress incrementally.

I will work on a PR for this issue. It _shouldn't_ be too complicated. We'll see.

groue:
Hello @mallman 

indeed the backup api is currently very naive. I'm quite ears for a progressive api 👍

PR #1075, opened 2021-10-08
Progress-reporting database backups

mallman:
This PR addresses issue #1074. It provides a small update to the GRDB backup API that allows a user to receive backup progress updates.

The existing backup unit tests have been enhanced, however I have not yet tried using this patch in a "real world" scenario with a large/huge database. I'm particularly interested in tuning the backup page step size to something useful. The value I've set, 5, is simply taken from the SQLite example. I have no idea if this is a sane value, so I'm keeping this as a draft PR until I've had the opportunity to test and tune that value.

### Pull Request Checklist

- [x] Changes are unit tested.
- [ ] Changes are "real world" tested.


mallman:
> > I have no idea why the build is failing. Help?
> 
> Hello @mallman, what do you do to get this failure?

Okay. I couldn't figure it out originally, but the build is working now.

mallman:
I had another thought on why this API does not return some kind of progress object. The function as written in this PR is not asynchronous—it doesn't return until the backup is complete. In that sense, it is an API-compatible replacement for the current backup function (assuming we use a default page size of -1). The only difference introduced by this PR is the provision of an optional capability to receive periodic progress reports while the backup is running.

groue:
About the above documentation that talks about throwing from the progress callback in order to abandon the backup, and the fact that this error is rethrown by the backup method: this is unfinished thoughts, and quick and dirty documentation.

I think it is valuable to let the user abandon the backup. It is possible according to the [SQLite doc](https://www.sqlite.org/c3ref/backup_finish.html) (search for "abandon").

But we must let the user understand clearly what is the final state of the destination database: _copied, or not_. In this context, I wonder when happens when one throws from the last step:

```swift
// 😈
try source.backup(to: destination) { completed, total in
    if completed == total { throw MyErrror() }
}
```

In the above sample code, is our implementation able to abandon the backup, or not? Maybe yes, maybe not. It's better if we can. Anyway, what's paramount is that if the destination database **is** copied, then `backup` method **must not throw**. On the other side, if the backup has failed in any way, then an error must be thrown.

This is important, considering the component that calls the `backup` method may not be the component that provides the callback:

```swift
func someComplexBackup(progress: ...) throws {
    do {
        try source.backup(to: destination, progress: progress)
        // Here the backup MUST have completed.
        // It is a GRDB bug if we can end up here with an failed/abandoned backup.
    } catch {
        // Here the backup MUST have failed.
        // It is a GRDB bug if we can end up here with a completed backup.
    }
}
```

mallman:
I see the dilemma.

In Example 2 from https://sqlite.org/backup.html, SQLite's "solution" is to omit a means by which the caller can abort a backup in progress. The callback is simply informative.

However, providing some kind of cancellation mechanism through the progress callback is very attractive for a user-oriented interface. We are in agreement here.

What if we omitted a call to the user-supplied progress callback if `sqlite3_backup_step()` returns `SQLITE_DONE`? That is to say, we only call the callback (and thereby provide the user a means to cancel the backup in progress) if `sqlite3_backup_step()` returns `SQLITE_OK`?

Generally speaking, should callers expect to get a "final" callback stating 100% completion of their task? From an end user's UI perspective we can clearly say the answer to this question is "no". As one concrete example, I recently saw my phone's iOS update go from something like 27% to being done—specifically, the progress bar never reported 100%.

groue:
> What if we omitted a call to the user-supplied progress callback if `sqlite3_backup_step()` returns `SQLITE_DONE`? That is to say, we only call the callback (and thereby provide the user a means to cancel the backup in progress) if `sqlite3_backup_step()` returns `SQLITE_OK`?

I would be surprised, as a user, if the last step would not trigger the callback, if I would not be _sure_ the callback was called _at least once, with completedPageCount == totalPageCount_.

If a user wants to feed a `Progress`, then it would look like:

```swift
let progress = Progress()
try source.backup(to: destination, ...) { completedPageCount, totalPageCount in
    progress.totalUnitCount = totalPageCount
    progress.completedUnitCount = completedPageCount
}
```

Otherwise, it's just very confusing:

```swift
let progress = Progress()
try source.backup(to: destination, ...) { completedPageCount, totalPageCount in
    progress.totalUnitCount = totalPageCount
    progress.completedUnitCount = completedPageCount
}
// Deal with last step
// But what if callback is never called? What is totalUnitCount?
progress.completedUnitCount = progress.totalUnitCount
```

And of course such code must also work without modification if pagesPerStep is -1.

groue:
Hello @mallman, maybe you met some unexpected difficulty?

mallman:
> Hello @mallman, maybe you met some unexpected difficulty?

Sorry, and thanks for your patience.

I've made significant progress. I've updated the API and implementation as we discussed, however I haven't finished the new tests. These are proving trickier than anticipated, and I got stuck on them. I'll have another go today and try to push a new commit.

mallman:
> Hello @mallman, maybe you met some unexpected difficulty?

Oh, also I took a break camping the last few days, and I'm just catching up on e-mail and other correspondence.

groue:
I'm glad you're fine - and you have all the time you want! 🙂

mallman:
The latest commit adds the cancellation API we discussed, as well as a backup API for the `Database` type. I spent a considerable amount of time trying to reduce code duplication in the tests to the bare minimum. There is still a lot of duplication, but I don't see a clear way to make further progress.

Once we've settled on the API and implementation I will work on documentation.

mallman:
Hmmm... seems I spent so much time refactoring the existing backup tests I forgot to add tests for the cancellation functionality. Derp. I don't think I'll get to that today, but it's coming.

mallman:
Hi @groue. Sorry for the delay on the incremental backup tests. I've been occupied by other priorities, but I will make time for them soon.

In the meantime, I want to ask if you have time to review the changes as they are, especially if the API looks good to you. If the API looks good I will continue to focus on testing the implementation.

groue:
All right, I'll have a look really soon :-) Thank you @mallman!

groue:
Hello @mallman,

I don't quite understand the reason for introducing a public `DatabaseBackupProgress` protocol, a `DatabaseBackupProgressImpl` private class, and a public `DatabaseBackupCancelled` error, when the callback could be implemented as a throwing closure (see https://github.com/groue/GRDB.swift/pull/1075#issuecomment-944058380). Do I miss something?

mallman:
> Hello @mallman,
> 
> I don't quite understand the reason for introducing a public `DatabaseBackupProgress` protocol, a `DatabaseBackupProgressImpl` private class, and a public `DatabaseBackupCancelled` error, when the callback could be implemented as a throwing closure (see [#1075 (comment)](https://github.com/groue/GRDB.swift/pull/1075#issuecomment-944058380)). Do I miss something?

Quite frankly, it's been so long since I wrote this code I can't remember why I wrote the API this way. So your question could well be posed—do _we_ miss something? 😉  I'm doing some careful review myself to uncover the rationale for adopting this API in my PR as opposed to your simpler alternative.

Having spent a couple of hours ruminating, I have an idea for a clear response to your question. However, my analysis hinges on an assumption about the API you propose. Specifically, how do you propose the user request the backup be cancelled? By throwing from the callback? That's my assumption. If not that, then how?

Cheers.

groue:
> Specifically, how do you propose the user request the backup be cancelled? By throwing from the callback? That's my assumption.

[Absolutely!](https://github.com/groue/GRDB.swift/pull/1075#issuecomment-944068149)

Tests/GRDBTests/BackupTestCase.swift
```diff
@@ -0,0 +1,135 @@
...
+
+        XCTAssertThrowsError(
+            try source.backup(to: destination, pageStepSize: 1) { completedPages, totalPages in
+                XCTAssertLessThan(completedPages, totalPages)
+                throw AbandonBackupError()
+            }
+        )
+
+        // Assert that the items table is as it was before the backup was abandoned.
+        // Is this a valid assertion?
```
mallman, line 44: I'm going to check with the mailing list on this.

mallman:
I shelved my response to your question. Instead I revised the API to align with what you wrote in your comment. Simpler is better. The only aspect I feel may surprise the user is the fact that an `Error` thrown from the `progress` callback will be silently ignored if the backup is already complete. To rethrow it would violate our rule that the backup must have failed if the call to the `backup` function throws. However, this behavior is documented in the code documentation.

groue:
That's great, @mallman :-)

> The only aspect I feel may surprise the user is the fact that an Error thrown from the progress callback will be silently ignored if the backup is already complete.

Indeed the SQLite documentation says that "the backup operation is complete [when] sqlite3_backup_step() returns SQLITE_DONE", so we can't cancel the backup from the last call to the `progress` function.

We don't do worse than SQLite itself, so this is just OK. And you documented it well 👍

All right, so it looks like we'll merge this soon now 🎉

GRDB/Core/Database.swift
```diff
@@ -1243,14 +1243,72 @@ public final class Database: CustomStringConvertible, CustomDebugStringConvertib
...
+    ///       step. By default, all pages are copied in one single step.
+    ///     - progress: An optional function that is notified of the backup
+    ///       progress.
+    ///     - completedPageCount: The number of copied pages.
+    ///     - totalPageCount: The total number of pages to be copied.
+    /// - throws: The error thrown by `progress` if the backup is abandoned, or
+    ///   any `DatabaseError` that would happen while performing the backup.
+    public func backup(
+        to destDb: Database,
+        pagesPerStep: Int32 = -1,
```
mallman, line 1293: I'm thinking we might want to avoid the C-style sentinel value and make this

```swift
pagesPerStep: Int32? = nil
```

instead. When `pagesPerStep == nil`, we pass -1 to `sqlite3_backup_step`. Otherwise we pass `pagesPerStep`. Thoughts? I'm a little wary of making `pagesPerStep == nil` mean "backup in one step", though using `-1` to mean that is not very Swifty either.

groue: I understand your question. I would have clearly preferred `nil` a few years ago, but I now prefer to leave raw SQLite concepts alone, unmodified, especially for advanced/expert apis. This helps users who already know SQLite, because they don't have to learn anything new. `pagesPerStep` is a clearly a raw SQLite concept, better understood by reading the SQLite doc. `-1` is a negative value that matches this doc, when `nil` would need an extra interpretation step. And since `-1` is a default value, it won't bother non-experts. To me, `-1` is the winner!

mallman:
I'm working on updating the readme.

I've made an assertion in the code documentation that

> An error thrown from `progress` when `completedPageCount == totalPageCount` is silently ignored.

That is, the last call to `sqlite3_backup_step` returned `SQLITE_DONE` if `completedPageCount == totalPageCount`. I'm not sure if this is a valid assumption. I've asked a question on the SQLite forum to see if I can get clarification on this. If I don't get a confident confirmation, I will probably excise this assumption from the documentation.

GRDB/Core/Database.swift
mallman: WFM

groue:
> I've made an assertion in the code documentation that

You mean, asserting that `completedPageCount == totalPageCount` on the last step? I'd rather not. As the saying goes, _don't test the framework_. Our implementation does not rely on this. And a hick-up in SQLite shouldn't make GRDB choke in debug builds, or have GRDB tests fail.

Ignoring the error on SQLITE_DONE, as you do, is the correct technique.

Our documentation can be updated without any reference to completedPageCount/totalPageCount comparison: "If `progress` throws before the backup is complete, the backup is aborted and the error is rethrown. The last call to `progress` marks the successful completion of the backup. On this last step, any error thrown by `progress` is ignored."

mallman:
Question. Do we want to allow the client to know when throwing an error from the progress callback will abort the backup and vice-versa?

I was hoping to be able to use completed versus total page counts as a guaranteed proxy for backup completion. However, we can't do that. Therefore, as is, the client cannot know what behavior to expect when throwing an error from the callback—it will either rethrow or be ignored.

However, we could add a third parameter to the callback to signal that the backup can no longer be aborted. It would look something like this:

```swift
progress: ((_ completedPageCount: Int, _ totalPageCount: Int, _ abortable: Bool) throws -> ())? = nil)
```

In `backupInternal`, the case statements in the while loop would look like

```swift
case SQLITE_DONE:
    try? afterBackupStep?(completedPages, totalPages, false)
    break backupLoop
case SQLITE_OK:
    try afterBackupStep?(completedPages, totalPages, true)
```

groue:
Do you have a use case in mind?

groue:
`sqlite3_backup_finish` is documented to release a lock, so it does I/O. If the database is hosted on a network disk, this can take some time.

I really do not like the extra boolean. At the same time, I have difficulties measuring of the chances for the user to witness:

- `completedPageCount >= totalPageCount` after sqlite3_step() has returned `OK` (user who relies on these numbers could falsely believe the backup is completed, or witness several "completions")
- `completedPageCount < totalPageCount` after sqlite3_step() has returned `DONE`  (user who relies on these numbers could miss the opportunity to perform specific backup completion handling from the `progress` closure).

Reading the SQLite doc does not help me reach a definitive conclusion.

Grrr

OK maybe our api is fundamentally flawed.

mallman:
I think we can work this out if we make this some kind of experts-only API. The GRDB step-wise backup API reflects the low-level SQLite step-wise backup behavior and all its subtlety. The progress we provide via the callback is a direct reflection of what we get from the `sqlite3_backup_remaining` and `backup_pagecount` calls. Users may be aware that if they want some hard assurances about the behavior of this API they have to consult the SQLite backup documentation and/or the mailing list.

If users want a simple API with simple semantics, they should pass `stepsPerPage = -1` and `progress = nil`, which is the default anyway.

groue:
I fully agree. Thanks for keeping the boat afloat :-)

Your concern about letting users know about the last step is valid, and still not addressed. The extra boolean argument in the closure is really ugly, so what about improving slightly the api like this:

```swift
public BackupProgress {
    public var completedPageCount: Int
    public var totalPageCount: Int
    public var isCompleted: Bool // set iff SQLITE_DONE
}
public func backup(
    to destDb: Database,
    pagesPerStep: Int32 = -1,
    progress: ((BackupProgress) throws -> Void)? = nil)
{ ... }
```

mallman:
I've (finally) started some "real world" testing of this API with my "real" project. From a usability perspective my immediate reaction to writing the progress callback is I can't remember which parameter is which. I think `BackupProgress` struct is a better way to go. I also like the idea of getting an actual completion flag that mirrors the result from the SQLite API. Since we are exposing some of SQLite's API in this interface, we should document what `isCompleted == true` means and vice-versa.

I'm working on another commit for this PR with this change and some readme updates.

mallman:
@groue I would still like to run some more "real world" tests, but otherwise I'd say this PR is ready for review.

groue:
Thanks @mallman! We're close to the end of the year, so I may be slow to review.

groue:
Hello @mallman, this looks just great.

I have merged the development branch, and exposed `DatabaseBackupProgress.remainingPageCount` for raw `sqlite3_backup_remaining()` access. Consider that we advise the user to read the SQLite documentation: the raw SQLite concepts found in the documentation should be available through GRDB apis so that the user feels empowered. In other words, when one goes expert mode, one should do it unabatedly. 

The handy `completedPageCount` property is still there, but it is now computed.

Let's wait for tests to run before I can merge 👍

groue:
We were just missing a test file in the SQLCipher 3 & 4 test projects, as revealed by `make smokeTest`.

I recently updated [CONTRIBUTING.md](https://github.com/groue/GRDB.swift/blob/master/CONTRIBUTING.md), you'll find a mention of `make smokeTest` there.

I should specify there and in the pull request template that this running this suite is recommended before one submits a PR. But most importantly I should definitely beef up the CI test suite 😅

groue:
The PR was not on the development branch, I had to force-push in order to revert the merge.

groue:
And https://github.com/mallman/GRDB.swift/tree/dev/issue-1074-incremental_backup_api was manually merged on https://github.com/groue/GRDB.swift/tree/development. 👍

groue:
Shipped in [v5.18.0](https://github.com/groue/GRDB.swift/tree/v5.18.0). Thanks @mallman!

mallman:
> The PR was not on the development branch, I had to force-push to master in order to revert the merge.

Mea culpa. I'll make a note of this for next time.

PR #1088, opened 2021-10-23
Shared ValueObservation

### LINKED ISSUE #1083: Sharing ValueObservations without Combine or RxSwift
**The gist:** The documentation recommends to share observations when possible. It also includes an example for RxSwift. However, in a regular UIKit application that doesn't use RxSwift or Combine, how do you recommend sharing an observation?

---

The particular use case is:
- I have a controller that controls audio playback. It holds the current playlist and the currently playing song. It observes both the playlist and the song, because changes in the playlist impact skip forward/back behaviour, and changes in the song impact information shown in the iOS Media Center.
- I have a second view controller in my app showing information about the currently playing song - a "now playing" view. This view controller isn't always visible/in the hierarchy, because users can close the view and browse through other parts of the app while the audio continues to play. The now-playing view also needs to stay up-to-date with changes in the currently playing playlist (to enable/disable skip-forward/back buttons), as well as changes to the song (such as changes to its title).

I have created two observations in the playback controller. The playback controller is a singleton that I can access throughout my application. Is it possible, in the now-playing view controller, to say something like:

```
self.songObservation = PlaybackController.shared.songObservation.replay()
```

The alternative solution is to just send notifications instead. But if there was a way to simply copy and share an observation, that would be fantastic.

And since I have the opportunity to write this here: thanks so much for your work on GRDB! ❤️

groue:
This pull request fixes #1083.

```swift
let observation = ValueObservation.tracking { db in ... }
let sharedObservation = observation.shared(in: dbQueue)

let cancellable = sharedObservation.start(
    onError: { error in ... }
    onChange: { value in ...})
```

Like a regular ValueObservation, a shared observation can be subscribed to several times. However it shares a single underlying database observation for all subscriptions, and thus spares database resources.

The sharing only applies if you start observing the database from the same `SharedValueObservation` instance:

```swift
// NOT shared
let cancellable1 = ValueObservation.tracking { db in ... }.shared(in: dbQueue).start(...)
let cancellable2 = ValueObservation.tracking { db in ... }.shared(in: dbQueue).start(...)

// Shared
let sharedObservation = ValueObservation.tracking { db in ... }.shared(in: dbQueue)
let cancellable1 = sharedObservation.start(...)
let cancellable2 = sharedObservation.start(...)
```

You can choose the scheduling of fresh values:

```swift
// Notifies the initial value synchronously, and schedules future changes
// asynchronously on the main queue. The .immediate scheduling
// requires that observation is started from the main queue.
let sharedObservation = observation.shared(in: dbQueue, scheduling: .immediate)

// The default: asynchronously on the main queue
let sharedObservation = observation.shared(in: dbQueue, scheduling: .async(onQueue: .main))

// Asynchronously on the DispatchQueue of your choice
let sharedObservation = observation.shared(in: dbQueue, scheduling: .async(onQueue: myQueue))
```

A shared observation starts observing the database as soon as it is subscribed. You can choose if database observation should stop, or not, when its number of subscriptions drops down to zero:

```swift
// Only stops observing the database when the shared observation is deallocated,
// and all subscriptions are cancelled.
//
// This extent prevents the shared observation from recovering from database
// errors. To recover from database errors, create a new shared
// SharedValueObservation instance.
let sharedObservation = observation.shared(in: dbQueue, extent: .observationLifetime)

// The default: stops observing the database when the number of subscriptions 
// drops down to zero, and restart database observation on the next subscription.
//
// Database errors can be recovered by resubscribing to the shared observation.
let sharedObservation = observation.shared(in: dbQueue, extent: .whileObserved)
```

TODO:
- [x] SharedValueObservation
- [x] Tests
- [x] Documentation
- [x] Combine publisher
- [ ] ~Operator `map`~ (won't implement, see [below](#issuecomment-950317611))
- [ ] ~Operator `removeDuplicates`~ (won't implement, see [below](#issuecomment-950317611))

cc @tcwalther

### DIFF
--- a/README.md
+++ b/README.md
@@ -6224,7 +6224,7 @@ let sharedObservation = ValueObservation
 >         .sink(...)
 >     ```
 >
-> - Unlike `ValueObservation`, `SharedValueObservation` prevents the database from closing. As long as there exists a `SharedValueObservation` instance, or an active suscription to a shared observation, the database connectionn won't close. 
+> - Unlike `ValueObservation`, `SharedValueObservation` prevents the database from closing. As long as there exists a `SharedValueObservation` instance, or an active suscription to a shared observation, the database connection won't close. 
 
 
 ### ValueObservation Performance


groue:
`SharedValueObservation` has shipped in v5.13.0 - [documentation](https://github.com/groue/GRDB.swift/blob/v5.13.0/README.md#valueobservation-sharing)

tcwalther:
Wow, really impressive. Thanks so much, @groue!

Issue #1101, opened 2021-12-01
ValueObservation while excluding specific columns


freak4pc:
### What did you do?

Let's say I have a model called `MyModel`. It has three props: `a`, `b` and `c`.
I'm using ValueObservation over it, for example:

```swift
let observation = ValueObservation.tracking { MyModel.fetchAll() }
return observation.start(
    in: dbQueue,
    onError: ...,
    onChange: ...
)
```

But, I'd like that specifically changes to column `b` won't cause a new set of values to be sent from the observation (basically ignoring changes to that column).

Is that possible? cc/ @nirlachman-monday

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 5.12
**Installation method:** CocoaPods
**Xcode version:** 12.5
**Swift version:** 5.4
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** Big Sur / Monterey

groue:
Hello @freak4pc,

It is not currently possible to exclude some columns from the observation.

It used to be... But the `ValueObservation.tracking { whatever }` syntax was so simple & sexy I dropped everything else! 😅 

If column `c` should be excluded from _tracking_, and if you can exclude it as well from _fetching_, you can try one those two techniques:

1. Fetch a "smaller model":

```swift
struct MySmallModel: FetchableRecord, TableRecord, Decodable {
    static let databaseTableName = MyModel.databaseTableName
    static let databaseSelection: [SQLSelectable] = [Column("a"), Column("b")]
    var a: ...
    var b: ...
}

let observation = ValueObservation.tracking(MySmallModel.fetchAll)
return observation.start(
    in: dbQueue,
    onError: ...,
    onChange: ...
)
```

2. Fetch raw rows, and turn them into `MyModel` with some dummy `c` value:

```swift
let request = MyModel
    .select(Column("a"), Column("b"))
    .asRequest(of: Row.self)
let observation = ValueObservation
    .tracking(request.fetchAll)
    .map { row in MyModel(a: row["a"], b: row["b"], c: someDummyValue) }
return observation.start(
    in: dbQueue,
    onError: ...,
    onChange: ...
)
```

You can use [`print()`](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservationprint) in order to precisely verify which tables & columns are tracked by the observation.

I think those techniques have advantage over excluding `c` from tracking, because you _control_ the value of `c` (missing, or initialised to some specific value).

Do you think one of them could work for you?

[...]

PR #1102, opened 2021-12-02
Explicit tracked region for ValueObservation

groue:
This PR addresses #1101

ValueObservation has a new factory method, `ValueObservation.tracking(region:fetch:)`, which allows to entirely separate the **observed region** from the **fetched value** itself:

```swift
// Triggered by all transactions
ValueObservation.tracking(
    region: .fullDatabase,
    fetch: { db in ... })

// Triggered by all changes to player with rowid 1
ValueObservation.tracking(
    region: Player.filter(id: 1),
    fetch: { db in ... })

// Triggered by all changes to the score column of the player table
ValueObservation.tracking(
    region: SQLRequest("SELECT score FROM player"),
    fetch: { db in ... })

// Triggered by all changes to player or team tables
ValueObservation.tracking(
    region: Player.all(), Team.all(),
    fetch: { db in ... })
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -5807,6 +5807,7 @@ Tracked changes are insertions, updates, and deletions that impact the tracked v
 - [ValueObservation Scheduling](#valueobservation-scheduling)
 - [ValueObservation Operators](#valueobservation-operators): [map](#valueobservationmap), [removeDuplicates](#valueobservationremoveduplicates), ...
 - [ValueObservation Sharing](#valueobservation-sharing)
+- [Specifying the Region Tracked by ValueObservation](#specifying-the-region-tracked-by-valueobservation)
 - [ValueObservation Performance](#valueobservation-performance)
 - :blue_book: [Combine Publisher](Documentation/Combine.md#database-observation)
 
@@ -6224,6 +6225,23 @@ let sharedObservation = ValueObservation
 >
 > - Unlike `ValueObservation`, `SharedValueObservation` retains the database connection. As long as there exists a `SharedValueObservation` instance, or an active suscription to a shared observation, the database connection won't be deinitialized. 
 
+
+### Specifying the Region Tracked by ValueObservation
+
+While the standard `ValueObservation.tracking { db in ... }` method lets you track changes to a fetched value and receive any changes to it, sometimes your use case might require more granular control. 
+
+Consider a scenario where you'd like to get a specific Player's row, but only when their `score` column changes. You can use `tracking(region:fetch:)` to do just that:
+
+```swift
+let observation = ValueObservation.tracking(
+    // Define what database region constitutes a "change"
+    region: Player.select(Column("score")).filter(id: 1),
+    // Define what to fetch upon such change
+    fetch: { db in try Player.fetchOne(db, id: 1) }
+)
+```
+
+This overload of `ValueObservation` lets you entirely separate the **observed region** from the **fetched value** itself, providing utmost flexibility. See [DatabaseRegionConvertible] for more information.
 
 ### ValueObservation Performance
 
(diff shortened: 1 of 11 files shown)


groue:
That would be quite interesting! On top of that, I don't quite know how to introduce the new method, actually 😅 

The [current doc](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservation-usage) presents `tracking(_:)` as the one and only way to observe, and later introduces `trackingConstantRegion(_:)` as a [performance optimization](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservation-performance). I don't quite know where and how the new `tracking(region:fetch:)` (variadic) and `tracking(regions:fetch:)` (array of regions) methods should be introduced in the current [ValueObservation journey](https://github.com/groue/GRDB.swift/blob/master/README.md#valueobservation). If you have an idea, that'd be lovely!

freak4pc:
It could be some sub-section titled "Separating observation and query" or "Fine-grained ValueObservation". WDYT?

freak4pc:
_Here's a sketch_ @groue 

---


## Fine-grained ValueObservation

While the standard `ValueObservation.tracking(_:)` method lets you track changed to the result of a specific query and receive any changes to it, sometimes your use case might require more granular control. 

Consider a scenario where you'd like to get a specific Player's row, but only when their `score` column changes. You can use `tracking(region:fetch:)` or `tracking(regions:fetch:)` to do just that:

```swift
let observation = ValueObservation.tracking(
    // Define what region constitutes a "change"
    region: Player.select(Column("score")).filter(id: 1),
    // Define which query to execute upon such change
    fetch: { db in Player.fetchOne(id: 1) }
)
```

This overload of `ValueObservation` lets you entirely separate the **observation** from the **query** itself, providing utmost flexibility.

groue:
Yeah, that's good, @freak4pc 👍

We will not use "the query" because one can perform several database accesses as needed inside the fetch closure. And the "database region" word has to be used because it is a concept that is reused in some other places in the doc.

Also, let's skip the array `regions:` overload. It will be discovered by people who need it (I can't wait until Swift ships its splat operator and we just have to ship one single variadic method 🙄)

With those two changes, this could give:

---

## Fine-grained ValueObservation

While the standard `ValueObservation.tracking { db in ... }` method lets you track changes to a fetched value and receive any changes to it, sometimes your use case might require more granular control. 

Consider a scenario where you'd like to get a specific Player's row, but only when their `score` column changes. You can use `tracking(region:fetch:)` to do just that:

```swift
let observation = ValueObservation.tracking(
    // Define what region constitutes a "change"
    region: Player.select(Column("score")).filter(id: 1),
    // Define what to fetch upon such change
    fetch: { db in try Player.fetchOne(db, id: 1) }
)
```

This overload of `ValueObservation` lets you entirely separate the **observed region** from the **fetched value** itself, providing utmost flexibility.

freak4pc:
I like it! 
Do you want to add it to this PR ?

groue:
> Do you want to add it to this PR ?

Sure! Do you want to have your name on the commit?

freak4pc:
Neh, I really don't mind :) whatever helps and is easier is good :D

README.md
```diff
@@ -6683,17 +6683,22 @@ After `stopObservingDatabaseChangesUntilNextTransaction()`, the `databaseDidChan
...
-**[DatabaseRegion](https://groue.github.io/GRDB.swift/docs/5.14/Structs/DatabaseRegion.html) is a type that helps observing changes in the results of a database [request](#requests)**.
+A `DatabaseRegion` is a reunion of database tables, and combination of columns and rows (identified by their rowid):
 
-A request knows which database modifications can impact its results. It can communicate this information to [transaction observers](#transactionobserver-protocol) by the way of a DatabaseRegion.
+    |Table1 |   |Table2 |   |Table3 |   |Table4 |   |Table5 |
+    |-------|   |-------|   |-------|   |-------|   |-------|
+    |x|x|x|x|   |x| | | |   |x|x|x|x|   |x|x| |x|   | | | | |
+    |x|x|x|x|   |x| | | |   | | | | |   | | | | |   | |x| | |
+    |x|x|x|x|   |x| | | |   | | | | |   |x|x| |x|   | | | | |
+    |x|x|x|x|   |x| | | |   | | | | |   | | | | |   | | | | |
```
freak4pc, line 6693: 👏 👏 👏 👏 👏

groue: I can't write doc in a single pass 😉

freak4pc: I was just applauding the ASCII art capabilities ;-)

groue: 🤣 I ❤️ ASCII art! I keep https://en.wikipedia.org/wiki/Box-drawing_character under my pillow!

groue: Oh yes, a great app. I have it somewhere on my disk. I have big trouble remembering its name when I need it 😬🤓

PR #1121, opened 2021-12-29
GRDB: quote PROJECT_DIR and SRCROOT environment variables.

byohay:
Using those variables without quoting will yield unexpected results if
they contain spaces.


groue:
Thank you very much, @byohay 👍 !

groue:
Shipped in [v5.18.0](https://github.com/groue/GRDB.swift/tree/v5.18.0).

Issue #1138, opened 2022-01-13
Why isn't there a way to delete columns?


SF-Simon:
I read the part of the documentation on migration, there are methods such as deleting tables and adding columns, but I don't see the method of deleting columns.

Is it forgotten, or will there be any sequelae after this drop column?

Shouldn't it be like this?

            try db.alter(table: *) { t in
                t.drop(column: "name")
            }

groue:
Hello @SF-Simon,

The ability to drop columns was introduced quite recently, in [SQLite 3.35.0](https://www.sqlite.org/releaselog/3_35_0.html).

iOS 14.4, for example, ships with SQLite 3.23.3 ([reference](https://github.com/yapstudios/YapDatabase/wiki/SQLite-version-(bundled-with-OS))), and thus is unable to drop columns.

In order to enable such a Swift API, we need to know which exact versions of iOS, macOS, tvOS and watchOS come with SQLite >= 3.35.

A pull request would be quite appreciated!

groue:
I guess the initial question was answered, actually. I'm closing this issue, but the discussion can of course continue :-)

groue:
Fixed by #1144

PR #1144, opened 2022-01-16
Add drop column support

TimAEllis:
Following on from #1138, This adds support for drop column. 

I did some poking and prodding with downloaded simulator versions and it would seem the first iOS version that shipped with >3.35 was iOS 15, was able to check against iOS 15.0 and iOS 14.5 (and those were the only simulator ranges available to download to check, i.e. couldn't check things like 14.6, 14.7 etc).

### Pull Request Checklist

```
...
Executed 2279 tests, with 6 tests skipped and 0 failures (0 unexpected) in 87.346 (88.464) seconds
...
Executed 2279 tests, with 6 tests skipped and 0 failures (0 unexpected) in 93.532 (94.811) seconds
...
```

> The `make smokeTest` terminal command runs without failure.

_Sort of. I don't have simulators older than iOS 14.5 installed so it failed when trying to test against 13.7, otherwise it passed. If you are able to test against that specific version, that would be great as my tiny hard drive suffers with every simulator I download to it_

PS: Please be kind, I use this repository heaps and just wanted to help and not sure if I've missed anything in making this PR :)

### DIFF
--- a/GRDB/Core/StatementAuthorizer.swift
+++ b/GRDB/Core/StatementAuthorizer.swift
@@ -41,10 +41,11 @@ final class StatementCompilationAuthorizer: StatementAuthorizer {
         _ cString4: UnsafePointer<Int8>?)
     -> Int32
     {
+        // Uncomment when debugging
         // print("""
         //     StatementCompilationAuthorizer: \
         //     \(AuthorizerActionCode(rawValue: actionCode)) \
-        //     \([cString1, cString2, cString3, cString4].compactMap { $0.map(String.init) })
+        //     \([cString1, cString2, cString3, cString4].compactMap { $0.map(String.init) }.joined(separator: ", "))
         //     """)
         
         switch actionCode {
@@ -128,6 +129,21 @@ final class StatementCompilationAuthorizer: StatementAuthorizer {
             return SQLITE_OK
             
         case SQLITE_FUNCTION:
+            guard let cString2 = cString2 else { return SQLITE_OK }
+            
+            // SQLite does not report ALTER TABLE DROP COLUMN with the
+            // SQLITE_ALTER_TABLE action code. So we need to find another way
+            // to set the `invalidatesDatabaseSchemaCache` flag for such
+            // statement, and it is SQLITE_FUNCTION sqlite_drop_column.
+            //
+            // See <https://github.com/groue/GRDB.swift/pull/1144#issuecomment-1015155717>
+            // See <https://sqlite.org/forum/forumpost/bd47580ec2>
+            //
+            // TODO: remove when SQLite properly reports SQLITE_ALTER_TABLE
+            if strcmp(cString2, "sqlite_drop_column") == 0 {
+                invalidatesDatabaseSchemaCache = true
+            }
+            
             // Starting SQLite 3.19.0, `SELECT COUNT(*) FROM table` triggers
             // an authorization callback for SQLITE_READ with an empty
             // column: http://www.sqlite.org/changes.html#version_3_19_0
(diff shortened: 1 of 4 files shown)


groue:
Hello @TimAEllis, this is great, thank you very much!

In order to complete the pull request, we just need to also expose `drop(column:)` to custom SQLite builds and SQLCipher. Those must not use availability checks. See the [implementation](https://github.com/groue/GRDB.swift/blob/v5.18.0/GRDB/QueryInterface/Schema/TableDefinition.swift#L739-L772) of `rename(column:to:)` for an example.

groue:
> Sort of. I don't have simulators older than iOS 14.5 installed so it failed when trying to test against 13.7, otherwise it passed. If you are able to test against that specific version, that would be great as my tiny hard drive suffers with every simulator I download to it.

I will :-) It's high time I found a proper CI solution so that pull requests are properly checked. You did everything well 👍

TimAEllis:
> Hello @TimAEllis, this is great, thank you very much!
> 
> In order to complete the pull request, we just need to also expose `drop(column:)` to custom SQLite builds and SQLCipher. Those must not use availability checks. See the [implementation](https://github.com/groue/GRDB.swift/blob/v5.18.0/GRDB/QueryInterface/Schema/TableDefinition.swift#L739-L772) of `rename(column:to:)` for an example.

This should be done with daf44d9baad443f985d93684d7bfa22cd717a83b

groue:
Thank you @TimAEllis :-)

Reading the changes, I noticed that `TableDefinitionTests` would not always test that schema alterations would properly invalidate the internal GRDB schema cache. So I added tests for that, including one test for ALTER TABLE DROP COLUMN.

And... This test fails. Your changes are not responsible at all. This is an SQLite bug.

ALTER TABLE statements are supposed to call the [GRDB authorization callback](https://github.com/groue/GRDB.swift/blob/v5.18.0/GRDB/Core/StatementAuthorizer.swift#L36) with the [SQLITE_ALTER_TABLE](https://www.sqlite.org/c3ref/c_alter_table.html) code. This is how GRDB is able to [invalidate the schema cache](https://github.com/groue/GRDB.swift/blob/v5.18.0/GRDB/Core/StatementAuthorizer.swift#L65) whenever such a statement is run.

But ALTER TABLE DROP COLUMN does not trigger this behavior.

The consequence is that after a `drop(column: "foo")`, GRDB still thinks that `foo` is a column of the table, and this can create bugs later in the application.

Such bugs should be infrequent, since most apps perform schema changes early in their lifetime, before the GRDB schema cache had an opportunity to get warmed.

But I don't feel comfortable shipping a know bug. So the plans are:

1. Report the bug to the [SQLite forum](https://sqlite.org/forum/forummain)
2. Find a workaround

I can take care of (1). Do you want to take care of (2)?

groue:
> 1. Report the bug to the [SQLite forum](https://sqlite.org/forum/forummain)

Reported as https://sqlite.org/forum/forumpost/bd47580ec2

TimAEllis:
Hi @groue, this is starting to stretch my knowledge of SQLite (as little as it is, coming from MySQL). I mean, I'm happy to try and tackle this but I fear it won't be of acceptable quality 😅. I'm assuming I would have to hook into something like `TransactionObserver` but even that I literally only looked at for the first time today because of your reported message.

groue:
All right @TimAEllis, I'll complete the PR :-)

groue:
OK, looks good :-)

groue:
Shipped in [v5.19.0](https://github.com/groue/GRDB.swift/tree/v5.19.0) 🚀

groue:
> > 1. Report the bug to the [SQLite forum](https://sqlite.org/forum/forummain)
> 
> Reported as https://sqlite.org/forum/forumpost/bd47580ec2 (with the mention of a possible workaround).

[Fixed on trunk](https://sqlite.org/forum/forumpost/6537d092d020b387).

PR #1012, opened 2021-07-21
Add brackets to urls in doc comments to make them clickable

ZevEisenberg:
Adds angle brackets around all URLs in doc comments, which makes them clickable in Xcode's rendered documentation views. Also adds a lint rule to ensure that this style is maintained throughout the code base.

Note: the regex is slightly lazy in that does not look at the end of the URL. It's probably possible to make such a regex, but it might be ungainly enough that it's not worth it. Happy to take a crack at it if you think it's necessary.

Before:

A screenshot of some documentation rendered in Xcode's Quick Help Inspector. There is a URL, but it is not styled differently than the rest of the text.

After:

A screenshot of some documentation rendered in Xcode's Quick Help Inspector. There is a URL, and it is rendered in blue, indicating that it is clickable.

### Pull Request Checklist

- [x] This pull request is submitted against the `development` branch.
- [x] Inline documentation has been updated.
- [x] README.md or another dedicated guide has been updated.
- [x] Changes are tested.

PR #1151, opened 2022-01-19
Fix documentation for foreignKeysEnabled

### LINKED ISSUE #1150: Possibly wrong documentation for foreignKeysEnabled
I noticed the docs for `foreignKeysEnabled` say as follows:
image

I assumed this means that it only enables validity checks - confirming an object with a foreign key refers to something that exists, etc. 

I was very surprised though to see that setting it to `false` also prevent things like `onDelete: .cascade` from working - even if there is nothing wrong with the foreign keys.

More specifically:
1. Have two records with a `belongsTo` relation with a CASCADE on deletion and a foreign key. 
2. Turn on `foreignKeysEnabled` and make sure there are no warnings/issues.
3. Delete a "Parent" record, I correctly see the cascaded delete of the linked records.
4. Turn off `foreignKeysEnabled`
5. Delete a "Parent" record, all linked records are still there

I'm not sure if this is a bug, or simply a documentation issue but I suspect the latter. If it is, I'm happy to submit a PR with updated docs. 

Thank you 🙏

freak4pc:
Fixes #1150

### DIFF
--- a/GRDB/Core/Configuration.swift
+++ b/GRDB/Core/Configuration.swift
@@ -6,7 +6,8 @@ public struct Configuration {
     
     // MARK: - Misc options
     
-    /// If true, foreign key constraints are checked.
+    /// If true (the default), support for foreign keys is enabled.
+    /// See https://www.sqlite.org/foreignkeys.html for more information.
     ///
     /// Default: true
     public var foreignKeysEnabled = true
--- a/GRDB/Core/Configuration.swift
+++ b/GRDB/Core/Configuration.swift
@@ -7,7 +7,7 @@ public struct Configuration {
     // MARK: - Misc options
     
     /// If true (the default), support for foreign keys is enabled.
-    /// See https://www.sqlite.org/foreignkeys.html for more information.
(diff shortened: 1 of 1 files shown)


GRDB/Core/Configuration.swift
```diff
@@ -6,7 +6,8 @@ public struct Configuration {
     
     // MARK: - Misc options
     
-    /// If true, foreign key constraints are checked.
+    /// If true (the default), support for foreign keys is enabled.
+    /// See https://www.sqlite.org/foreignkeys.html for more information.
```
groue, line 10:
```suggestion
    /// See <https://www.sqlite.org/foreignkeys.html> for more information.
```

(Based on https://github.com/groue/GRDB.swift/pull/1012)

groue (changes requested):
🎉 Thank you Shai :-)

Please pardon the nit-picking

groue:
I think that's a quick win 🙂

freak4pc:
> 🎉 Thank you Shai :-)
> 
> Please pardon the nit-picking

I am pro nit-picking ;)
Thank you!

Issue #1157, opened 2022-01-26
Make queue public


Jasperav:
I have several places in my code base which should be executed only once at a time. I currently have an own queue and a DatabasePool. The DatabasePool is backed by a private DispatchQueue. Lot's of time, I need to do something with the database so I use pool.write and I know the code is executed serial. Sometimes I don't need the database so I use my own queue.

I now check if the code is on one of either queues:

```
func runningSerialCode() -> Bool {
    guard
        let queueName = String(
            cString: __dispatch_queue_get_label(nil),
            encoding: .utf8
        ) else {
        return false
    }

    return queueName == "GRDB.DatabasePool.writer" || queueName ==
        myDispatchQueue
}
```

However a race condition can still occur when code is executed in the GRDB queue and code is executed in my own queue and both call the same mutating function. Opening a write block everytime on the DatabasePool to just use the queue is a bit expensive I guess.

Is it possible to just expose that queue? That way I can remove my own queue. Or make a sync method on DatabasePool which can be used which doesn't open a transaction.

groue:
Hello @Jasperav,

I'm not sure I understand you question. `DatabasePool` uses several dispatch queues. That's how it serializes accesses to each individual SQLite connections, and yet support concurrent reads and writes.

Maybe you should better explain what you're after, so that we avoid talking past to each other in a nasty loop of [XY problem](https://xyproblem.info).

Jasperav:
Some more background on why I want this, I have a method which changes a variable. That method can be called while being in the `write` block of a `DatabasePool`. I also call that method when I am not in that `write` block, so I created my own `DispatchQueue` to eliminate race conditions. 

To make sure the method isn't called concurrently, I am stuck with lot's of `queue.sync` and `DispatchQueue` label checking. When I am in the `write` block, I should call `sync` on my own queue to not get race conditions. That looks a bit overkill for me, I thought `write` blocks can not run concurrently.

If I could reuse the queue from `SerializedDatabase` I wouldn't have that hassle anymore.

groue:
> That method can be called while being in the write block of a DatabasePool. I also call that method when I am not in that write block.

Did you challenge this practice before asking for a GRDB change?

groue:
> Did you challenge this practice before asking for a GRDB change?

Maybe I wasn't clear as well.

Considering a database write _can fail_, I'd be wary of modifying some shared state from within a write:

```swift
// Dubious. If the write eventually fails, `written` was set to true anyway.
[...]

PR #1161, opened 2022-01-29
Target DispatchQueue for non-read-only database connections

groue:
The new `Configuration.writeTargetQueue` property makes it possible to control the dispatch of non-read-only database connections.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 
 - **Fixed**: Fixed a bug of `including(all:)` for composite foreign keys and limited requests.
 - **New**: [#1160](https://github.com/groue/GRDB.swift/pull/1160) by [@groue](https://github.com/groue): Hide statement arguments by default.
+- **New**: [#1161](https://github.com/groue/GRDB.swift/pull/1160) by [@groue](https://github.com/groue): Target DispatchQueue for non-read-only database connections.
 - **Documentation Update**: The [Database Configuration](README.md#database-configuration) chapter explains how to opt in for public statement arguments in DEBUG builds.
 
 
--- a/GRDB/Core/Configuration.swift
+++ b/GRDB/Core/Configuration.swift
@@ -248,18 +248,28 @@ public struct Configuration {
     /// Default: .default
     public var qos: DispatchQoS = .default
     
-    /// The target queue for all database accesses.
+    /// A target queue for database accesses.
     ///
-    /// When you use a database pool, make sure the queue is concurrent. If
-    /// it is serial, no concurrent database access can happen, and you may
-    /// experience deadlocks.
+    /// Database connections which are not read-only will prefer
+    /// `writeTargetQueue` instead, if it is not nil.
+    ///
+    /// When you use a database pool, make sure this queue is concurrent. This
+    /// is because in a serial dispatch queue, no concurrent database access can
+    /// happen, and you may experience deadlocks.
     ///
     /// If the queue is nil, all database accesses happen in unspecified
-    /// dispatch queues whose quality of service and label are determined by the
-    /// `qos` and `label` Configuration properties.
+    /// dispatch queues whose quality of service is determined by the
+    /// `qos` property.
     ///
     /// Default: nil
     public var targetQueue: DispatchQueue? = nil
+    
+    /// The target queue for database connections which are not read-only.
+    ///
+    /// If this queue is nil, writer connections are controlled by `targetQueue`.
+    ///
+    /// Default: nil
+    public var writeTargetQueue: DispatchQueue? = nil
     
     // MARK: - Factory Configuration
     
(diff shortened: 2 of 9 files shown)


mallman:
Hi @groue. I found this in the release notes.

I'm intrigued by this capability, however I don't quite understand the motivation or utility, even after reading the discussion in #1157. After all, you didn't seem too keen on the OP's use case. Can you document or describe here when you believe a GRDB user might want to configure the target write queue?

If you don't want to, I won't complain. You already do a hell of a lot for this community. Cheers.

groue:
Hello @mallman, I'm not quite 100% sure myself 😅 Here was my reasoning:

#### Quality of Service

One use case of target queues is to specify a quality of service. In this context, specifying distinct target queues for reader/writer connections would allow providing a different quality of service for reader and writer connections. For example, you can make a DatabasePool with high priority reads, and low priority writes. Is it _very_ useful? I'm not sure, to be honest.

#### Protecting a Shared Mutable Resource

Another use case of target queues is to guarantee strict serialization (when the target queue is serial). I can imagine users who want to serialize some tasks with the database, in order to protect access to a shared resource:

```swift
var someSharedMutableResource = ...
let protectionQueue = DispatchQueue()
var config = Configuration()
config.targetQueue = protectionQueue
let dbQueue = DatabaseQueue(configuration: config)

protectionQueue.sync {
    // Safe access to someSharedMutableResource
}

try dbQueue.write { db in
    // Safe access to someSharedMutableResource
}
```

Admittedly, I never met this use case myself. And we know that this can create pain points:

```swift
// Deadlock
protectionQueue.sync {
    ... // 1
    try dbQueue.write { db in
        ... // 2
    }
    ... // 3
}
```

This is, basically, the problem with #1157. Until we learn more about the context, I don't know why the above deadlocking code could not be written instead:

```swift
try dbQueue.writeWithoutTransaction { db in
    ... // 1
    try db.inTransaction {
        ... // 2
        return .commit
    }
    ... // 3
}
```

I'm not saying this refactoring is _ideal_. I'm saying this refactoring solves a problem that #1157 presents as intractable. So I put aside the #1157 problem until more information is provided by @Jasperav.

But there is another problem to solve:

If some users want to use a target DispatchQueue in order to serialize some tasks with the database, then until this PR they could not use a DatabasePool. All writes and reads would be serialized, and this just ruins what DatabasePool is all about: concurrent reads and writes.

This is the reason for this PR: allow users to serialize some tasks with the write accesses (and gain thread-safe accesses to `someSharedMutableResource`), while preserving concurrent reads (where `someSharedMutableResource` is not used).

So the actual gains are hypothetical, I admit. Purely based on speculated user needs. I really wish we had more context about #1157.

Jasperav:
@groue I have created a simple reproduction path of why I created #1157. I now use the `Wrapper` in my application, although there is a huge penalty for performance because I check on every time I want to write to the database on which queue I am and switch to a queue in some cases.

The reproduction project can be found here: https://github.com/Jasperav/GRDBRaceCondition
The relevant class is shown here: https://github.com/Jasperav/GRDBRaceCondition/blob/master/GRDBRaceCondition/RaceCondition.swift

You can just run the application and see the logging.

In short: there is a `work` `func` in the `RaceCondition` class. That method should only be called from 1 thread at any given time . Sometimes I need to call that `func` while interacting with the database and sometimes when I am not. When I am not interacting with the database, I use the `DispatchQueue`.

I have multiple methods in multiple classes which can be called in sequence and no other thread should be calling these sensitive methods in the mean time. This is the reason why I don't want to use `Actor`s. 

I currently don't see any other way than the `Wrapper` way.

groue:
It is useful to forget about GRDB for a while, and step back a little bit.

**When one protects a resource with a serial DispatchQueue, aiming at [reentrancy](https://en.wikipedia.org/wiki/Reentrancy_(computing)) is fighting the framework.**

```swift
// Assuming protectedAccess uses `DispatchQueue.sync`...
performProtectedAccess {
    // Deadlock!
    performProtectedAccess { ... }
}
```

Yes, there are workarounds based on checks for the current queue. This can be done with `__dispatch_queue_get_label` as you do in your sample code, or with ["Contextual Data" (aka DispatchQueue specifics)](https://developer.apple.com/documentation/dispatch/dispatchqueue).

Such techniques are important to know, and sometimes useful.

But they remain fragile and can't protect all reentrant calls from deadlocks:

```swift
// No deadlock
performProtectedAccessCheckingForCurrentQueue {
    performProtectedAccessCheckingForCurrentQueue { ... }
}

performProtectedAccessCheckingForCurrentQueue {
    someOtherQueue.sync {
        // Deadlock!
        performProtectedAccessCheckingForCurrentQueue { ... }
    }
}
```

**There is no _perfect_ solution for DispatchQueue reentrancy. Designing an api that supports reentrancy is fundamentally risky.**

This is one of the reasons why GRDB database access methods are not reentrant, and warn the users as soon as possible:

```swift
try dbQueue.write { db in _
    // Fatal Error: Database methods are not reentrant.
    try dbQueue.write { db in _ ... }
}
```

Here the deadlock is replaced with a fatal error so that users "learn" that reentrancy is not supported. Reentrancy is detected with a DispatchQueue specific.

There remains unavoidable opportunities for deadlocks, though:

```swift
try dbQueue.write { db in _
    someOtherQueue.sync {
        // Deadlock!
        try dbQueue.write { db in _ ... }
    }
}
```

Now, @Jasperav, if you _want_ to design a reentrant api, then you _have_ to perform dispatch queue checks with `__dispatch_queue_get_label` or specifics. But this is not because of GRDB. This is because of DispatchQueue.

Jasperav:
@groue You are right, thanks for the explanation. It is, after some thinking, not something GRDB should support. Thanks for the tip of NSRecursiveLock, I didn't knew about that.

PR #1165, opened 2022-02-03
Add dynamic library to products to support dynamic linking

guidedways:
Add a new dynamic target to the products list. This is useful for apps with multiple extensions, where each extension would need to link with GRDB separately. The produced dynamic library can then be bundled once with the main target while other extensions load the shared 'framework'.

This also helps in situations where a conventional framework (in addition to the app's main target) used by an app may itself require linking with GRDB. Xcode will link GRDB statically to both the framework and the main app and will produce several warnings at runtime about duplicate symbols.

### Pull Request Checklist

- [✔️] CONTRIBUTING: You have read https://github.com/groue/GRDB.swift/blob/master/CONTRIBUTING.md
- [✔️] BRANCH: This pull request is submitted against the `development` branch.
- [✔️] DOCUMENTATION: Inline documentation has been updated.
- [✔️] DOCUMENTATION: README.md or another dedicated guide has been updated.
- [✔️] TESTS: Changes are tested.
- [✔️] TESTS: The `make smokeTest` terminal command runs without failure.


groue:
Hello @guidedways,

Thank you for your contribution!

The [documentation for `library(name:type:targets:)`](https://developer.apple.com/documentation/swift_packages/product/2878196-library) says:

> `type`: The optional type of the library that’s used to determine how to link to the library. Leave this parameter unspecified to let to let Xcode choose between static or dynamic linking (recommended). If you don’t support both linkage types, use .[Product.Library.LibraryType.static](https://developer.apple.com/documentation/swift_packages/product/library/librarytype/static) or [Product.Library.LibraryType.dynamic](https://developer.apple.com/documentation/swift_packages/product/library/librarytype/dynamic) for this parameter.

Should I understand that GRDB "doesn’t support both linkage types"? I also found [this message](https://forums.swift.org/t/specify-dependencies-should-be-built-as-dynamic-or-static/35257/2) which actually advises **against** providing two targets as you do.

So please provide a much stronger explanation for the need of your pull request.

Besides, since merging your commit would create a long-term commitment for the repository maintainers (making sure this new library works in the future, and answering support requests), please complete this pull request with documentation and tests:

- Update [the documentation for installing GRDB with SPM](https://github.com/groue/GRDB.swift/blob/master/README.md#swift-package-manager) so that users are guided towards the choice of the regular `GRDB` target, or the new `GRDBDynamic` target. The default choice, for people who do not know the difference, must be clearly identified. Provide a link to a reference document about dynamic SPM libraries. Keep the sentences short, and make sure that **all users** can understand you (even those who don't have a clue about the topic).
- Extend the [`test_install_SPM`](https://github.com/groue/GRDB.swift/blob/v5.20.0/Makefile#L382) Makefile task so that the new target is tested.

guidedways:
Xcode does not in fact allow you to link to a package dynamically. At least I am not aware of a way. See this:

https://forums.swift.org/t/how-to-link-a-swift-package-as-dynamic/32062

The argument I proposed is really it; we use GRDB with multiple targets and stand-alone frameworks within an app. Since xcode always statically links to GRDB, the compiled binaries each increase in size (static linking) and the console is riddled with runtime errors of "duplicate symbols found". This leads to another issue where two different versions of GRDB got linked (older in one of the frameworks and a newer in the main target).

There are two solutions:

* Create a wrapper module around the actual module you wish to consume as `dynamic`. This way the wrapper module links dynamically once while it is itself linked with GRDB statically.
* Update the module itself to support both dynamic and static linking.

guidedways:
Sure - I can look into updating the document and extend the tests :)

groue:
Thanks for the extra explanation 👍 OK, so basically Xcode integration of SPM sucks. I'll be glad to help 🙂

groue:
> Xcode integration of SPM sucks

It sucks a lot. What a nuisance. All "serious" packages double their libraries:

- https://github.com/RxSwiftCommunity/RxDataSources/blob/f671317335633fe9396fd3522dca1f5e64fc29c1/Package.swift
- https://github.com/apollographql/apollo-ios/blob/0c9aa54aeaed68d12987a6af39144d606d48c4a9/Package.swift
- https://github.com/Swinject/Swinject/blob/f10b6e9ebff440f985c43008f7c2d097639fcb81/Package.swift
- https://github.com/SnapKit/SnapKit/blob/466db39ee4ef32b0f7c8cd9f009aaf42f80beace/Package.swift
- https://github.com/ReactiveX/RxSwift/blob/7741a11cb4252c379c8e034508cb1b92b8b40427/Package.swift
- https://github.com/danielgindi/Charts/blob/85cfba96eb3492124d645ac0901b66f0cb266267/Package.swift

Gee, GRDB will be "serious" as well 😉

guidedways:
In short, yes 😬. Right now I have multiple wrapper modules around some of the `static` libraries we consume. It's difficult to maintain them because we need to remember to update each of these when the underlying library gets updated. Xcode is unable to infer if the library our dynamic wrapper is dependent upon also needs to be update (no amount of clicking on "Update Swift Packages" works). 

I'm sure once Xcode allows consumers to be able to choose between dynamic and static, all this will go away, but it has not for several years now. Below is a test to show this on Xcode 13.3 Beta 1. Notice how there's no way to do this and then it statically links the library instead of offering a way to link dynamically. We've got several extensions within the app that load dynamic libraries at runtime.

Screenshot 2022-02-03 at 12 51 25@2x

Screenshot 2022-02-03 at 12 51 38@2x

guidedways:
@groue done, to the best of my ability :) Let me know if there's anything else I need to do.

guidedways:
Just realized the first target should explicitly say `type: .static`

guidedways:
To elaborate, I see Apple's own example shows how a package produces both a `static` and a `dynamic` variant of the same library:

https://developer.apple.com/documentation/swift_packages/product

However given the `type is always either `.static` or `.dynamic`, I would argue since leaving that unspecified is considered `automatic`, a future update of Xcode may produce undesirable results where it picks either of the two types. Thus clearly marking one as static and the other dynamic serves the purpose of making the distinction obvious.

README.md
```diff
@@ -397,6 +397,8 @@ GRDB can be installed as a framework, or a static library.
 
 The [Swift Package Manager](https://swift.org/package-manager/) automates the distribution of Swift code. To use GRDB with SPM, add a dependency to `https://github.com/groue/GRDB.swift.git`
 
+> GRDB offers a `static` as well as `dynamic` library. Choose the default `GRDB` (static) library if you're not sure about the difference. Choose `GRDB-dynamic` if you are going to link it with multiple targets within your app and only wish to link to a shared, dynamic framework once. See [#How to link as dynamic](https://forums.swift.org/t/how-to-link-a-swift-package-as-dynamic/32062)
+>
```
groue, line 401: The default `GRDB` library is not static. It's the unfortunate behavior of Xcode that treats it as static only. But maybe Xcode will improve ;-) So I suggest instead:

```suggestion
GRDB two libraries, `GRDB` and `GRDB-dynamic`. Pick only one. When in doubt, prefer `GRDB`. The `GRDB-dynamic` library can reveal useful if you are going to link it with multiple targets within your app and only wish to link to a shared, dynamic framework once. See [#How to link as dynamic](https://forums.swift.org/t/how-to-link-a-swift-package-as-dynamic/32062)

```

guidedways: Yes that's correct, makes sense.

groue: Besides this wording change, the PR looks pretty good. I'll run the added test on my machine shortly.

I've read your conversation on https://forums.swift.org/t/specify-dependencies-should-be-built-as-dynamic-or-static/35257, and the resistance you find doesn't change my mind.

Quoting @SDGGiesbrecht there:

> If his IDE lacks a way to control the linking directly, the top‐level user should create his own package with a single .dynamic library, which in turn depends (statically) on all nodes that are common to the various executable components he plans to produce. Then each executable should depend solely on that library instead.

I can't disagree. But I also clearly see how unergonomic this is. And [I'm not the only one](https://github.com/groue/GRDB.swift/pull/1165#issuecomment-1028739487).

Your pull request is _pragmatic_, and _does not harm_. 👍

README.md
```diff
@@ -397,6 +397,8 @@ GRDB can be installed as a framework, or a static library.
 
 The [Swift Package Manager](https://swift.org/package-manager/) automates the distribution of Swift code. To use GRDB with SPM, add a dependency to `https://github.com/groue/GRDB.swift.git`
 
+GRDB two libraries, `GRDB` and `GRDB-dynamic`. Pick only one. When in doubt, prefer `GRDB`. The `GRDB-dynamic` library can reveal useful if you are going to link it with multiple targets within your app and only wish to link to a shared, dynamic framework once. See [#How to link as dynamic](https://forums.swift.org/t/how-to-link-a-swift-package-as-dynamic/32062)
```
groue, line 400: Excuse me, I suggested a typo 😅 

```suggestion
GRDB offers two libraries, `GRDB` and `GRDB-dynamic`. Pick only one. When in doubt, prefer `GRDB`. The `GRDB-dynamic` library can reveal useful if you are going to link it with multiple targets within your app and only wish to link to a shared, dynamic framework once. See [#How to link as dynamic](https://forums.swift.org/t/how-to-link-a-swift-package-as-dynamic/32062)
```

guidedways:
@groue yes - the need to link dynamically is real and it surfaces almost immediately when the complexity of what you're working on increases beyond a simple target. The PR should certainly not harm - if anything, it'll make it easier to drop the wrapper package I use for GRDB :)

README.md
guidedways: You know I saw that, and thought "I've just come out of debating with the dude who worked on SwiftPM and now I'm going to debate Gwendal? I'm sure he wanted it to read that way" 😅

groue:
Hello @guidedways,

The new DynamicProject project builds, and `make test_install_SPM_Dynamic_Project` passes.

But DynamicProject does not run (see error below). Is it normal?

```
dyld[76940]: Library not loaded: @rpath/GRDB-dynamic.framework/Versions/A/GRDB-dynamic
  Referenced from: /Users/groue/Library/Developer/Xcode/DerivedData/Dynamic-bpdoamabzjqspvbfajvkuobuajbs/Build/Products/Debug/Dynamic
  Reason: tried: '/Users/groue/Library/Developer/Xcode/DerivedData/Dynamic-bpdoamabzjqspvbfajvkuobuajbs/Build/Products/Debug/GRDB-dynamic.framework/Versions/A/GRDB-dynamic' (no such file), '/Users/groue/Library/Developer/Xcode/DerivedData/Dynamic-bpdoamabzjqspvbfajvkuobuajbs/Build/Products/Debug/PackageFrameworks/GRDB-dynamic.framework/Versions/A/GRDB-dynamic' (code signature in <FAC9DA34-413E-39CF-9A1E-DD5C74BC2616> '/Users/groue/Library/Developer/Xcode/DerivedData/Dynamic-bpdoamabzjqspvbfajvkuobuajbs/Build/Products/Debug/PackageFrameworks/GRDB-dynamic.framework/Versions/A/GRDB-dynamic' not valid for use in process: mapped file has no Team ID and is not a platform binary (signed with custom identity or adhoc?)), '/usr/lib/swift/GRDB-dynamic.framework/Versions/A/GRDB-dynamic' (no such file), '/usr/lib/swift/GRDB-dynamic.framework/Versions/A/GRDB-dynamic' (no such file), '/Library/Frameworks/GRDB-dynamic.framework/Versions/A/GRDB-dynamic' (no such file), '/System/Library/Frameworks/GRDB-dynamic.framework/Versions/A/GRDB-dynamic' (no such file)
```

guidedways:
@groue of course 🤦‍♂️ A terminal binary does not create a bundled app, which means there's no framework shipped along with it. Although compilation worked, running the app failed since it didn't know where to find `GRDB.framework`

I've replaced the example with a working iOS app that bundles the dynamic framework and embeds it correctly. I've verified it works.

guidedways:
Going to fix the test to point to the new project next

guidedways:
Done.

groue:
👍 Thank you @guidedways! I somewhat messed up the merging, but I think everything is cleared up now. I added one commit, https://github.com/groue/GRDB.swift/commit/9d4eb8a6dde2c66f9397884c930febd5c95b5def, which makes sure the new test uses an iOS simulator, and does not require any code signing.

groue:
Would you accept an invitation that grants you a push access to GRDB? I do intend to keep the lead of the project, but it is good for everyone that people of various levels of Xcode, Swift, and SQLite expertise can act on the repository when it becomes necessary. Of course, your free time would remain just as free as it was before!

groue:
(And I'll prepare a release shortly)

guidedways:
> 👍 Thank you @guidedways! I somewhat messed up the merging, but I think everything is cleared up now. I added one commit, [9d4eb8a](https://github.com/groue/GRDB.swift/commit/9d4eb8a6dde2c66f9397884c930febd5c95b5def), which makes sure the new test uses an iOS simulator, and does not require any code signing.

Thank YOU! That was the missing piece :)

guidedways:
> Would you accept an invitation that grants you a push access to GRDB? I do intend to keep the lead of the project, but it is good for everyone that people of various levels of Xcode, Swift, and SQLite expertise can act on the repository when it becomes necessary. Of course, your free time would remain just as free as it was before!

I'd be honored to :) Thank you for your consideration!

groue:
Done! Welcome :-)

guidedways:
Thank you 🥳

groue:
This PR has shipped in version 5.21.0! However, I can not presently perform the CocoaPods release (???)

```
% pod trunk push --allow-warnings GRDB.swift.podspec
Updating spec repo `trunk`
Validating podspec
 -> GRDB.swift (5.21.0)
    - ERROR | [iOS] unknown: Encountered an unknown error (809: unexpected token at 'objc[17593]: Class AMSupportURLConnectionDelegate is implemented in both /usr/lib/libamsupport.dylib (0x20724b678) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x105af42c8). One of the two will be used. Which one is undefined.
objc[17593]: Class AMSupportURLSession is implemented in both /usr/lib/libamsupport.dylib (0x20724b6c8) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x105af4318). One of the two will be used. Which one is undefined.
') during validation.

[!] The spec did not pass validation, due to 1 error.
```

Google searches for those errors are not conclusive so far.

guidedways:
The PR should not have affected CocoaPods. Could this be something else? 

https://apple.stackexchange.com/questions/407603/amsupporturl-is-implemented-in-both

guidedways:
Check this out: https://stackoverflow.com/a/69252338

And the answer it leads to suggests running it from within Xcode may help? Could it be some build setting affecting this?

groue:
Yes, I don't quite know what it is. I tried performing various reset techniques, reinstalled [Command line tools](https://developer.apple.com/download/all/?q=command), using Rosetta, reboot, etc. This is a new problem on my development machine, and maybe due to my recent installation of Xcode 13.3 beta. Still investigating 😅

groue:
> And the answer it leads to suggests running it from within Xcode may help? Could it be some build setting affecting this?

Do you know how to run a shell command from within Xcode?

guidedways:
Do you mean a post build command? You could do so via Build Phases > New Run Script Phase? 

I'm not sure the steps I need to perform to reproduce this on my machine, I could try and take a look

guidedways:
Perhaps this is more relevant and a known CocoaPods issue: https://github.com/CocoaPods/CocoaPods/issues/11179

guidedways:
You're probably on a M1 machine? A RoR app we manage was failing deployment recently on an M1 mac and we had to similarly get rid of the ffi dependency as it's not available on `arm64`. Someone on the thread seems to believe it was ffi causing it ¯\_(ツ)_/¯

groue:
I'm on a M1 machine indeed, on which I have already published several GRDB versions. My Ruby stack is supposedly ready for Apple Silicon. Something has changed, but I don't know quite what yet. I'll report what I can find.

```
% which ruby
/opt/homebrew/opt/ruby/bin/ruby
% ruby --version
ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [arm64-darwin21]
% gem update cocoapods
Nothing to update
% pod --version
1.11.2
```

A post build command would be _weird_. I admit I haven't tried yet.

groue:
I switched Terminal to Rosetta mode. Then the `pod trunk push` command worked fine.

👍 and 😕 at the same time...

guidedways:
So basically an arm64 issue 😞 oh well, at least there's nothing wrong with the library 🙂

groue:
Yeah :-)

PR #1186, opened 2022-03-20
DatabaseError provide error message as localizedFailureReason 

benrb:
Where `DatabaseError` conforms to `CustomNSError` it provides an `errorUserInfo` dictionary with a localised description key. This item in the dictionary is a string with the error code, message and sql in a single paragraph.

I think it would be good to provide another entry in the `errorUserInfo` dictionary with just the error message. 

This would be helpful as error codes and often SQL are not meant to be presented to the user and it is currently not possible to access just the message in places where `DatabaseError` has been type-erased to the `Error` type.

The userInfo key `NSLocalizedFailureReasonErrorKey` looks like it would be a good fit for this. 

Note:
I have modified the `testNSErrorBridging()` test to check that `localizedFailureReason` is not nil. It *is* able to be nil, but not for the specific error in that test. I thought it best just to check that the dictionary was being filled with a value where one is available, rather than check for exact strings which might change.

### Pull Request Checklist

- [n/a] DOCUMENTATION: Inline documentation has been updated.
- [n/a] DOCUMENTATION: README.md or another dedicated guide has been updated.

### DIFF
--- a/GRDB/Core/DatabaseError.swift
+++ b/GRDB/Core/DatabaseError.swift
@@ -506,5 +506,11 @@ extension DatabaseError {
     
     /// NSError bridging: the user-info dictionary.
     /// :nodoc:
-    public var errorUserInfo: [String: Any] { [NSLocalizedDescriptionKey: description] }
+    public var errorUserInfo: [String: Any] {
+        var userInfo = [NSLocalizedDescriptionKey: description]
+        if let message = message {
+            userInfo[NSLocalizedFailureReasonErrorKey] = message
+        }
+        return userInfo
+    }
 }
--- a/Tests/GRDBTests/DatabaseErrorTests.swift
+++ b/Tests/GRDBTests/DatabaseErrorTests.swift
@@ -226,6 +226,7 @@ class DatabaseErrorTests: GRDBTestCase {
                 XCTAssertEqual(DatabaseError.errorDomain, "GRDB.DatabaseError")
                 XCTAssertEqual(error.domain, DatabaseError.errorDomain)
                 XCTAssertEqual(error.code, 787)
+                XCTAssertNotNil(error.localizedFailureReason)
             }
         }
     }


groue:
Hello @benrb,

> This would be helpful as error codes and often SQL are not meant to be presented to the user and it is currently not possible to access just the message in places where DatabaseError has been type-erased to the Error type.

SQLite error messages are not for user consumption either:

- `FOREIGN KEY constraint failed`
- `UNIQUE constraint failed`
- `cannot start a transaction within a transaction`
- `attempt to write a readonly database`
- `file is not a database`

I'm not sure I understand what this pull request allows. Do you intend, in your app(s), to use the new `NSLocalizedFailureReasonErrorKey` key?

benrb:
> SQLite error messages are not for user consumption either

True, but some are quite readable.

The situation in which I would like to use this is for technically capable users to be able to see why a database action failed. They don't control the SQL being executed, but would at least be able to see what failed (Eg. `Index 'foo' already exists`) where relevant.

However this part of my app receives errors as an `NSError` object. I don't think there is currently a way to just extract just the error message from that - only the full contents of `description()`. Adding the message as a failure reason key would solve this.

---

Working around not having this isn't hard - just a couple of lines. If this isn't something you see others having a use for, it's certainly not a disaster for me :), but it's something I thought might be a small quality of life improvement.

groue:
> The situation in which I would like to use this is for technically capable users to be able to see why a database action failed. They don't control the SQL being executed, but would at least be able to see what failed (Eg. `Index 'foo' already exists`) where relevant.

I'm still not quite following, because the error message is not quite always enough to understand what's wrong: the SQL is useful in order to understand the "why".

- `UNIQUE constraint failed` - ❓ Hugh, what is the app doing?
- `UNIQUE constraint failed while executing INSERT INTO players ...` - 💡 Hey, aren't we supposed to check that each team has a single captain?

Database errors are targeted at developers, and are intended to help debugging. They are not suitable for application users. In this context, hiding information does not help.

> Working around not having this isn't hard - just a couple of lines. If this isn't something you see others having a use for, it's certainly not a disaster for me :), but it's something I thought might be a small quality of life improvement.

Maybe you have a *presentation* problem? You may want to display the error message on screen, and since the current error message can be long, you lack screen estate for displaying it?

You know, it is not that I'm against your pull request, but I'm still surprised you even need it.

groue:
> Or maybe you're developing an application that let users enter and execute SQL? In this context, SQL is known by the user, and there is no point repeating it in the error message?

Actually, this use case alone is enough to provide a good reason for your pull request 🙂

Sorry for having asked a lot of questions to you, and doubting the value of your contribution. Let's merge your pull request!

groue:
Done! Thank you @benrb!

groue:
Shipped in 5.22.0

benrb:
Thank you @groue, that's really great.

I am always happy to be questioned - it helps keep mistakes down!

Issue #1194, opened 2022-03-25
Deadlock after updating from 5.21.0 to 5.22.0


cosmer-work:
### What did you do?

Updated from 5.21.0 to 5.22.0

### What did you expect to happen?

Nothing.

### What happened instead?

Crash due to deadlock.

### Environment

**GRDB flavor(s): GRDB
**GRDB version: 5.22.0
**Installation method: SPM
**Xcode version: 13.3
**Swift version: 5.6
**Platform(s) running GRDB: iOS
**macOS version running Xcode: 12.2.1

cosmer-work:
Minimal reproducer.

```
struct Record: Codable, FetchableRecord, PersistableRecord {
    var id: Int64
}

var configuration = Configuration()
configuration.targetQueue = DispatchQueue(label: "crash.test", qos: .userInitiated)

let database = DatabaseQueue(configuration: configuration)

var migrator = DatabaseMigrator()
migrator.registerMigration("v1") { (db) in
    try db.create(table: Record.databaseTableName) { (t) in
        t.autoIncrementedPrimaryKey("id")
    }
}

try migrator.migrate(database)

let observation = ValueObservation.tracking { (db) in
    try Record.fetchCount(db)
}

var cancellable: AnyCancellable?
cancellable = observation.publisher(in: database, scheduling: .immediate)
    .map { _ in
        database.readPublisher { (db) in
            try Record.fetchCount(db)
        }
    }
    .switchToLatest()
    .sink { _ in

    } receiveValue: { (value) in
        print(value)
        cancellable?.cancel()
    }
```

cosmer-work:
Thread 1 Queue : GRDB.DatabaseQueue (serial)
#0	0x0000000105fba58b in __DISPATCH_WAIT_FOR_QUEUE__ ()
#1	0x0000000105fb9f3a in _dispatch_sync_f_slow ()
#2	0x00007fff5967b3c4 in implicit closure #2 in implicit closure #1 in OS_dispatch_queue.sync<τ_0_0>(execute:) ()
#3	0x00007fff5967ba92 in partial apply for implicit closure #2 in implicit closure #1 in OS_dispatch_queue.sync<τ_0_0>(execute:) ()
#4	0x00007fff5967ab1c in OS_dispatch_queue._syncHelper<τ_0_0>(fn:execute:rescue:) ()
#5	0x00007fff5967b06b in OS_dispatch_queue.sync<τ_0_0>(execute:) ()
[...]

PR #1025, opened 2021-08-05
Added locking around `reducer` to mitigate Release build crashes.

mattgallagher:
This PR is related to https://github.com/groue/GRDB.swift/issues/1026

This PR demonstrates the workarounds we're using to avoid Release build crashes on an app project. The crashes appear to be around access and maintenance of the `reducer` property on `ValueObserver`.

I suggest broader consideration of the thread safety model around `reducer` before merging this PR as-is. There might be a more elegant way to do this.

PR #1196, opened 2022-03-31
Fix a crash when an observation is quickly cancelled

### LINKED ISSUE #1195: Cancelling an observation quickly after starting it results in a crash
### What did you do?

Our application, in some circumstances, starts DB observations and quickly cancels them as the parameters change.

GRDB should tear down these short-lived observations successfully.

### What happened instead?

GRDB frequently crashes at `ValueWriteOnlyObserver.swift:269` which is `fatalError("can't start a cancelled or failed observation")`.

### Environment

**GRDB flavor(s):** Vanilla
**GRDB version:**  v5.22.1
**Installation method:** SPM
**Xcode version:** 13.3
**Swift version:** 5.6
**macOS version running Xcode:** N/A

### Demo Project

```
        configuration.targetQueue = DispatchQueue(label: "crash.test", qos: .userInitiated)

        let database = DatabaseQueue(configuration: configuration)

        var migrator = DatabaseMigrator()
        migrator.registerMigration("v1") { (db) in
            try db.create(table: Record.databaseTableName) { (t) in
                t.autoIncrementedPrimaryKey("id")
            }
        }

        try! migrator.migrate(database)

        let observation = ValueObservation.tracking { (db) in
            try Record.fetchCount(db)
        }

        var cancellable: AnyCancellable?
        cancellable = observation.publisher(in: database, scheduling: .async(onQueue: DispatchQueue.global()))
            .assertNoFailure()
            .sink { value in
                print(value)
            }
        cancellable?.cancel()
```

layoutSubviews:
This should fix #1195.

In `ValueWriteOnlyObserver.fetchAndStartObservation()`, it is possible to have `eventsOpt` be `nil` if the observation is cancelled right after the `isNotifying` check in `ValueWriteOnlyObserver.asyncStart()` as the lock is released in-between.

I'm making `ValueWriteOnlyObserver.fetchAndStartObservation()` return `nil` in case the observation was cancelled before it even got started (maybe a dedicated `Error` would be better?), and `ValueWriteOnlyObserver.asyncStart()` simply bails out in that case.
On the other hand, having `ValueWriteOnlyObserver.fetchAndStartObservation()` return `nil` in `ValueWriteOnlyObserver.syncStart()` is unexpected, so add a `fatalError` there.

### Pull Request Checklist

- [X] CONTRIBUTING: You have read https://github.com/groue/GRDB.swift/blob/master/CONTRIBUTING.md
- [X] BRANCH: This pull request is submitted against the `development` branch.
- [X] DOCUMENTATION: Inline documentation has been updated.
- [X] DOCUMENTATION: README.md or another dedicated guide has been updated.
- [X] TESTS: Changes are tested.
- [X] TESTS: The `make smokeTest` terminal command runs without failure.

### DIFF
--- a/GRDB/ValueObservation/ValueConcurrentObserver.swift
+++ b/GRDB/ValueObservation/ValueConcurrentObserver.swift
@@ -192,7 +192,8 @@ extension ValueConcurrentObserver {
             (self.notificationCallbacks, self.databaseAccess)
         }
         guard let notificationCallbacks = notificationCallbacksOpt, let databaseAccess = databaseAccessOpt else {
-            // Likely a GRDB bug
+            // Likely a GRDB bug: during a synchronous start, user is not
+            // able to cancel observation.
             fatalError("can't start a cancelled or failed observation")
         }
         
--- a/GRDB/ValueObservation/ValueWriteOnlyObserver.swift
+++ b/GRDB/ValueObservation/ValueWriteOnlyObserver.swift
@@ -198,7 +198,10 @@ extension ValueWriteOnlyObserver {
         // from a database access.
         try writer.unsafeReentrantWrite { db in
             // Fetch & Start observing the database
-            let fetchedValue = try fetchAndStartObservation(db)
+            guard let fetchedValue = try fetchAndStartObservation(db) else {
+                // Likely a GRDB bug
+                fatalError("can't start a cancelled or failed observation")
+            }
             
             // Reduce
             return reduceQueue.sync {
@@ -218,12 +221,11 @@ extension ValueWriteOnlyObserver {
         // Start from a write access, so that self can register as a
         // transaction observer.
         writer.asyncWriteWithoutTransaction { db in
-            let isNotifying = self.lock.synchronized { self.notificationCallbacks != nil }
-            guard isNotifying else { return /* Cancelled */ }
-            
             do {
                 // Fetch & Start observing the database
-                let fetchedValue = try self.fetchAndStartObservation(db)
+                guard let fetchedValue = try self.fetchAndStartObservation(db) else {
+                    return /* Cancelled */
+                }
                 
                 // Reduce
                 //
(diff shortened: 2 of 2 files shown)


groue:
Hello @layoutSubviews,

Your pull request fixes a logic bug: we were performing two cancellation checks in two distinct critical sections, but the second one would incorrectly handle cancellation:

1. Check if observation can notify events. If not, this means that observation was cancelled, and give up.
2. Check if observation can fetch. If not, this means that observation was cancelled, and assumes it is a GRDB bug due to the previous check.

Step 2 is incorrect, because between those two checks, those two critical sections, cancellation could happen.

Your pull request is the perfect fix for this logical bug: it performs a single cancellation check in a single critical section 👍

I attempted at performing the same fix on my own (pretending I did not read your code) and ended up with the exact same changes as yours :-)

---

Writing a regression test is very difficult, because:

- There is no synchronization technique that would allow to insert a cancellation between steps 1 and 2.
- The fix merges those steps in a single one (removing the ability to expose a private hook).

I'm thus comfortable with a pull request that is 100% based on reasoning, without any regression test.

Now maybe I'm just lacking testing skills! If anyone has an idea, I'm very curious :-)

groue:
Hello @layoutSubviews,

Your pull request fixes a logic bug: we were performing two cancellation checks in two distinct critical sections, but the second one would incorrectly handle cancellation:

1. Check if observation can notify events. If not, this means that observation was cancelled, and give up.
2. Check if observation can fetch. If not, this means that observation was cancelled, and assumes it is a GRDB bug due to the previous check.

Step 2 is incorrect, because between those two checks, those two critical sections, cancellation could happen.

Your pull request is the perfect fix for this logical bug: it performs a single cancellation check in a single critical section 👍

I attempted at performing the same fix on my own (pretending I did not read your code) and ended up with the exact same changes as yours :-)

---

Writing a regression test is very difficult, because:

- There is no synchronization technique that would allow to insert a cancellation between steps 1 and 2.
- The fix merges those steps in a single one (removing the ability to expose a private hook).

I'm thus comfortable with a pull request that is 100% based on reasoning, without any regression test.

Now maybe I'm just lacking testing skills! If anyone has an idea, I'm very curious :-)

groue:
I took the liberty to add a few comments, for clarity.

I took a risk with #1183 and its complete refactoring of ValueObservation internals 😅. Since 5.22.0, we had two issues: #1194 (fixed in 5.22.1), and now #1195, fixed by this pull request. The refactoring goal is somewhat achieved, because fixing those two bugs was relatively straightforward 🤓.

5.22.2 will ship shortly, and let's hope we'll reach the expected robustness level.

groue:
Merged!

@layoutSubviews, @mattgallagher, you both are the only people I know who dare committing changes to ValueObservation (here and #1025): your help is precious, and rare.

Would you accept an invitation that grants you a push access to GRDB? I do intend to keep the lead of the project, but it is good for everyone that people of various levels of expertise can improve the repository. Of course, your free time would remain just as free as it was before!

groue:
🚀 v5.22.2 has shipped

layoutSubviews:
Thanks @groue! I'd be happy to be granted push access to GRDB, and hopefully be able to contribute more meaningfully to this wonderful project :)

groue:
Invitation sent :-) Bienvenue !

mattgallagher:
Hi @groue, I'm happy to help, if needed. GRDB is a critical component of shipping apps that I maintain so I'd like to keep it healthy.

groue:
It's a pleasure to invite you both :-) If anything can help you feel closer to such a "critical component", please chime in any time, because I'll be happy to share.

Just as you know, this includes telling when you rely on undocumented behaviors, so that they are preserved until they become public in a way or an other.

Issue #1209, opened 2022-04-25
EXC_BAD_ACCESS: Exception 1, Code 1868215664, Subcode 8 > Stack overflow in __cxa_throw


xmanu:
### What did you do?

We recently released a new app using GRDB for the first time. We are really happy with GRDB so far.
Unfortunately we get weird stack overflow errors in our bug tracker that seem to contain no code from our codebase, but seem to lie solely in the GRDB source.

### What did you expect to happen?

The app should not crash.

### What happened instead?

The app crashes with EXC_BAD_ACCESS: Exception 1, Code 1868215664, Subcode 8 > Stack overflow in __cxa_throw.

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 5.19.0
**Installation method:** SPM
**Xcode version:** 13.3.1
**Swift version:** 5.6
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 12.3.1

I unfortunately cannot provide a simplified demo project as we cannot even reproduce the bug locally.

I can provide this stacktrace:
```
OS Version: iOS 15.4.1 (19E258)
Report Version: 104

Exception Type: EXC_BAD_ACCESS (SIGBUS)
Exception Codes: BUS_NOOP at 0x000000016f5ab570
Crashed Thread: 10

Application Specific Information:
Exception 1, Code 1868215664, Subcode 8 >
Stack overflow in __cxa_throw

Thread 10 Crashed:
0   <MyApp>                    0x200ff431c         SQLExpression.sql (SQLExpression.swift:871)
1   <MyApp>                    0x200ff5238         SQLExpression.sql (SQLExpression.swift:997)
2   <MyApp>                    0x20103728c         SQLQualifiedJoin.sql (SQLQueryGenerator.swift:780)
3   <MyApp>                    0x20102ab84         [inlined] sql (SQLQueryGenerator.swift:733)
4   <MyApp>                    0x20102ab84         SQLQueryGenerator.requestSQL (SQLQueryGenerator.swift:45)
5   <MyApp>                    0x201029c04         SQLQueryGenerator.makeStatement (SQLQueryGenerator.swift:112)
6   <MyApp>                    0x201028858         SQLQueryGenerator.makePreparedRequest (SQLQueryGenerator.swift:96)
7   <MyApp>                    0x200fdb980         QueryInterfaceRequest.makePreparedRequest (QueryInterfaceRequest.swift:52)
[...]

PR #1225, opened 2022-05-24
Prevent heavy DatabasePool concurrent reads from creating too many threads

groue:
This pull request limits the number of threads (prevents thread explosion) for applications that perform a lot of asynchronous reads with `DatabasePool.asyncRead` and related apis (`await dbPool.read`, `dbPool.readPublisher`, `dbPool.rx.read`, etc.)

Thanks to @xmanu for spotting in #1209 that the behavior of GRDB under heavy loads was lacking control :-)

Thanks to @khanlou for sharing how to use `DispatchSemaphore` in a sane way in [My GCD Handbook](https://khanlou.com/2016/04/the-GCD-handbook/) :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,10 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 
 ---
 
+## Next Release
+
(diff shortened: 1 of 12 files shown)


groue:
Merging despite one failing CI test (probably due to a too low timeout)

PR #1233, opened 2022-06-07
Fix warnings of redundant conformance constraint on Xcode 14.

GetToSet:
### Pull Request Checklist

After upgrading to latest Xcode 14 beta beta 1, I got some warnings about redundant conformance constraint.

It should be safe to remove these constraints.

### DIFF
--- a/GRDB/Fixit/GRDB-5.0.swift
+++ b/GRDB/Fixit/GRDB-5.0.swift
@@ -314,19 +314,14 @@ extension ValueObservation {
     { preconditionFailure() }
     
     @available(*, unavailable, message: "combine is no longer available. See the \"Migrating From GRDB 4 to GRDB 5\" guide.")
-    public func combine<
-        R1: _ValueReducer,
-        Combined>(
+    public func combine<R1, Combined>(
         _ other: ValueObservation<R1>,
         _ transform: @escaping (Reducer.Value, R1.Value) -> Combined)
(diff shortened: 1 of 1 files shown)


groue:
Thanks @GetToSet,

Those changes are compatible with Swift 5.3 (the GRDB5 requirement), so I think I'll merge very soon :-)

GetToSet:
> Thanks @GetToSet,
> 
> Those changes are compatible with Swift 5.3 (the GRDB5 requirement), so I think I'll merge very soon :-)

BTW, I found that `make smokeTest` fails locally, I'm currently not sure if it's related to Xcode 14, would you mind me trying to investigate and fix it (probably with another PR)?

groue:
> BTW, I found that `make smokeTest` fails locally, I'm currently not sure if it's related to Xcode 14

Yes it is due to Xcode 14. The Makefile [describes](https://github.com/groue/GRDB.swift/blob/b947065d253cc04b27a5cb0388c7f780ed6a3f28/Makefile#L53-L62) test destinations and Swift versions for each Xcode version. I'm not aware of a way to automatically grab those values, so this is a manual process!

> [...] would you mind me trying to investigate and fix it (probably with another PR)?

This would be much welcome! I don't remember why `MIN_TVOS_DESTINATION` is commented out since Xcode 13.1, so you can keep on ignoring it (it looks like we do not miss it, because no tvOS-related problem has been reported).

PR #1248, opened 2022-06-30
Avoid double notification of the initial value for ValueObservation on DatabasePool

groue:
The various SDKs that ship with Xcode 14 (currently in beta 2) expose the SQLite apis enabled by the [SQLITE_ENABLE_SNAPSHOT](https://www.sqlite.org/compile.html) compile-time option. For some reasons, those apis have been shipping for ages (iOS 10, macOS 10.12, watchOS 3.0, tvOS 10.0), but they were considered private by Apple. Now they're public!

That's all we needed for a cool improvement to `ValueObservation`. Until now, when started on a DatabasePool, an observation would always notify its initial value twice, even if the database was not changed. This annoyance is now (mostly) gone (starting Xcode 14+).

Explanation: `ValueObservation` performs its initial fetch on a reader connection, without waiting for an access to the writer connection (that's an expected benefit of using a DatabasePool). When the observation can finally access the writer connection, and start observing database transactions, it used to perform a second fetch, just in case some change would have been performed, unnoticed, since the initial fetch. Thanks to SQLITE_ENABLE_SNAPSHOT, GRDB can now compare the states of the database, between the initial fetch, and the initial access to the writer connection. When the database states are identical, the observation can avoid performing the second fetch.

Note that double initial notifications can still happen. For example, if your app tracks the player with id 1, and performs a change in any other table somewhere between the initial fetch and the beginning of transactions observation, then the initial player will be notified twice: GRDB has no way to tell if the change detected by SQLITE_ENABLE_SNAPSHOT is related to the tracked player or not (since transaction observation was not started yet), and has to assume the player was modified. Later on, after transaction observation has started, only changes to the tracked database region will trigger the observation, as one would expect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,11 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 
 ---
 
+## Next Version
+
+- **New**: [#1248](https://github.com/groue/GRDB.swift/pull/1248) by [@groue](https://github.com/groue): Avoid double notification of the initial value for ValueObservation on DatabasePool
+
+
 ## 5.25.0
 
 Released June 11, 2022 &bull; [diff](https://github.com/groue/GRDB.swift/compare/v5.24.1...v5.25.0)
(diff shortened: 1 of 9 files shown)


groue:
Xcode 14 RC can't build this feature on macOS and Mac Catalyst (#1276). The feature has been disabled on macOS and Mac Catalyst in [GRBD v5.26.1](https://github.com/groue/GRDB.swift/tree/v5.26.1). It will come back when Xcode is able to build it.

PR #1253, opened 2022-07-05
Make memory management optional.

george-signal:
Memory management may not be desirable for all
users of GRDB. `DatabasePool.releaseMemory`
creates a barrier which may cause normally fast
readers to block for a long time. If this is not a
desirable tradeoff, the new
`automaticMemoryManagement` flag in
`Configuration` allows you to disable this feature
of GRDB.

- [✓ ] CONTRIBUTING: You have read https://github.com/groue/GRDB.swift/blob/master/CONTRIBUTING.md
- [ ✓] BRANCH: This pull request is submitted against the `development` branch.
- [ ✓] DOCUMENTATION: Inline documentation has been updated.
- [ ✓] TESTS: Changes are tested.
- [ ✓] TESTS: The `make smokeTest` terminal command runs without failure.

### DIFF
--- a/GRDB/Core/Configuration.swift
+++ b/GRDB/Core/Configuration.swift
@@ -276,15 +276,11 @@ public struct Configuration {
     /// Default: nil
     public var writeTargetQueue: DispatchQueue? = nil
 
+#if os(iOS)
     /// Sets whether GRDB will release memory when entering the background or
     /// upon receiving a memory warning in iOS.
     ///
-    /// Enabling this setting may help keep iOS from terminating your app when
-    /// memory pressure becomes high. However, it can also cause database
-    /// readers to block longer than they normally would.
-    ///
     /// Default: true
-#if os(iOS)
     public var automaticMemoryManagement = true
 #endif
 
--- a/GRDB/Core/Configuration.swift
+++ b/GRDB/Core/Configuration.swift
@@ -275,7 +275,19 @@ public struct Configuration {
     ///
     /// Default: nil
     public var writeTargetQueue: DispatchQueue? = nil
-    
+
+    /// Sets whether GRDB will release memory when entering the background or
+    /// upon receiving a memory warning in iOS.
+    ///
+    /// Enabling this setting may help keep iOS from terminating your app when
+    /// memory pressure becomes high. However, it can also cause database
+    /// readers to block longer than they normally would.
+    ///
+    /// Default: true
+#if os(iOS)
+    public var automaticMemoryManagement = true
+#endif
+
     // MARK: - Factory Configuration
     
     /// Creates a factory configuration
--- a/GRDB/Core/Database.swift
+++ b/GRDB/Core/Database.swift
@@ -1184,7 +1184,8 @@ public final class Database: CustomStringConvertible, CustomDebugStringConvertib
     
     // MARK: - Memory Management
     
-    func releaseMemory() {
+    public func releaseMemory() {
+        SchedulingWatchdog.preconditionValidQueue(self)
         sqlite3_db_release_memory(sqliteConnection)
         schemaCache.clear()
         internalStatementCache.clear()
--- a/GRDB/Core/DatabasePool.swift
+++ b/GRDB/Core/DatabasePool.swift
@@ -103,7 +103,9 @@ public final class DatabasePool: DatabaseWriter {
         // Be a nice iOS citizen, and don't consume too much memory
         // See https://github.com/groue/GRDB.swift/#memory-management
         #if os(iOS)
-        setupMemoryManagement()
+        if configuration.automaticMemoryManagement {
+            setupMemoryManagement()
+        }
         #endif
     }
     
@@ -160,6 +162,17 @@ public final class DatabasePool: DatabaseWriter {
     fileprivate func forEachConnection(_ body: (Database) -> Void) {
         writer.sync(body)
         readerPool?.forEach { $0.sync(body) }
+    }
+
+    var numberOfReaders: Int {
+        guard let readerPool = readerPool else {
+            return 0
+        }
+        var count = 0
+        readerPool.forEach { _ in
+            count += 1
+        }
+        return count
     }
 }
 
--- a/GRDB/Core/DatabasePool.swift
+++ b/GRDB/Core/DatabasePool.swift
@@ -156,24 +156,6 @@ public final class DatabasePool: DatabaseWriter {
         
         return configuration
     }
-    
-    /// Blocks the current thread until all database connections have
-    /// executed the *body* block.
-    fileprivate func forEachConnection(_ body: (Database) -> Void) {
-        writer.sync(body)
-        readerPool?.forEach { $0.sync(body) }
-    }
-
-    var numberOfReaders: Int {
-        guard let readerPool = readerPool else {
-            return 0
-        }
-        var count = 0
-        readerPool.forEach { _ in
-            count += 1
-        }
-        return count
-    }
 }
 
 #if swift(>=5.6) && canImport(_Concurrency)
--- a/GRDB/Core/DatabasePool.swift
+++ b/GRDB/Core/DatabasePool.swift
@@ -180,7 +180,14 @@ extension DatabasePool {
         // Release writer memory
         writer.sync { $0.releaseMemory() }
         
-        // Release readers memory by closing all connections
+        // Release readers memory by closing all connections.
+        //
+        // We must use a barrier in order to guarantee that memory has been
+        // freed (reader connections closed) when the method exits, as
+        // documented.
+        //
+        // Without the barrier, connections would only close _eventually_ (after
+        // their eventual concurrent jobs have completed).
         readerPool?.barrier {
             readerPool?.removeAll()
         }
--- a/GRDB/Core/DatabasePool.swift
+++ b/GRDB/Core/DatabasePool.swift
@@ -185,16 +185,39 @@ extension DatabasePool {
     
     // MARK: - Memory management
     
-    /// Free as much memory as possible.
+    /// Frees as much memory as possible, by disposing non-essential memory from
+    /// the writer connection, and closing all reader connections.
     ///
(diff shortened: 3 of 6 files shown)


groue:
Hello @george-signal,

Thanks for your contribution! May I ask a few questions?

Isn't your app interested in freeing non-essential memory on `UIApplication.didReceiveMemoryWarningNotification` and `UIApplication.didEnterBackgroundNotification`? Assuming your pull request would be merged, wouldn't you want to handle those notifications, maybe in a different way, but with the same effect?

Your sentence "a barrier may cause normally fast readers to block for a long time" deserves an explanation - a practical one, not a theoretical one. The barrier indeed prevents reads performed after `releaseMemory()` from starting until all readers are closed, but this can only be a problem if the reader pool is already saturated with many enqueued reads that postpone the barrier. Even then, the barrier makes things slightly worse, but it does not completely changes the picture: reads are delayed anyway due to the pool saturation. In other words: did you try to lower database contention on the application side, instead of giving up memory management? On this topic, did you upgrade to version 5.24.1+? It ships with #1225, which improves the performance of frequent async reads.

Finally, assuming you are already past 5.24.1, you can't lower database contention on the application side, and you can't afford this barrier (which is unfortunate, but I won't judge you), why not attempting at preserving automatic memory management, but make it affordable for apps like yours?

george-signal:
> Assuming your pull request would be merged, wouldn't you want to handle `UIApplication.didReceiveMemoryWarningNotification` and `UIApplication.didEnterBackgroundNotification`, maybe in a different way, but with the same effect? Do you have any insight about the technique you would use?

I'm not an expert in GRDB and it's certainly possible that resources could be returned without needing the dispatch barrier. I see that the barrier was added in e495df0114c03ef23ae214d2d58585609e35bc38 but I don't know exactly why. I suppose it prevents adding more connections until the existing ones finish, but if the barrier completes quickly new readers could begin once again.

My team's experience with memory warnings is that they are hard to use well. An app often has very little time to react before getting killed. Guidance from a former Apple engineer was that within Apple, memory warnings are rarely used.

Anyway, I'm totally open to trying to release memory—it's just the barrier that makes automatic memory management an anti-feature for Signal—but I don't know quite how GRDB could do it.

> 
> Your sentence "a barrier may cause normally fast readers to block for a long time" deserves an explanation. The barrier indeed prevents reads performed after `releaseMemory()` from starting until the barrier has executed and all readers are closed, but this can only be a problem if the reader pool is already saturated with many enqueued reads that postpone the barrier. Even then, the barrier makes things slightly worse, but it does not completely change the picture: reads are delayed anyway due to the pool saturation. In other words: did you try to lower database contention on the application side, instead of giving up memory management? On this topic, did you upgrade to version 5.24.1+? It ships with #1225, which improves the performance of frequent concurrent async reads from a DatabasePool (if your app uses them).

The problem is not with pool saturation. We have 10 readers but only one or two are active in the crash reports I've seen. One of them is a very slow operation called "orphan data cleaning". In normal operation this works fine:

1. Orphan data cleaning begins on some private dispatch queue
2. Main thread reads your profile picture from database
3. Many seconds pass
4. Orphan data cleaning ends

The problem scenario is:

1. Orphan data cleaning begins on some private dispatch queue
2. Memory warning or backgrounding occurs
3. Main thread tries to read your profile picture from database. This blocks behind the barrier.
4. 10 seconds pass
5. iOS kills the app because the main thread is unresponsive

The root cause of the problem is that we should never do synchronous db operations on the main queue. But there are countless instances of this pattern and, at least for the foreseeable future, we're stuck with it. We just don't have the engineering resources to fix it.

We aren't on 5.24.1, but based on my reading of that code it wouldn't solve this particular problem. We don't have too many threads, just one really slow thread.

> When `didEnterBackgroundNotification` is posted, the application is not under the eyes of the user. How is the "long time" a problem?

Either the app is suspended or backgrounded. The main thread watchdog in iOS runs as long as it isn't suspended. The barrier can cause the main thread to hang when backgrounded (and subsequently foregrounded). iOS kills it after 10 seconds of non-suspension.

This is not always as harmful as the memory warning case since sometimes it'll crash while in the background. State restoration is imperfect so some unsaved user data or state can be lost. It burns battery and network to redo any pending work, such as fetching messages from the server. It introduces a delay when they return to the app as it starts up from a cold launch.

I believe the reason to release memory when getting backgrounded is to avoid getting killed when the system is low on memory while another app is active. Unfortunately the experience for our app in particular is more terminations rather than fewer because of a stuck main thread.

> Assuming your app already uses 5.24.1+, can't lower database contention, and can't afford this barrier (which is unfortunate, but I won't judge), shouldn't we attempt at preserving and improving automatic memory management, instead of disabling it?

Maybe! If it can be freed really quickly then there's value in reducing the frequency of getting killed. Based on my experience (at least in our app) it's not worth doing otherwise. For example, if you have cached data, it can often be freed quickly. That's an obvious win, and there may well be caches lurking within GRDB that I don't know about.

> Finally, can your original issue be rephrased as (pick one or many):
> 
> * You app depends on the order of operations performed on `didReceiveMemoryWarningNotification` and `didEnterBackgroundNotification`, and you lack control over it in the current state of GRDB?
> * You wish `releaseMemory()` were asynchronous?

I don't think either of this is a good description of the problem. Maybe a simpler rephrasing is "I wish releaseMemory didn't cause a priority inversion". 

I think "priority inversion" is fair because the main thread requires low-latency responsiveness while the orphan data cleaner is expected to be very slow. We end up with a high priority queue blocking on a low-priority queue.

groue:
Thanks for your detailed answer, @george-signal 👍 I'll come back very soon.

groue:
> The problem scenario is:
> 
> 1. Orphan data cleaning begins on some private dispatch queue
> 2. Memory warning or backgrounding occurs
> 3. Main thread tries to read your profile picture from database. This blocks behind the barrier.
> 4. 10 seconds pass
> 5. iOS kills the app because the main thread is unresponsive

This is a perfect description of the problem.

GRDB has and will always happily perform synchronous db operations on the main queue, because it just works so well in so many applications. Many db accesses are quite fast. DatabasePool precisely aims at being always able to read (unless the maximum number of reader connections has been reached).

So it is important that the library does not interfere with apps that perform database jobs on the main queue, and I now understand how this barrier is contrary to this goal.

With this interpretation, your pull request is a bug report.

I'm unable to remember quite well why the barrier was introduced. I suppose I found a nasty scenario. My current plan is to dig in my memory, find the problem that the barrier is supposed to solve, and find an alternative.

Meanwhile, are you stuck, or can you use your fork in your application?

george-signal:
For now I have removed DatabasePool from NSNotificationCenter so it doesn't know about memory warnings and backgrounding. I'm not stuck, but I can't say that I love this solution!

If you have a fix coming in the nearish future, we can wait until then to upgrade GRDB (I don't want to upgrade as long as my notification center hack is present to avoid unexpected surprises). If it looks like it'll be 6+ months I'd consider a fork.

Thank you so much for your help with this!

groue:
All right, we found an agreement 👍 

Meanwhile, I'm still interested in your PR! A few comments above, I made an hypothesis:

> Your app depends on the order of operations performed on `didReceiveMemoryWarningNotification` and `didEnterBackgroundNotification`, and you lack control over it in the current state of GRDB

Even if this was not your initial problem, this _can_ be a problem for some users. And your PR is a great opportunity to fix this.

So we'll merge your `automaticMemoryManagement` anyway. If you are ok with this, make sure I can push some commits to your branch: it will speed up the merge!

I'll address the barrier problem independently.

groue:
The `testDatabasePoolReleasesMemoryOnPressureEvent` as provided in the PR does not pass. Indeed it is not written in a way that guarantees that the test for the number of readers is performed after the reader pool has been emptied.

groue:
> The `testDatabasePoolReleasesMemoryOnPressureEvent` as provided in the PR does not pass. Indeed it is not written in a way that guarantees that the test for the number of readers is performed after the reader pool has been emptied.

Never mind, I'll take over it.

george-signal:
> > The `testDatabasePoolReleasesMemoryOnPressureEvent` as provided in the PR does not pass. Indeed it is not written in a way that guarantees that the test for the number of readers is performed after the reader pool has been emptied.
> 
> Never mind, I'll take over it.

Oops. Thank you so much for your help with this. I'm absolutely fine with landing this and dealing with the barrier later.

groue:
> My current plan is to [...] find the problem that the barrier is supposed to solve [...]

The barrier guarantees that non-essential memory is actually freed when the `releaseMemory()` method returns.

Before the barrier was introduced, the method would only make sure all readers would get closed after they stop being used (eventually). This did not match the documented behavior of the method.

I've added a regression test for the [undesired scenario](https://github.com/groue/GRDB.swift/pull/1253#issuecomment-1177166630) in caab403282db13cf1c9049c066afa65901007e68, and fixed it in d72b7109ad77c662b3b23e6e8a7f73888e0f9e27.

As a conclusion, @george-signal:

- `DatabasePool.releaseMemory` keeps on using a barrier, because that's how it can behave as documented. The undesired consequences of the barrier are now made clear in the documentation.

- A new method `DatabasePool.releaseMemoryEventually()` is introduced. It does not use any barrier, and won't prevent concurrent reads, but it can't tell when memory has actually been released.

- Automatic memory management no longer uses any barrier.

    This is only one exception: the barrier is still used on `UIApplication.didEnterBackgroundNotification`, for applications that do not support background tasks.

    But the barrier is never used on `UIApplication.didReceiveMemoryWarningNotification`.

- Users can opt out of automatic memory management with your new `Configuration.automaticMemoryManagement` flag. Your app is not supposed to be one of those that need to use this flag.

Your application is not supposed to opt out of automatic memory management, because:

- It does not match the intended use case, which is controlling the order of actions performed when notifications are posted.
- The initial reason for your desire to opt out is fixed: the barrier has been removed.

groue:
The PR has shipped in v5.26.0 :-) Thank you for your contribution, @george-signal 👍 Please report if anything does not behave as expected!

PR #1254, opened 2022-07-23
Accept ColumnExpressions for table column creation.

GetToSet:
This PR allows `ColumnExpression`s to be used for table column creation.

### Pull Request Checklist

### DIFF
--- a/GRDB/QueryInterface/Schema/TableDefinition.swift
+++ b/GRDB/QueryInterface/Schema/TableDefinition.swift
@@ -362,6 +362,25 @@ public final class TableDefinition {
         return column
     }
     
+    /// Appends a table column.
+    ///
+    ///     try db.create(table: "player") { t in
+    ///         t.column(Player.Columns.name, .text)
+    ///     }
+    ///
+    /// See <https://www.sqlite.org/lang_createtable.html#tablecoldef>
+    ///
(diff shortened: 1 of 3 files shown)


groue:
Hello @GetToSet,

Your pull request is well understood and perfectly formed, but I have to discard it. If you want to create table columns from column expressions, please feel free to do so, but embed these helper methods into your application.

The reason why I don't want migrations to support `ColumnExpression`, and even less to _document_ this feature, is because it is most often a very bad practice.

The role of each individual migration is to put the database schema in one specific state. The actions performed by one migration must remain the same, in all future versions of an application. The easiest way to make sure this rule is never broken in to never modify the code executed by a migration. And the easiest way to make sure that one never needs to modify a migration is to make it independent from the rest of the application code.

Unfortunately, your pull request fosters migrations that depends on record types. What happens, then, when you want to remove a column? You'll need to modify a migration. It is doable, of course. But it must not be fostered.

**A good migration is a migration that you never modify once it has shipped.**

Migrations and the rest of the application code do not live at the same "moment". By definition, migrations describe the past, and the rest of the application code describe the present. The present can depend on the past, but the opposite is not philosophically sound 😉 

Thank you very much for your contribution, even if I must discard it. You're always welcome 🙂

PR #1278, opened 2022-09-13
Immediate (synchronous) read & write Combine publishers

groue:
While using with [GRDBQuery](https://github.com/groue/GRDBQuery), it happens that some SwiftUI views don't want to stay in sync with the database content (thanks to database observation), but rather only fetch a unique database value that is not automatically updated. This is useful when building forms, for example: you feed the form with one state of the database, and then let the user deal with it.

Fetching a single database value with a Combine publisher is already possible with the `readPublisher` method. However, this publisher is asynchronous, meaning that any SwiftUI view that uses it has to deal with an absence of value, until it has been fetched. The Optional type is a good candidate for representing such situation: a nil value means that the fetch has not yet completed:

```swift
struct PlayerCountRequest: Queryable {
    static let defaultValue: Int? = nil

    func publisher(in dbQueue: DatabaseQueue) -> AnyPublisher<Int?, Error> {
        dbQueue
            .readPublisher { try Player.fetchCount($0) }
            .eraseToAnyPublisher()
    }
}

struct PlayerCountView: View {
    @Query(PlayerCountRequest()) var playerCount: Int?

    var body: some View {
        if let playerCount {
            Text("\(playerCount) players")
        } else {
            ProgressView()
        }
    }
}
```

This technique suffers from two problems:

- Some apps do not want the view to have a loading state. They want to opt in for synchronous, aka "immediate", database fetches - even if the database is accessed from the main thread. SQLite is very fast: this is not an unreasonable decision.
- Optional can't distinguish between 1. value was not fetched yet, 2. fetch was performed and value is missing, 3. value was fetched.

Consider:

```swift
struct PlayerRequest: Queryable {
    static let defaultValue: Player? = nil

    var playerId: Player.ID

    func publisher(in dbQueue: DatabaseQueue) -> AnyPublisher<Player?, Error> {
        dbQueue
            .readPublisher { try Player.fetchOne($0, id: playerId) }
            .eraseToAnyPublisher()
    }
}

struct PlayerView: View {
    @Query<PlayerRequest> var player: Player?

    init(playerId: Player.ID) {
        _player = Query(PlayerRequest(playerId: playerId))
    }

    var body: some View {
        if let player = player {
            Text(player.name)
        } else {
            // Player is not loaded yet, or missing 🤔
        }
    }
}
```

A workaround for the two problems is to build a synchronous publisher by hand:

```swift
struct PlayerRequest: Queryable {
    static let defaultValue: Player? = nil

    var playerId: Player.ID

    func publisher(in dbQueue: DatabaseQueue) -> AnyPublisher<Player?, Error> {
        Deferred {
            Result {
                try dbQueue.read { try Player.fetchOne($0, id: playerId) }
            }.publisher
        }.eraseToAnyPublisher()
    }
}

struct PlayerView: View {
    @Query<PlayerRequest> var player: Player?

    init(playerId: Player.ID) {
        _player = Query(PlayerRequest(playerId: playerId))
    }

    var body: some View {
        if let player = player {
            Text(player.name)
        } else {
            Text("Player does not exist") // 🙂
        }
    }
}
```

The `Deferred { Result { try ... }.publisher }` construct is totally OK, but requires a lot of Combine skills to come up with.

That's why this PR introduces the `immediateReadPublisher`, which is equivalent:

```swift
struct PlayerRequest: Queryable {
    static let defaultValue: Player? = nil

    var playerId: Player.ID

    func publisher(in dbQueue: DatabaseQueue) -> AnyPublisher<Player?, Error> {
        dbQueue
            .immediateReadPublisher { try Player.fetchOne($0, id: playerId) }
            .eraseToAnyPublisher()
    }
}
```

The goal of `immediateReadPublisher` is to streamline some documentations, as well as reuse the "immediate" word that has ValueObservation publishers perform an initial synchronous fetch as well:

```swift
ValueObservation
    .tracking { try Player.fetchOne($0, id: playerId) }
    .publisher(in: dbQueue, scheduling: .immediate)
    //                                  ~~~~~~~~~~
```

By symmetry with `immediateReadPublisher`, this PR defines `immediateWritePublisher` as well.

### DIFF
--- a/GRDB.xcodeproj/project.pbxproj
+++ b/GRDB.xcodeproj/project.pbxproj
@@ -198,6 +198,7 @@
 		567F45A81F888B2600030B59 /* TruncateOptimizationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 567F45A71F888B2600030B59 /* TruncateOptimizationTests.swift */; };
 		568068311EBBA26100EFB8AA /* SQLRequestTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 568068301EBBA26100EFB8AA /* SQLRequestTests.swift */; };
 		5682D721239582AA004B58C4 /* DatabaseSuspensionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5682D71A239582AA004B58C4 /* DatabaseSuspensionTests.swift */; };
+		56832C8F28D0571800E43AFA /* DatabaseWriterImmediateWritePublisherTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 56832C8E28D0571800E43AFA /* DatabaseWriterImmediateWritePublisherTests.swift */; };
 		56848973242DE36F002F9702 /* ValueObservationScheduler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 56848972242DE36F002F9702 /* ValueObservationScheduler.swift */; };
 		56894F752606576600268F4D /* FoundationDecimalTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 56894F742606576600268F4D /* FoundationDecimalTests.swift */; };
 		56894FB72606589700268F4D /* Decimal.swift in Sources */ = {isa = PBXBuildFile; fileRef = 56894F94260657D600268F4D /* Decimal.swift */; };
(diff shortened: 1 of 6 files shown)


groue:
The initial problem was solved by an updated GRDBQuery documentation.

Issue #1281, opened 2022-09-28
Xcode causes SIL deserialization failure in optimized release builds


anlaital:
### What did you do?

Compiling an iOS application with release configuration that has GRDB 6.0 as an SPM dependency.

GCC_OPTIMIZATION_LEVEL=-Os
SWIFT_OPTIMIZATION_LEVEL=-O

### What did you expect to happen?

The application should compiler and link successfully. There is a workaround for the issue which is to define `OTHER_SWIFT_FLAGS=-Xllvm -sil-disable-pass=performance-linker`.

### What happened instead?

```
---
4.	While evaluating request ExecuteSILPipelineRequest(Run pipelines { PrepareOptimizationPasses, EarlyModulePasses, HighLevel,Function+EarlyLoopOpt, HighLevel,Module+StackPromote, MidLevel,Function, ClosureSpecialize, LowLevel,Function, LateLoopOpt, SIL Debug Info Generator } on SIL for Project)
5.	While running pass #46361 SILFunctionTransform "Devirtualizer" on SILFunction "@$s4GRDB3RowCyxSicAA24DatabaseValueConvertibleRzAA015StatementColumnE0RzluigSiSg_Tg5".
 for getter for subscript(_:) (in module 'GRDB')
6.	While deserializing SIL function "$sSq4GRDBAA26StatementColumnConvertibleRzlE04fromB0_16atUncheckedIndexxSgSgs13OpaquePointerV_s5Int32VtFZ"
7.	While deserializing SIL function "sqlite3_column_type"
8.	*** DESERIALIZATION FAILURE ***
module 'GRDB' with full misc version '5.7(5.7)/Apple Swift version 5.7 (swiftlang-5.7.0.127.4 clang-1400.0.29.50)'
module 'SQLite3' was not loaded

Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
0  swift-frontend           0x0000000104c23e70 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 56
1  swift-frontend           0x0000000104c22e74 llvm::sys::RunSignalHandlers() + 112
2  swift-frontend           0x0000000104c244f4 SignalHandler(int) + 344
3  libsystem_platform.dylib 0x000000019fae34a4 _sigtramp + 56
4  libsystem_pthread.dylib  0x000000019facbee0 pthread_kill + 288
[...]

PR #1282, opened 2022-09-29
fix: Remove inlining of functions that cause optimizer to crash

anlaital:
I also created a [bug against Swift](https://github.com/apple/swift/issues/61350) about this. This enables us to remove the optimizer hacks and also use WMO in the app.


groue:
This closed PR is related to #1281.

Issue #1287, opened 2022-10-12
Strict table option only available in GRDBCUSTOMSQLITE


jcavar:
`.strict` table option is only available in custom SQLite builds. It seems that this was intentionally done here:

https://github.com/groue/GRDB.swift/pull/1133

I was wondering why is this only available in custom SQLite builds and not in OS build?

iOS 15.4.1 ships with 3.37.0, which supports it, so it might be behind availability check?

I am happy to make a PR, but just wanted to check for the approach before I do the work.

groue:
Hello @jcavar,

Thanks for noticing that STRICT tables are available on iOS 15.4.1 👍 A pull request would be great!

This should be enough to enable the feature:

```swift
#if GRDBCUSTOMSQLITE || GRDBCIPHER
    public static let strict = TableOptions(rawValue: 1 << 3)
#else
    @available(iOS 15.4, macOS 12.4, tvOS 15.4, watchOS 8.5, *) // SQLite 3.37+
    public static let strict = TableOptions(rawValue: 1 << 3)
#endif
```

(Yes, let's enable STRICT tables for SQLCipher as well).

And if you have a little more time, please take a look at tests and documentation, in case they would need an update!

Thank you very much :-)

groue:
Any news, @jcavar?

jcavar:
Hey @groue, sorry I've been little busy, but PR is now up.

groue:
Fixed in v6.1.0 :-)

Thank you very much @jcavar 👍

PR #1288, opened 2022-10-19
Add support for .strict table option in 3.37+

jcavar:
Resolves https://github.com/groue/GRDB.swift/issues/1287

### Pull Request Checklist

### DIFF
--- a/Tests/GRDBTests/TableDefinitionTests.swift
+++ b/Tests/GRDBTests/TableDefinitionTests.swift
@@ -76,6 +76,12 @@ class TableDefinitionTests: GRDBTestCase {
                 "e" ANY\
                 ) STRICT
                 """)
+            
+            do {
+                try db.execute(sql: "INSERT INTO test3 (id, a) VALUES (1, 'foo')")
+                XCTFail("Expected DatabaseError.SQLITE_CONSTRAINT_DATATYPE")
+            } catch DatabaseError.SQLITE_CONSTRAINT_DATATYPE {
+            }
         }
     }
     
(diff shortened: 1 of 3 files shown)


Tests/GRDBTests/TableDefinitionTests.swift
```diff
@@ -63,18 +66,37 @@ class TableDefinitionTests: GRDBTestCase {
...
                 "e" ANY\
-                ) STRICT, WITHOUT ROWID
+                ) STRICT
                 """)
-            
-            do {
-                try db.execute(sql: "INSERT INTO test3 (id, a) VALUES (1, 'foo')")
-                XCTFail("Expected DatabaseError.SQLITE_CONSTRAINT_DATATYPE")
-            } catch DatabaseError.SQLITE_CONSTRAINT_DATATYPE {
-            }
```
groue, line 73: This test looks like it tests SQLite itself, and has little value.

Instead, let's say it tests that GRDB properly handles the `.strict` option, by enabling the side-effect that is intended by the user: SQLite throws an error when the application tries to insert a value with the wrong data type.

From this point of view, this test has a value. Would you please restore it in both `testStrictTableCreationOptionCustomAndCipher` and `testStrictTableCreationOption`?

groue (changes requested):
Thank you @jcavar :-) This looks very good.

During the refactoring, one test was lost, and I think we should restore it.

groue:
I took the liberty to apply the necessary fixes. Also, the test is now skipped for SQLCipher 3.

groue:
OK, tests pass 👍 Thank you very much @jcavar, I'll prepare a release shortly.

groue (approved):

jcavar:
Hey @groue, sorry, I am little slower.

Indeed, I've probably deleted the most useful part of the strict tables test.

Thanks for fixing it for me and thank you for your great work on this library!

groue:
You're welcome, and happy GRDB :-)

PR #1296, opened 2022-11-23
Add @discardableResult to DatabaseWriter.write methods

StuClift:
There are some use-cases where we only perform a single expression in the write and don't care about the result's value.

```swift
    func deleteAllPlayers() async throws {
        try await dbWriter.write { db in
            _ = try Player.deleteAll(db)
        }
    }
```
In this case we explicitly discard the result with `_ =` despite `deleteAll` having the `@discardableResult` annotation. Removing the explicit discard will result in a warning from Xcode as the `write` method now returns the result of `deleteAll`.

Screenshot 2022-11-23 at 09 50 13

This pull request extends the use of `@discardableResult` to the `write` methods to reduce the need to explicitly discard the result of writes with a single expression.

#### Questions
- Is there a preferred order for these annotations?
- Should we update existing samples as part of this?

### Pull Request Checklist

### DIFF
--- a/GRDB/Core/DatabaseWriter.swift
+++ b/GRDB/Core/DatabaseWriter.swift
@@ -100,6 +100,7 @@ public protocol DatabaseWriter: DatabaseReader {
     /// - parameter updates: A closure which accesses the database.
     /// - throws: The error thrown by `updates`.
     @_disfavoredOverload // SR-15150 Async overloading in protocol implementation fails
+    @discardableResult
     func writeWithoutTransaction<T>(_ updates: (Database) throws -> T) rethrows -> T
     
     /// Executes database operations, and returns their result after they have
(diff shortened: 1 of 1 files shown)


groue:
Hello @StuClift,

Thank you very much for this pull request. I'm actually much aware of the issue, and frequently run into it :)

I'd rather not merge the PR, though. It would not be a good idea to allow database access methods to produce unused data without any warning.

Maybe the closure is returning the result of a costly database access or computation that should be removed? Bringing the developer's attention to this kind of issue is the purpose of this warning. This is a great feature of Swift.

Now, It is unfortunate that the Swift language can't forward the `@discardableResult` attribute of the last method call to the enclosing method. The same problem exists elsewhere, in the `DispatchQueue.sync` method, for example. Maybe this is an issue that can be reported to https://github.com/apple/swift?

So... Yes, `_ = try Player.deleteAll(db)` indeed.

StuClift:
Good shout, wasn't thinking of that angle as the result of the `write` is always unused or `Void` for us.

PR #892, opened 2020-12-17
Provide SQLCipher passphrase as Data

eburns-vmware:
Added the ability to specify passphrase for database encryption as a Data object.
Also added resetBytes call for Data objects created when using a string passphrase.
Updated the documentation with a section about passphrase datatypes.
Unit tests.

### Pull Request Checklist

- [x] This pull request is submitted against the `development` branch.
- [x] Inline documentation has been updated.
- [x] README.md or another dedicated guide has been updated.
- [x] Changes are tested.

PR #1328, opened 2023-02-09
Update README.md

ytti:
In SQLCipher passphrase as Data example, fix the use of Data instance

Update README.md for SQLCipher Data passphrase example. I'm not sure if the change is desired, but I felt the PR is more communicative than issue.

### Pull Request Checklist

- [x ] CONTRIBUTING: You have read https://github.com/groue/GRDB.swift/blob/master/CONTRIBUTING.md
- [ ] BRANCH: This pull request is submitted against the `development` branch.
- [x ] DOCUMENTATION: README.md or another dedicated guide has been updated.
- [ ] TESTS: Changes are tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5684,11 +5684,11 @@ For even better control over the lifetime of the passphrase in memory, use a Dat
 // RECOMMENDED: only load the passphrase when it is needed and reset its content immediately after use
 var config = Configuration()
 config.prepareDatabase { db in
-    let passphrase = try getPassphraseData() // Data
+    var passphraseData = try getPassphraseData() // Data
     defer {
-        passphrase.resetBytes(in: 0..<data.count)
+        passphraseData.resetBytes(in: 0..<passphraseData.count)
     }
-    try db.usePassphrase(passphrase)
+    try db.usePassphrase(passphraseData)
 }
 ```
 


ytti:
I do have some curiosity on if some type of randomBytes from secRandomCopyBytes would outperform randomBytes in terms of adversary inspecting memory after free? 

I have profound lack of understanding how any of this works, but would secRandomCopyBytes hide from adversary the length of the passphrase? Or would adversary be able to determine this regardless of the values of the bytes because the Data itself betrays it?

groue:
Hello @ytti.

Your pull request improves the documentation, thank you very much 👍 

> I do have some curiosity on if some type of randomBytes from secRandomCopyBytes would outperform resetBytes in terms of adversary inspecting memory after free?

You can run performance tests and use the Instruments app in order to answer your question.

Now I'm not sure this performance has any relevance. Applications are not supposed to frequently open database connections anyway. This is the first rule of [GRDB concurrency](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/concurrency).

ytti:
Mane thanks for this great project and the massive effort you've put into it, including stellar documentation. 

Again apologies for confusion. But I have further questions regarding utility of resetBytes(). And what exactly is the attack vector it might protect against. 

Data is pass-by-value, is it not? I am bit confused on what do we actually gain from resetBytes() in the call site? Isn't the `public func usePassphrase(_ passphrase: Data) throws` itself getting immutable copy of the Data object? So even if on the call site we resetBytes before it goes out of scope, do we not have the same problem on `usePassphrase` implementation? That is, can we really gain anything on the call site, without passing Data as reference?

groue:
Maybe we should ask the author of the original pull request https://github.com/groue/GRDB.swift/pull/892, @eburns-vmware.

ytti:
@eburns-vmware I have few separate questions

1. What are we protecting against with resetBytes(), I assume sanity of bytes after free after function goes out-of-scope
2. If so, are we not retaining the immutable pass-by-value copy? Would inout help?
3. If resetBytes() does something, would randomBytes() add more, could it obscure length? 
4. Even if we can make Data totally safe after free, there isn't way in SwiftUI to gain user string as Data, you have to use SecureField? How do Data.resetBytes() and String.removeAll() security guarantees differ?

ytti:
If I read this right, sqlite3_key does not modify the unsafe immutable bytes, and we have before and after call same bytes in memory. I interpret this as it doesn't matter what we do on the call-site. If we need `resetBytes()` to improve hygiene, it appears we would need some solution on the library side as well.
Näyttökuva 2023-2-10 kello 13 22 12

However I'm lacking tremendous amount of context to determine if any of this is needed.
1. What does Swift do to memory before free? Maybe it already guarantees this for us.
2. Would inout help at all, if it would, would be it be implementation detail we now observe or guaranteed behaviour

ytti:
I remain with tremendous lack of context and have largely exhausted my ability to troubleshoot this, I currently believe that 

1. Swift types as they currently are designed, make no contractual promises to developers about any type offering any type of memory hygiene, and that neither Data.resetBytes or String.removeAll are contractually saying anything about reusing the same bytes for post-call representation.  Such behaviour might be observed, but  that would be implementation detail and some future optimisation might change the behaviour.

2. Developer could write their own String or Data implementation which makes hard guarantee about this, runtime doesn't appear to preclude such implementation. 

3. Casual googling suggests this hygiene problem has been raised to Apple and Apple is aware that practical developer has no way to guarantee such thing, and appears to consider this a problem for the target system, which may be fair separation of concerns.  

But I may be confused. If I am right, it may be unfair to suggest to users that this is a problem that they should/could worry about. And it might be useful, if some definitive answer exists, to explain to them why this is currently not their concern.

ytti:
* https://mjtsai.com/blog/2022/09/20/zeroing-freed-memory/
* https://developer.apple.com/documentation/macos-release-notes/macos-13-release-notes

I read this as iOS 16.2 and macOS 13.1 zero memory out before free. And as Swift is ARC instead of tracing, free should occur deterministically after we go out of scope. 

Therefore I believe

1. Data.resetBytes does not guarantee we are resetting bytes from the original storageObject1, instead of allocating new storageObject2 and letting the old storageObject1 to go to reference 0 and be free'd

2. Such guarantee is not needed and may be antipattern for developer to be concerned of, as chance of confusion and mistake is very high. So it may be best to trust the system to handle it or for system to be broken and fixed by someone else.

eburns-vmware:
Hey - thanks for looking into this and raising the question.  The thinking behind my suggested implementation is:

> var passphraseData = try getPassphraseData() // Data

We now have a Data object with sensitive passphrase data in memory.  This is a necessary evil for the moment, but once we've unlocked the data, it becomes an unnecessary evil.  (Thanks for the var fix...)

>     defer { passphraseData.resetBytes(in: 0..<passphraseData.count) }

As we exit the block, and before the block goes out of scope (and the object), reset the memory backing the passphraseData object.  My understanding is this sets the memory content to 0 for the specified range:

https://developer.apple.com/documentation/foundation/data/1780301-resetbytes
**resetBytes(in:)**
_Sets a region of the data buffer to 0._

>     try db.usePassphrase(passphraseData)

Attempt to unlock the database using the passphraseData.  Regardless of success or failure, post try the block will terminate, executing the defer block.

- - -

So why would I recommend doing it this way?  It is just a defensive approach to [memory safety](https://en.wikipedia.org/wiki/Memory_safety).  In the case of an overrun, the secret could (extremely unlikely scenario that involves both a skilled hacker and a tremendous amount of luck) be captured and applied.  Even though this is not common or likely, it is just a good stance to leave nothing there to be discovered, should any exposure occur.

I don't read Apple's documentation and guaranteeing a zero out upon release.  If that happens, then that is great.  But this is a relatively cheap guarantee that it gets zeroed and then freed.

Of your two beliefs in the last comment:

1. By reading the docs for _resetBytes(in:)_, it is instance method marked as mutating so it seems to me that it does affect the memory space backing the object being operated on.

2. I think this is security by intent, not by hoping that Apple is doing it for you.  Every dev is free to choose their own security stance, but this is definitely the approach I would recommend.

ytti:
resetBytes certainly resets the bytes of the object you see after the call, but as to my reading of Swift, there is no guarantees you are looking at same storage bytes before and after call. You might just as well get entirely new location of memory where bytes are now zero, with no promise on what happened to the old bytes.

eburns-vmware:
I set this up in a playground and ran it through its paces over and over, and the behavior I am seeing is the behavior that I'm expecting from reading the docs.

> var string = "String"
guard var data = string.data(using: .utf8) else { exit(0) }
>
> print(NSString(format: "%p", address(o: &data)))
print("Data: \(data.map { String(format: "%02hhx", $0) }.joined())")
>
>data.resetBytes(in: 0..<data.count)
>
>print(NSString(format: "%p", address(o: &data)))
>print("Data: \(data.map { String(format: "%02hhx", $0) }.joined())")

Output:

>0x16b202d38
Data: 537472696e67
0x16b202d38
Data: 000000000000

Is this a guarantee that it will always use the same memory?  I don't think it is a guarantee, but it certainly seems to indicate that is the case.  Instantiating other value types between the instantiation of data and the call to resetBytes does not seem to have any effect - the data object always points to the same address after calling resetBytes.

I know that Data is a value-type, but Apple appears to have taken the nature of resetBytes from ObjC and applied it here in Swift as well.

And yes, we cannot control what happens inside the library.  We either cross our fingers, do extensive memory testing or avoid the library if we cannot trust it.  But what we can do is the best possible practices on our side that we do have control over.

I believe this approach has merit.  You do not - which is fine.  Feel free to make an argument against it in the README.

ytti:
https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/Data.swift#L378-L392 calls https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/Data.swift#L183-L282 which I read as calling https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/Data.swift#L176-L181 to determine if or not we allocate new bytes. If `guard isExternallyOwned || newLength > _capacity else { return }` is true, we seem to use same bytes, if that is false, we use new bytes.

So my reading (low confidence) is, given that the new bytes fit and that it is internally owned we are today using same bytes in memory. 
This does not appear contractual, as we could imagine ensureUniqueBufferReference having some optimisation made which changes this behaviour. I'm also not sure if further along the compilation this behaviour is kept, or if some optimisation further along the compilation may change this behaviour for performance.

But I think we can ignore this if it's contractual or not discussion for now, as we won't make progress. I don't have high confidence, but I doubt if guarantees exist. However point not discussed sufficiently is the library side of the call.

If we assume for sake of argument that it's contractual , and we correctly resetBytes at the call site as shown in the PR, we're still passing immutable copy to the library, which is not resetting the bytes, so we are still keeping unreset bytes of the immutable copy in the memory, and relying on iOS/macOS to zero those on free.

Issue #1256, opened 2022-08-04
GRDB 5.26.0 cost more time on value Observation


karthuszY:
### What did you do?

I update GRDB from 5.24.0 to 5.26.0. And then I feel my App got not smooth.

### Environment

**GRDB flavor(s):**   GRDB.swift/SQLCiphe
**GRDB version:** 5.26.0
**Installation method:** `pod 'GRDB.swift/SQLCipher', :git => 'git@github.com:groue/GRDB.swift.git'`
**Xcode version:** Version 13.0 (13A233)
**Swift version:** Swift 5
**Platform(s) running GRDB:** iOS
**macOS version running Xcode:** 12.0.1 (21A559)

### Related
GRDB **5.24.0** open cost：

![image](https://user-images.githubusercontent.com/13864988/182755224-59d2648e-45f5-43ba-8aa5-64d7e6652966.png)

GRDB **5.26.0** open cost：

![image](https://user-images.githubusercontent.com/13864988/182755349-671e4dc5-ce5f-4baf-9053-45f5bc583fb8.png)

groue:
Hello @karthuszY,

Do you happen to use a DatabaseQueue, or DatabasePool?

Also, I won't be able to look at the issue until mid of August. Meanwhile, it would be useful if you would extend the issue with enough detail so that it is possible to **reproduce** the problem.

karthuszY:
I use DatabasePool.

Here is the code:

```
       appInfoCancelable?.cancel()
        chainInfoCancelable?.cancel()
        
        let appInfoObserveation = ValueObservation.tracking(AppInfoModel
                                                                .filter(AppInfoModel.Columns.id == appId)
                                                                .fetchOne)
        appInfoCancelable = appInfoObserveation.start(in: db,
                                                      scheduling: .immediate) { error in
            
        } onChange: { [weak self] appInfo in
       
        }
        
        let chainInfoObserveation = ValueObservation.tracking(ChainInfoModel.fetchAll(_:))
        chainInfoCancelable = chainInfoObserveation.start(in: db,
                                                          scheduling: .immediate,
                                                          onError: { error in
            
        }, onChange: { [weak self] _ in
        })
```

groue:
Hello @karthuszY,

I can't reproduce any significant performance difference. Is it possible for you to setup a minimal sample project?

groue:
No news is good news! Feel free to reopen when it is possible to provide some more detailed feedback :)

[...]

PR #1350, opened 2023-03-16
DatabasePool won't close read-only connections if requested, and ValueObservation no longer opens a new database connection when it starts

groue:
This PR completes #1248, and aims at solving #1256:

- Enhances performances of `ValueObservation` on `DatabasePool`, because it no longer opens a new database connection when it starts.
- Allows applications to keep read-only connections open.

    `Configuration` has gained a `persistentReadOnlyConnections` property. The default is false: `DatabasePool` manages read-only connections in order to spare memory. When true, `DatabasePool` keeps read-only connections open.
    
    ```swift
    // Keep read-only connections open
    var config = Configuration()
    config.persistentReadOnlyConnections = true
    let dbPool = try DatabasePool(path: "...", configuration: config)
    ```
    
    When profiling your application reveals that a lot of time is spent opening a new SQLite connection at the wrong place:
    
    - Set `persistentReadOnlyConnections` to true.
    - Make sure the needed read-only connection was opened beforehand. A plain `try dbPool.read { _ in }`, for example, creates a read-only connection if no one is available yet. There's no built-in api for creating multiple read-only connections (open a feature request if you need this).
    - Check your eventual use of `invalidateReadOnlyConnections()`, because this method closes read-only connections regardless of the `persistentReadOnlyConnections` configuration.

---

Since #1248, ValueObservation on DatabasePool creates a new database connection when it starts, on the thread that starts the observation. #1256 has revealed that opening a connection can take a long time, with unfortunate consequences for the main thread :-/

In this PR, ValueObservation on DatabasePool just uses one of the available read-only connections:

- Starting a synchronous `.immediate` observation still blocks the main thread until the initial value is fetched, but no time is spent opening a new connection if there is an available read-only connection. If there is no available read-only connection, the main thread is blocked until one is opened, or becomes available. That's where `persistentReadOnlyConnections` is useful: it helps finding an available read-only connection, as long as the app has already performed reads before the observation starts.
- Starting an asynchronous observation schedules the fetch of the initial value and returns immediately. Again, `persistentReadOnlyConnections` is useful, because starting an observation asynchronously also needs an available read-only connection.

When WAL snapshots are available, a long-running transaction is kept opened on the read-only connection, until we can acquire a write access, compare database versions, perform a fetch+notification if and only if the database was modified, and start observing the write transactions.

When WAL snapshots are not available (Xcode 14.0 and macOS, and some SQLCipher/SQLite builds), a double notification of the initial value is always performed, because we can't prove that the database was not changed between the initial fetch and the write access (we need WAL snapshots for that).

### DIFF
--- a/GRDB/Core/SerializedDatabase.swift
+++ b/GRDB/Core/SerializedDatabase.swift
@@ -222,7 +222,7 @@ final class SerializedDatabase {
         }
     }
     
-    // Schedules database operations for execution, and returns immediately.
+    /// Schedules database operations for execution, and returns immediately.
     func async(_ block: @escaping (Database) -> Void) {
         queue.async {
             block(self.db)
--- a/GRDB/Core/SerializedDatabase.swift
+++ b/GRDB/Core/SerializedDatabase.swift
@@ -9,6 +9,8 @@ final class SerializedDatabase {
     var configuration: Configuration { db.configuration }
     
     /// If true, overrides `configuration.allowsUnsafeTransactions`.
+    ///
+    /// See `WALSnapshotTransaction`.
     var allowsUnsafeTransactions = false
     
     /// The path to the database file
--- a/GRDB/Core/SerializedDatabase.swift
+++ b/GRDB/Core/SerializedDatabase.swift
@@ -140,6 +140,21 @@ final class SerializedDatabase {
         }
     }
     
+    /// Executes database operations, returns their result after they have
+    /// finished executing, and allows or forbids long-lived transactions.
+    ///
+    /// This method is not reentrant.
+    ///
+    /// - parameter allowingLongLivedTransaction: When true, the
+    ///   ``Configuration/allowsUnsafeTransactions`` configuration flag is
+    ///   ignored until this method is called again with false.
+    func reentrantSync<T>(allowingLongLivedTransaction: Bool, _ body: (Database) throws -> T) rethrows -> T {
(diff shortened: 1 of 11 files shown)


GRDB/Core/SerializedDatabase.swift
```diff
@@ -122,8 +140,23 @@ final class SerializedDatabase {
         }
     }
     
-    /// Synchronously executes a block the serialized dispatch queue, and
-    /// returns its result.
+    /// Executes database operations, returns their result after they have
+    /// finished executing, and allows or forbids long-lived transactions.
+    ///
+    /// This method is not reentrant.
```
rpassis, line 146:
```suggestion
    /// This method is reentrant.
```

rpassis (commented):
Tested this with `SQL_ENABLE_SNAPSHOT` turned on and off. 

In both scenarios the changes here drastically reduced the thread blocking we were experiencing with `ValueObservation` and `SQLCipher` when opening a database connection.

Turning the `persistentReadOnlyConnections` flag to `true` didn't seem to have a noticeable effect however it seems like it would be an effective solution to minimise opening of database connections for cases like ours where this operation can be process intensive.

I've left one minor comment on the documentation where you seem to be using a reentrant method but the documentation says otherwise.

Thanks for working on this fix so quickly, our team truly appreciates your efforts 👊👍💙

groue:
v6.10.0 has shipped

Issue #1362, opened 2023-04-12
Deadlock in Pool.swift in v6.10.1


mattgallagher:
Since updating to 6.10, I've been seeing deadlocks with threads stuck at `itemsSemaphore.wait()` at Pool.swift line 80. The stack traces themselves aren't hugely informative but I have a suspicion that something in v6.10 has lead to unbalanced calls to `itemsSemaphore.signal()` – possibly https://github.com/groue/GRDB.swift/pull/1350.

![Screenshot 2023-04-12 at 16 47 42](https://user-images.githubusercontent.com/28080/231374170-eba9413d-db04-4142-a9ac-e2be429cb51d.png)

groue:
Hello @mattgallagher,

Yes, #1350 is certainly the origin of the regression, I'm sorry 😕. Is it possible for you to downgrade to v6.9.2 until the deadlock is understood and fixed?

A possible explanation is that the internal `WALSnapshotTransaction` class "forgets" to call its `release` closure - because the class contains a logic bug, or because an instance is leaked. Please hold on.

mattgallagher:
Yes, we've already reverted (and never shipped 6.10).

groue:
Phew

mattgallagher:
I'll take a closer look. The scenario we have is where 1 table with approximately 20 observers is written. The readers all finish but subsequent writes get blocked.

I might need to run an editable checkout of GRDB and instrument the itemsSemaphore to see what's happening.

groue:
> In your memory, was it a real dead-lock? I mean, could it have been reader starvation?

I can describe how #1350 has enabled a new scenario for reader starvation. First imagine there's a long-running background write transaction. For example, your app is synchronizing a big server payload. Meanwhile, your app starts a lot of observations. The first ones grab a reader connection. But they won't release it until they can acquire a write access, until the long-running write transaction has completed. Only then they'll register as transaction observers and start actual observation. Eventually all readers are taken, waiting for the completion of the long-running write transaction. Subsequent observations will thus be enqueued. They won't be able to perform their initial fetch. They will just look as if they were stuck in a deadlock. But they are only stuck temporarily - that's what I could test in various observation scenarios.

I do not know if this is what your application is suffering from, to be clear.

Before #1350, #1248 has started opening a new connection whenever an observation would start.

[...]

PR #1363, opened 2023-04-18
A test replicating Issue #1362

mattgallagher:
This is a relatively simple test demonstrating issue #1362: when the number of ValueObservers exceeds the size of the DatabasePool, GRDB will deadlock.

This PR exists purely for demonstration. I don't know if this test is appropriately located or following expectations of the codebase. I built this test from the components in ValueObservationTests, I've placed it there. But I'm sure this test could be located in another file or renamed, etc.


mattgallagher:
I've messed this up. I need to take another attempt.

Issue #966, opened 2021-04-22
Feature Request: debounce/throttle ValueObservers


steipete:
In my current SwiftUI app, I listen to ~20 different queries in a model that all update the UI. So there's a large list of `@Published` for each value queried. Currently value observers fire immediately, as soon as the database is changed.

Imagine I am running a background-process that imports 50.000 rows in chunks every ~0.05 seconds. This results in the UI being extremely slow, as the value observer fires all the time, resulting in constant UI updates. 

This is neither needed nor wanted; what I am looking for is a "eventual consistent" value observer, that I can configure to debounce. In this example I'd likely set it to 0.3 seconds, so it still feels like the UI is updating in realtime, but the main thread no longer maxes out a CPU core.

Even worse, some value observers trigger on the write queue if they are marked as being non-deterministic, which delays the background import and is another case I absolutely want to avoid.

```swift
ValueObservation
    .trackingConstantRegion(fetchValue) {
		$0.throttle(for: 0.5, latest: false)
	} 
    .publisher(in: dbReader)
    .removeDuplicates()
	...

OR

ValueObservation
    .trackingConstantRegion(fetchValue, frequency: .throttle(0.3))
    .publisher(in: dbReader)
    .removeDuplicates()
	...

// Inside ValueObservation:
enum Frequency {
	case immediate, throttle(NSTimeInterval), debounce(NSTimeInterval)
```

The case with just an enum is likely easier, because what do we do if someone tries a more complex publisher in the block and schedulers also don't quite fit.

Alternative, a publisher that just informs of changes could also work and would allow utmost flexibility to combine different observers for more efficient updating.

[Good overview of throttle vs debounce](https://rhonabwy.com/2019/12/15/combine-throttle-and-debounce/)

groue:
Hello @steipete,

This issue covers a large area, so we'll divide and conquer. Please give me a little time to prepare a few plans, and we'll discuss them.

groue:
Hello again @steipete, I understand why you present this topic as a "debounce/throttle" issue, because, well, your app does not care very much about when the view is refreshed, as long as it is, eventually.

I frankly don't see this topic solved very quickly, even if your issue creates the opportunity to start addressing it. We're talking about back pressure, which is not precisely an easy topic.

But I'd like to help right now, with the current state of GRDB.

[...]

PR #1364, opened 2023-04-18
Fix ValueObservation deadlock

mattgallagher:
This test demonstrates the possible deadlock between observers and writers as reported in issue #1362 . The observers all hold the `Pool.itemSemaphore` and the writer holds the `SerializedDatabase.queue` and neither can proceed since each is waiting for the other's resource.

To get this test reliable, I needed to insert a `Thread.sleep` inside the `ValueConcurrentObserver`. Obviously, this PR is not for merging. However, this `Thread.sleep` does not change the overall logic and exists primarily to simulate the system being under load.

The `observerCount` of 10 is not arbitrary. If you reduce the `observerCount` by 1, then the code doesn't deadlock so you can confirm that the test is otherwise logically sound.

### DIFF
--- a/GRDB/ValueObservation/Observers/ValueConcurrentObserver.swift
+++ b/GRDB/ValueObservation/Observers/ValueConcurrentObserver.swift
@@ -101,6 +101,19 @@ final class ValueConcurrentObserver<Reducer: ValueReducer, Scheduler: ValueObser
         }
     }
     
+    /// The fetching state for observation of constant regions.
+    enum FetchingState {
+        /// No need to fetch.
+        case idle
+        
+        /// Waiting for a fetched value.
+        case fetching
+        
+        /// Waiting for a fetched value, and for a subsequent  fetch after
+        /// that, because a change has been detected as we were fetching.
+        case fetchingAndNeedsFetch
+    }
+    
     /// Ability to notify observation events
     private struct NotificationCallbacks {
         let events: ValueObservationEvents
@@ -131,6 +144,9 @@ final class ValueConcurrentObserver<Reducer: ValueReducer, Scheduler: ValueObser
     
     /// Ability to notify observation events, protected by `lock`.
     private var notificationCallbacks: NotificationCallbacks?
+    
+    /// The fetching state for observation of constant regions.
+    @LockedBox private var fetchingState = FetchingState.idle
     
     /// Support for `TransactionObserver`, protected by the serialized writer
     /// dispatch queue.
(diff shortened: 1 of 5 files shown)


groue:
Thank you very much @mattgallagher.

This is a dead lock indeed 😖

Many observations are started. Some of them (set A) are able to access the writer and register a transaction observer. The remaining observations (set B) could not yet access a writer access. Each one of them has captured a reader. Before observations in the set B are enqueued for a write access, an actual write is performed, that triggers one of the early observations. From the write access that has just performed the modification, this observation waits until it can find a reader and fetch the fresh values. But no reader is available, because they've all been captured by the observations of set B.

The writer waits for a reader, and all readers are held by observations of the set B that are waiting for the writer. That's the deadlock.

Apologies, of course.

groue:
OK, so I had a good shower, and I have a plan :-) We're just paying the price of early days of observations, when I was obsessed about notifying every single change without ever ever missing one.

This is an atavism that was first broken in GRDB 5: ValueObservation on a DatabasePool has learned to start even if write transactions are currently running, with a price: individual changes performed during those early writes are not individually notified.

But after those early moments, an observation still exhibits a legacy behavior: it insists on notifying every single change.

That's why the triggered observation of the set A above blocks the writer until it could acquire a reader: it insists on fetching from the exact same state of the database that was left by the write.

My plan is to have this observation perform an asynchronous fetch (or, more precisely, something more akin to "setNeedsFetching", so that fast subsequent writes don't trigger more fetches than needed). This will remove the deadlock, without breaking any documented behavior of ValueObservation.

I hope this is clear 🤞

groue:
> Sounds okay to me. For my part, I'm happy to receive slightly fewer observations.

This reminds me of #966 by @steipete. I'm not promising I'll beat both horses at the same time, but this definitely belongs to the same topic.

> I leave it to you to decide if this requires breaking behavior.

The described changes are not supposed to be breaking, because the [documentation](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/valueobservation#ValueObservation-Behavior) clearly states (in order to support the behavior introduced by GRDB 5):

> `ValueObservation` may coalesce subsequent changes into a single notification.
>
> `ValueObservation` may notify consecutive identical values.

Those rules are supposed to give us enough freedom. Of course, some people may rely on some aspects of the current behavior. We'll see, then.

I don't want to lose what we have progressively gained for DatabasePool observations:

1. GRDB 5: The ability to fetch the initial value even if there is a concurrent write. This has introduced two side effects: the two relaxed rules above, but also the double notification of the initial value.
2. #1248: The removal of this double notification when WAL snapshots are available. This has introduced a side effect: the creation of more reader connections.
3. #1350: The removal of those extra connections. Side effect: our deadlock 😬 
4. And now, let's now remove the deadlock!

groue:
Rewrote the test and removed the FIXME. Now I can look for the solution.

groue:
Last CI tests have passed, but the previous ones did not. Things have turned flaky.

I have to check if tests are badly written, or if the fix for the deadlock has introduced a bug.

groue:
OK, I think we're good.

groue:
Crap, wrong destination branch 😬😅

groue:
The fix has shipped in v6.12.0. Thanks again for your help, @mattgallagher 👍

PR #1384, opened 2023-06-02
Allow ValueObservation to start from a truncated wal file

### LINKED ISSUE #1383: SQL logic error when starting a ValueObservation immediately after opening a DatabasePool
### TL;DR

GRDB v6.14.0 fixes this issue. If you can't upgrade, and your app doesn't perform SQLite checkpoints, here's a workaround - to be run right after creating the `DatabasePool`:

        let dbPool: DatabasePool = ... // your init code here
        
        // https://github.com/groue/GRDB.swift/issues/1383
        // FIXME: remove this after upgrading to GRDB v6.14.0
        try dbPool.writeWithoutTransaction { db in
            try db.inSavepoint {
                try db.execute(sql: """
                CREATE TABLE grdb_issue_1383 (id INTEGER PRIMARY KEY);
                DROP TABLE grdb_issue_1383;
                """)
                return .commit
            }
        }

### What did you do?

I created a new `DatabasePool` with an existing .sqlite database file that did not have -wal or -shm files next to it. I noticed that by the time the `DatabasePool` was created, there was already a zero-length -wal file and an -shm file next to the .sqlite file.

I debugged into it, and executing this line creates the -wal and -shm files: https://github.com/groue/GRDB.swift/blob/v6.13.0/GRDB/Core/Database.swift#L420
Notably, this `if` condition evaluated to `false`, because a zero-length -wal file existed at this point: https://github.com/groue/GRDB.swift/blob/v6.13.0/GRDB/Core/DatabasePool.swift#L87

Then, I tried to start a ValueObservation with the `DatabasePool`, which failed on this line: https://github.com/groue/GRDB.swift/blob/v6.13.0/GRDB/Core/WALSnapshot.swift#L45

I suspect it fails because of [this requirement](https://www.sqlite.org/c3ref/snapshot_get.html):
> One or more transactions must have been written to the current wal file since it was created on disk (by any connection). This means that a snapshot cannot be taken on a wal mode database with no wal file immediately after it is first opened. At least one transaction must be written to it first.

If I place a breakpoint at the end of the `DatabasePool` init, and execute this command from the debugger, the `ValueObservation` seems to work as expected: https://github.com/groue/GRDB.swift/blob/v6.13.0/GRDB/Core/DatabasePool.swift#L93

I'm not sure if this is important, but the .sqlite file was first created using GRDB 5.12.0, then I tried opening it with GRDB 6.13.0.

I expect a `ValueObservation` to work even if I start it right after opening a `DatabasePool`.

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 6.13.0
**Xcode version:** 14.2
**Swift version:** 5.7.2
**macOS version running Xcode:** 12.6.5

groue:
This pull request fixes #1383.

When `ValueObservation` starts from a `DatabasePool`, on a database that has an empty wal file (the `-wal` temporary file next to the main database file), it would fail with an SQLite error 1: SQL logic error.

The cause of this error is that [it is impossible to create a wal snapshot from a missing or empty wal file](https://www.sqlite.org/c3ref/snapshot_get.html). `ValueObservation` uses wal snapshots since #1248.

In order to fix this error, this pull request brings the following changes:

- The `DatabasePool` initializer makes sure the wal file is not empty. [^1]
- `ValueObservation` fallbacks to a "degraded" mode (pre-[#1248](https://github.com/groue/GRDB.swift/pull/1248)), whenever it starts from an empty wal file. The consequence of this degradation is that an observation always notifies two values when it starts, even when the database content is not modified. See [#1248](https://github.com/groue/GRDB.swift/pull/1248) for more information.

Thanks to those changes, it is expected that `ValueObservation` only runs in the degraded mode in applications that perform explicit truncating wal checkpoints. Those applications will need to recreate a non-empty wal file if they want to avoid the degraded observation mode. This can not be done by GRDB, due to the race condition between the write access that recreates the wal file, and the read accesses performed by observations: only apps can guarantee proper scheduling (i.e. recreate the wal file *before* starting an observation).

cc @kustra

[^1]: This may annoy some applications. We'll add a way to opt-out if requested, because making observations work out of the box has higher priority.


groue:
@kustra, it is now shipped in v6.15.0.

PR #1397, opened 2023-06-22
Swift API for creating database views

groue:
This pull request addresses https://github.com/groue/GRDB.swift/discussions/1394 and introduces the `create(view:options:columns:as:)` and `drop(view:)` methods.

You can create a view with an [`SQL`](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/sql) literal:

```swift
// CREATE VIEW hero AS SELECT * FROM player WHERE isHero == 1
let sql: SQL = """
    SELECT * FROM player WHERE isHero == 1
    """
try db.create(view: "hero", asLiteral: sql)
```

You can create a view with an [`SQLRequest`](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/sqlrequest):

```swift
// CREATE VIEW hero AS SELECT * FROM player WHERE isHero == 1
let request = SQLRequest(literal: """
    SELECT * FROM player WHERE isHero == 1
    """)
try db.create(view: "hero", as: request)
```

You can also create a view with a [`QueryInterfaceRequest`](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/queryinterfacerequest):

```swift
// CREATE VIEW hero AS SELECT * FROM player WHERE isHero == 1
let request = Player.filter(Column("isHero") == true)
try db.create(view: "hero", as: request)
```

When creating views in [migrations](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/migrations), it is recommended to avoid using record types defined in the application. Instead of the `Player` record type, prefer `Table("player")`:

```swift
// RECOMMENDED IN MIGRATIONS
let request = Table("player").filter(Column("isHero") == true)
try db.create(view: "hero", as: request)
```

@mallman, what do you think? That should be enough, right?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,9 +115,11 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 
 ## Next Release
 
+- **New**: [#1397](https://github.com/groue/GRDB.swift/pull/1397) by [@groue](https://github.com/groue): Swift API for creating database views
 - **New**: [#1401](https://github.com/groue/GRDB.swift/pull/1401) by [@kustra](https://github.com/kustra): Linux compilation fixes
 - **New**: [#1402](https://github.com/groue/GRDB.swift/pull/1402) by [@groue](https://github.com/groue): Upgrade custom SQLite builds to 3.42.0
 - **New**: [#1403](https://github.com/groue/GRDB.swift/pull/1403) by [@groue](https://github.com/groue): GitHub CI: test Xcode 14.3.1, macOS 13
+
 
 ## 6.15.1
 
--- a/GRDB/Documentation.docc/DatabaseSchema.md
+++ b/GRDB/Documentation.docc/DatabaseSchema.md
@@ -16,7 +16,7 @@ try db.create(table: "player") { t in
 
 When you plan to evolve the schema as new versions of your application ship, wrap all schema changes in <doc:Migrations>.
 
-Prefer Swift methods over raw SQL queries. They allow the compiler to check if a schema change is available on the target operating system. Only use a raw SQL query when no Swift method exist (when creating views or triggers, for example).
+Prefer Swift methods over raw SQL queries. They allow the compiler to check if a schema change is available on the target operating system. Only use a raw SQL query when no Swift method exist (when creating triggers, for example).
 
 When a schema change is not directly supported by SQLite, or not available on the target operating system, database tables have to be recreated. See <doc:Migrations> for the detailed procedure.
 
--- a/GRDB/Documentation.docc/DatabaseSchema.md
+++ b/GRDB/Documentation.docc/DatabaseSchema.md
@@ -372,6 +372,13 @@ extension Team: TableRecord {
 - ``TableOptions``
 - ``VirtualTableModule``
 
+### Database Views
+
+- ``Database/create(view:options:columns:as:)``
+- ``Database/create(view:options:columns:asLiteral:)``
+- ``Database/drop(view:)``
+- ``ViewOptions``
+
 ### Database Indexes
 
 - ``Database/create(indexOn:columns:options:condition:)``
--- a/GRDB/QueryInterface/SQL/SQLExpression.swift
+++ b/GRDB/QueryInterface/SQL/SQLExpression.swift
@@ -848,7 +848,7 @@ extension SQLExpression {
     ///
     ///     try "foo'bar".databaseValue.quotedSQL(db) // "'foo''bar'""
     func quotedSQL(_ db: Database) throws -> String {
-        let context = SQLGenerationContext(db, argumentsSink: .forRawSQL)
+        let context = SQLGenerationContext(db, argumentsSink: .literalValues)
         return try sql(context)
     }
     
--- a/GRDB/QueryInterface/SQLGeneration/SQLGenerationContext.swift
+++ b/GRDB/QueryInterface/SQLGeneration/SQLGenerationContext.swift
@@ -172,15 +172,21 @@ class StatementArgumentsSink {
     private(set) var arguments: StatementArguments
     private let rawSQL: Bool
     
-    /// A sink which does not accept any arguments.
-    static let forRawSQL = StatementArgumentsSink(rawSQL: true)
+    /// A sink which turns all argument values into SQL literals.
+    ///
+    /// The `"WHERE name = \("O'Brien")"` SQL literal is turned into the
+    /// `WHERE name = 'O''Brien'` SQL.
+    static let literalValues = StatementArgumentsSink(rawSQL: true)
     
     private init(rawSQL: Bool) {
         self.arguments = []
         self.rawSQL = rawSQL
     }
     
-    /// A sink which accepts arguments
+    /// A sink which turns all argument values into `?` SQL parameters.
+    ///
+    /// The `"WHERE name = \("O'Brien")"` SQL literal is turned into the
+    /// `WHERE name = ?` SQL.
     convenience init() {
         self.init(rawSQL: false)
     }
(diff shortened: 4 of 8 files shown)


mallman:
My quick-n-dirty implementation was very simple, so I can provide it inline here:

```swift
extension Database {
    public func create(
        view: String,
        inSchema schema: String? = nil,
        as sqlLiteral: SQL,
        options: ViewOptions = [])
    throws
    {
        let context = SQLGenerationContext(self, argumentsSink: .forRawSQL)
        let sqlRequestSql = try sqlLiteral.sql(context)
        var chunks: [String] = []
        chunks.append("CREATE")
        if options.contains(.temporary) {
            chunks.append("TEMPORARY")
        }
        chunks.append("VIEW")
        if options.contains(.ifNotExists) {
            chunks.append("IF NOT EXISTS")
        }
        if let schema {
            chunks.append(schema.quotedDatabaseIdentifier + "." + view.quotedDatabaseIdentifier)
        } else {
            chunks.append(view.quotedDatabaseIdentifier)
        }
        chunks.append("AS")
        chunks.append(sqlRequestSql)
        let sql = chunks.joined(separator: " ")
        try execute(sql: sql)
    }
}

public struct ViewOptions: OptionSet {
    public let rawValue: Int

    public init(rawValue: Int) { self.rawValue = rawValue }

    /// Only creates the view if it does not already exist.
    public static let ifNotExists = ViewOptions(rawValue: 1 << 0)

    /// Creates a temporary view.
    public static let temporary = ViewOptions(rawValue: 1 << 1)
}
```

As you can see, our API and implementation are nearly identical. (Great minds think alike!)

There is some divergence. Your API takes column names, which is more complete. It also subsumes the "schema" part of a view identifier into the view name itself. That seems perfectly reasonable to me, and my implementation doesn't validate the name components anyway. If SQLite doesn't like it, SQLite will complain. That's the way to go.

The main difference is my API takes a `SQL` instance to define the view, while your API takes a `SQLRequest`. The latter provides more safety, because it has already verified that the item in question is a valid select. A `SQL` literal can be just about anything. The `SQL` literal worked fine for me, but I see how a `SQLRequest` is desirable. It makes more sense when we consider the other API that takes a `QueryInterfaceRequest`.

If I have any more specific feedback, I will create a formal review or leave comments inline.

Thanks for taking this on!

mallman:
I also want to add one, somewhat orthogonal, but related comment. I also created an API for `INSERT INTO ... SELECT ...`. This is unnecessary, because I can do this with an ordinary `try db.execute(literal: ...)`. However, it feels a bit awkward given all of the structured APIs for inserting a single table row. GRDB does not appear to have a structured API for the `INSERT INTO ... SELECT ...` statement. My implementation looks like:

```swift
extension Database {
    public func insert(
        into table: String,
        inSchema schema: String? = nil,
        columns: [String]? = nil,
        select sqlLiteral: SQL)
    throws
    {
        let context = SQLGenerationContext(self, argumentsSink: .forRawSQL)
        let sqlRequestSql = try sqlLiteral.sql(context)
        var chunks: [String] = []
        chunks.append("INSERT INTO")
        if let schema {
            chunks.append(schema.quotedDatabaseIdentifier + "." + table.quotedDatabaseIdentifier)
        } else {
            chunks.append(table.quotedDatabaseIdentifier)
        }
        if let columns {
            var columnsChunks: [String] = []
            columnsChunks.append("(")
            columnsChunks.append(columns.map({$0.quotedDatabaseIdentifier}).joined(separator: ", "))
            columnsChunks.append(")")
            let columnsString = columnsChunks.joined()
            chunks.append(columnsString)
        }
        chunks.append(sqlRequestSql)
        let sql = chunks.joined(separator: " ")
        try execute(sql: sql)
    }
}
```

Just food for thought really. If you like the idea, you might consider adding this as a feature as well. In the end, to avoid maintaining a patch, I abandoned this API and just use the existing interface.

groue:
> There is some divergence. Your API takes column names, which is more complete. It also subsumes the "schema" part of a view identifier into the view name itself. That seems perfectly reasonable to me, and my implementation doesn't validate the name components anyway. If SQLite doesn't like it, SQLite will complain. That's the way to go.

I dropped the schema part for consistency with existing public apis, which do not support schemas at all.

If the user provides a qualified name, the current implementation will not generate the expected SQL:

```swift
// Creates "mySchema.myView" in the "main" schema
// CREATE VIEW "mySchema.myView" AS ...
try db.create(view: "mySchema.myView", as ...)
```

It will be impossible to support both identifier escaping (so that users can create tables and views named like SQL keywords), and schemas, from raw strings: if we escape `group` into `"group"`, then we have to escape `mySchema.myView` into `"mySchema.myView"`.

So your `inSchema` parameter is much probably a glimpse in the future :-) But I'd rather add support for schema globally: `create(view:inSchema:)`, `create(table:inSchema:)`, `create(index:inSchema:)`, etc, in another PR.

> The main difference is my API takes a `SQL` instance to define the view, while your API takes a `SQLSubqueryable`.

Yes, `SQLSubqueryable` is the protocol we need here.

That's because we generally try to respect the SQL grammar, and use the protocols that match what SQLite expects at a given position (`SQLSelectable`, `SQLExpressible`, [etc](https://github.com/groue/GRDB.swift/blob/master/Documentation/QueryInterfaceOrganization.md)).

Also, if we'd **only** support `SQL`, users would need to build one from their requests. This is possible, but I'm not sure anybody would expect it:

```swift
let request = ... // SQLRequest or QueryInterfaceRequest
try db.create(view: "hero", as: "\(request)") // 😮
```

(Pause)

This discussion makes me wonder whether if it would be possible to have `SQL` conform to `SQLSubqueryable`... 🤔 It already conforms to `SQLExpressible`, which makes it possible to use `SQL` as an expression, a selection, and an ordering term. Why not using `SQL` as a (sub)query as well? I'll check if there isn't any undesired side effect.

groue:
> I also want to add one, somewhat orthogonal, but related comment. I also created an API for `INSERT INTO ... SELECT ...`.

Thanks for the inspiration! I have to go to work now, but I'll have a closer look 👍

groue:
> This discussion makes me wonder whether if it would be possible to have `SQL` conform to `SQLSubqueryable`... 🤔

No test breaks when I do. And it makes it possible to write:

```swift
// This matches your own version!
try db.create(view: "view1", as: SQL("SELECT ... "))
```

I'm still investigating, though, because unlike expressions, subqueries often have to be wrapped in parenthesis in raw SQL. I'm afraid this could break some existing code where some `SQL` expressions would be misinterpreted as subqueries after this change.

mallman:
> > I also want to add one, somewhat orthogonal, but related comment. I also created an API for `INSERT INTO ... SELECT ...`.
> 
> Thanks for the inspiration! I have to go to work now, but I'll have a closer look 👍

What??? I thought GRDB _was_ your work??? 😮 🤣

mallman:
> > There is some divergence. Your API takes column names, which is more complete. It also subsumes the "schema" part of a view identifier into the view name itself. That seems perfectly reasonable to me, and my implementation doesn't validate the name components anyway. If SQLite doesn't like it, SQLite will complain. That's the way to go.
> 
> I dropped the schema part for consistency with existing public apis, which do not support schemas at all.
> 
> If the user provides a qualified name, the current implementation will not generate the expected SQL:
> 
> ```swift
> // Creates "mySchema.myView" in the "main" schema
> // CREATE VIEW "mySchema.myView" AS ...
> try db.create(view: "mySchema.myView", as ...)
> ```

I hadn't noticed that before. I never tried using non-default schemata. I use attached databases, but the names in those databases are distinct from the names in the other databases—there's no ambiguity.

I added a separate `schema` parameter, because I originally contemplated putting my `NSCollectionView` views in another schema.

groue:
> I hadn't noticed that before. I never tried using non-default schemata. I use attached databases, but the names in those databases are distinct from the names in the other databases—there's no ambiguity.

A feature that was never requested so far :-)

> I added a separate `schema` parameter, because I originally contemplated putting my `NSCollectionView` views in another schema.

That counts as half a request, at least ;-)

groue:
Since #1398 was rejected, I added `create(view:asLiteral:)` that accepts an `SQL` literal as an argument. For consistency with other existing apis, there is "literal" in the argument name.

groue:
Shipped in v6.16.0!

PR #1398, opened 2023-06-23
Make `SQL` able to feed SELECT statements, and deprecate redundant `CommonTableExpression` initializers

groue:
This pull request has `SQL`, the type that enables [SQL interpolation](https://github.com/groue/GRDB.swift/blob/master/Documentation/SQLInterpolation.md), conform to [`SQLSubqueryable`](https://swiftpackageindex.com/groue/grdb.swift/v6.15.1/documentation/grdb/sqlsubqueryable).

This protocol conformance makes `SQL` able to feed the [`select-stmt` SQLite grammar pattern](https://www.sqlite.org/syntax/select-stmt.html) in GRDB apis that need such a value.

This idea comes from #1398, as a way to make it easier to create views from `SQL` instances.

For consistency with #1398, and in order to simplify the api surface, the `CommonTableExpression` initializers that accepts a `literal` argument, or a `sql`/ `arguments` pair, have been deprecated:

```swift
// Deprecated
let playerNameCTE = CommonTableExpression(
    named: "playerName",
    literal: "SELECT ...")
```

Instead, use an `SQL` value with the `request` argument:

```swift
// Replacement
let playerNameCTE = CommonTableExpression(
    named: "playerName",
    request: SQL("SELECT ..."))
```

cc @mallman 

TODO:

- [ ] update the QueryInterfaceOrganization.md illustration
- [ ] Warning: check the impacts of this change. For example, `SQL` values that are turned into an expression via their `SQLSubqueryable` conformance trigger a different code path in `SQLExpression.qualified(with:)` and `SQLExpression.sql(_:wrappedInParenthesis:)`.

### DIFF
--- a/Documentation/CommonTableExpressions.md
+++ b/Documentation/CommonTableExpressions.md
@@ -42,12 +42,12 @@ let name = "O'Brien"
 // WITH playerName AS (SELECT 'O''Brien') ...
 let playerNameCTE = CommonTableExpression(
     named: "playerName",
-    sql: "SELECT ?", arguments: [name])
+    request: SQL(sql: "SELECT ?", arguments: [name]))
 
 // WITH playerName AS (SELECT 'O''Brien') ...
 let playerNameCTE = CommonTableExpression(
     named: "playerName",
-    literal: "SELECT \(name)")
+    request: SQL("SELECT \(name)"))
 
 // WITH playerName AS (SELECT 'O''Brien') ...
 let request = SQLRequest("SELECT \(name)")
@@ -63,7 +63,7 @@ All CTEs can be provided with explicit column names:
 let pairCTE = CommonTableExpression(
     named: "pair", 
     columns: ["a", "b"], 
-    sql: "SELECT 1, 2")
+    request: SQL("SELECT 1, 2"))
 ```
 
 Recursive CTEs need the `recursive` flag. The example below selects all integers between 1 and 1000:
@@ -75,11 +75,11 @@ let counterCTE = CommonTableExpression(
     recursive: true,
     named: "counter",
     columns: ["x"],
-    sql: """
+    request: SQL("""
         VALUES(1)
         UNION ALL
         SELECT x+1 FROM counter WHERE x<1000
-        """)
+        """))
 ```
 
 > **Note**: many recursive CTEs use the `UNION ALL` SQL operator. The query interface does not provide any Swift support for it, so you'll generally have to write SQL in your definitions of recursive CTEs.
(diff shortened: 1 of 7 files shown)


groue:
I close this pull request and avoid merging it.

The main problem is that _subqueries_ are very different from _expressions_ with regard to "qualification".

Qualification is the process of decorating expressions with table names in order to avoid ambiguities. It happens all the time in GRDB requests that involve several tables, or tables that are aliased with an explicit name:

```swift
try DatabaseQueue().write { db in
  try db.create(table: "player") { $0.column("name") }
  
  let column = Column("name")
  let columnAsSQL: SQL = "\(column)"
  let alias = TableAlias(name: "p")
  let request = Table("player")
    .aliased(alias)
    .select(columnAsSQL)
  let statement = try request.makePreparedRequest(db).statement
  
  // The name column was qualified with the "p" alias:
  // SELECT "p"."name" FROM "player" "p"
  print(statement.sql)
}
```

This behavior is very important for [Embedding SQL in Query Interface Requests](https://github.com/groue/GRDB.swift/blob/master/README.md#embedding-sql-in-query-interface-requests). This is how users can define their own Swift functions that extend the query interface in a way that works well with complex queries (often without the user being even aware of it):

```swift
/// Returns a date in the `YYYY-MM-DD` format
func date(_ value: some SQLSpecificExpressible) -> SQLExpression {
  // The value will be automatically qualified when needed
  SQL("DATE(\(value))").sqlExpression
}

try DatabaseQueue().write { db in
  try db.create(table: "player") { $0.column("dateTime") }
  
  let alias = TableAlias(name: "p")
  let request = Table("player")
    .aliased(alias)
    .select(date(Column("dateTime")))
  let statement = try request.makePreparedRequest(db).statement
  
  // SELECT DATE("p"."dateTime") FROM "player" "p"
  print(statement.sql)
}
```

Such qualification does never happen in subqueries, because subqueries have a namespace of their own:

```sql
CREATE TABLE t1(a);
CREATE TABLE t2(a);

-- Those `a` columns are not ambiguous
SELECT * FROM t1 WHERE a in (SELECT a FROM t2);
```

Having `SQL` conform to `SQLSubqueryable` would create a conceptual mess in GRDB:

```swift
let column = Column("a")
let sql1: SQL = "SELECT \(column) FROM t2"
let sql2: SQL = "\(column)"
-- Who needs qualification, who does not?
let query: SQL = "SELECT * FROM t1 WHERE \(sql2) in (\(sql1))"
```

That's why I close this PR. It's just a promise for future headaches.

PR #1405, opened 2023-07-06
Simplify the declaration of BelongsTo associations in the database schema

groue:
This pull requests introduces a new `belongsTo` method on table definitions that helps building associations and foreign keys:

```swift
    // NEW
    t.belongsTo("team")
    t.belongsTo("country").notNull()
}
```

This is strictly equivalent to:

```swift
// The old way
    t.column("teamId", .integer)
        .references("team")
        .indexed()
    t.column("countryCode", .text)
        .references("country")
        .indexed()
        .notNull()
}
```

All primary keys are supported, including composite ones. Foreign keys declared with `belongsTo` are indexed by default (this is often useful, easily forgotten with the old apis, and it can be disabled when needed).

The intent of this PR is to reduce boilerplate, and also to help users choose their associations (thanks to the common "belongsTo" method name):

```swift
// Player table "belongs to" team and country
    t.belongsTo("team")
    t.belongsTo("country").notNull()
}

// Player record "belongs to" Team and Country
struct Player: TableRecord {
    static let team = belongsTo(Team.self)
    static let country = belongsTo(Country.self)
    var teamId: Int64?
    var countryCode: String
}
```

Examples of usage can be found in the [Associations Guide](https://github.com/groue/GRDB.swift/blob/master/Documentation/AssociationsBasics.md) as well as in the [reference documentation](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/tabledefinition/belongsto(_:intable:ondelete:onupdate:deferred:indexed:)) (both available after the pull request has been merged).


groue:
TODO: allow the definition of several `belongsTo` to the same table, with a `key` (or `prefix`) argument.

PR #1429, opened 2023-09-11
Extend `DatabaseValueConvertible` JSON parsing format customization

### LINKED ISSUE #1428: Extend `DatabaseValueConvertible` JSON parsing format customization
### What did you do?

Added a `Codable` struct with `Double` values that can be `.infinity` as a field in a table. As expected, this throws an error: 

> Unable to encode Double.infinity directly in JSON. Use JSONEncoder.NonConformingFloatEncodingStrategy.convertToString to specify how the value should be encoded.

That we could customize the `JSONDecoder` and `JSONEncoder` just as we can do for `FetchableRecord` and `EncodableRecord`

```swift
extension DataWeNeedButNotNormalized: DatabaseValueConvertible {
    public static func databaseJSONDecoder() -> JSONDecoder {
        customDecoder
    }

    public static func databaseJSONEncoder() -> JSONEncoder {
        customEncoder
    }
}
```

### What happened instead?

Because `DatabaseValueConvertible` doesn't have methods to customize the `JSONDecoder` and `JSONEncoder` used, we had to manually conform to the protocol, which it is easy, but as a convenience we think it would be nice to match `FetchableRecord` and `EncodableRecord` here and allow `DatabaseValueConvertible` to still be auto-synthesized for `Codable` entities while providing a custom Decoder and/or Encoder.

**In summary**, I suggest we add the following two methods to `DatabaseValueConvertible`:

`public static func databaseJSONDecoder() -> JSONDecoder`
`public static func databaseJSONEncoder() -> JSONEncoder`

### Environment

**GRDB flavor(s):** GRDB
**GRDB version:** 6.10.1
**Xcode version:** 14.3.1
**Swift version:** 5.6
**Platform(s) running GRDB:** iOS, watchOS and tvOS
**macOS version running Xcode:** 13.4

### Demo Project

[GRDBAsyncDemo+InfinityValueTest.zip](https://github.com/groue/GRDB.swift/files/12571449/GRDBAsyncDemo%2BInfinityValueTest.zip)

JhonnyBillM:
### Summary

Fixes #1428.

Brings `DatabaseValueConvertible` to parity with `FetchableRecord` and `EncodableRecord` by allowing users to customize their JSON encoding and decoding using these two added methods:
- `static func databaseJSONDecoder() -> JSONDecoder`
- `static func databaseJSONEncoder() -> JSONEncoder`

This work follows the existing implementation for Records, but with the caveat of having to introduce an optional `jsonEncoder` property in `DatabaseValueEncoder`. Let me know if we should handle this differently, or if things like these just make expanding this protocol not worth it and keep encouraging users to just implement the required methods.

---

### Pull Request Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,7 @@ How you can Contribute
     - Get familiar with the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/).
     - Spaces, not tabs.
     - Whitespace-only lines are not trimmed.
-    - Documentation comments are hard-wrapped at column 80 (Xcode > Preferences > Text Editing > Display > [X] Page guide at column: 80).
+    - Documentation comments are hard-wrapped at column 76 (Xcode > Preferences > Text Editing > Display > [X] Page guide at column: 76).
     - No Swiftlint warning after a build.
 
 
--- a/GRDB/Core/DatabaseValueConvertible.swift
+++ b/GRDB/Core/DatabaseValueConvertible.swift
@@ -62,13 +62,13 @@ public protocol DatabaseValueConvertible: SQLExpressible, StatementBinding {
     /// - returns: A decoded value, or, if decoding is impossible, nil.
     static func fromMissingColumn() -> Self?
     
-    /// Returns the `JSONDecoder` that decodes the value for a given column.
+    /// Returns the `JSONDecoder` that decodes the value.
     ///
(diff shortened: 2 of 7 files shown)


groue:
Thank you very much @JhonnyBillM! We have seen in #1248 why this pull request is useful and very welcomed :-)

I added a few changes. When CI tests are completed, we'll be able to merge!

groue:
Thank you @JhonnyBillM, this will ship in the next version :-)

Issue #1483, opened 2024-01-17
Recommend the IMMEDIATE transaction type in the Sharing a Database guide


groue:
[This Mastodon post](https://hachyderm.io/@anze3db@fosstodon.org/111765707809150937) by @anze3db explains why IMMEDIATE transactions are useful in a context of concurrent writes.

Concurrent writes do not happen when a database is only ever accessed from a single GRDB connection (`DatabaseQueue` or `DatabasePool`) from a single process: those connections make sure all writes are serialized.

But as soon as multiple processes can access a database, concurrent writes can happen. The [Sharing a Database](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/databasesharing) guide should recommend immediate transactions.

PR #1485, opened 2024-01-17
Recommend IMMEDIATE transactions when a database is shared between multiple processes

groue:
This pull request addresses #1483 by recommending to add a single line of code in applications that share a database across multiple processes:

```swift
configuration.defaultTransactionKind = .immediate
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 
 ## Next Release
 
+- **Documentation Update**: The [Sharing a Database](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/databasesharing) guide was updated with a new recommendation. When an SQLite database is shared between multiple processes, writers should always perform IMMEDIATE transactions in order to avoid the `SQLITE_BUSY` error that can occur when transactions overlap. The new recommendation fits in a single line of code: `configuration.defaultTransactionKind = .immediate`.
 - **New**: Associations that involve views instead of tables are supported, but they require an explicit `ForeignKey` in their definition. Now a clear diagnostic message is emitted, instead of a unhelpful "no such table" runtime error.
 
 ## 6.24.1
--- a/GRDB/Documentation.docc/DatabaseSharing.md
+++ b/GRDB/Documentation.docc/DatabaseSharing.md
@@ -152,11 +152,14 @@ If several processes want to write in the database, configure the database pool 
 
 ```swift
 var configuration = Configuration()
+configuration.defaultTransactionKind = .immediate
 configuration.busyMode = .timeout(/* a TimeInterval */)
 let dbPool = try DatabasePool(path: ..., configuration: configuration)
 ```
 
-With such a setup, you may still get `SQLITE_BUSY` errors from all write operations. They will occur if the database remains locked by another process for longer than the specified timeout. You can catch those errors:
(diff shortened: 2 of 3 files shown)


groue:
Thank you @anze3db for [your post](https://blog.pecar.me/django-sqlite-dblock) that has clarified this topic 👍 It's not the first time Django has inspired GRDB :-)

groue:
cc @FanisaKim, @mrackwitz If you are experiencing `SQLITE_BUSY` (5) on dbs that can be written to from multiple processes, you may try the configuration change described above. It is already available in GRDB 6.24 (this PR is just a documentation change).

It transforms scenarios that fail when concurrent write transactions overlap:

1. Victim opens a transaction
2. Victim reads
3. Attacker opens a transaction and writes
4. Victim writes 🛑 SQLITE_BUSY, ignoring the `busyMode` timeout

and 

1. Attacker opens a transaction
2. Victim opens a transaction
3. Victim reads
4. Attacker writes
5. Victim A writes 🛑 SQLITE_BUSY, ignoring the `busyMode` timeout

Into the success scenarios:

1. Victim opens an immediate transaction
2. Victim reads
3. Attacker opens an immediate transaction and is ⏸️ suspended according to `busyMode`
4. Victim writes ✅
5. Victim commits ✅
6. Attacker is resumed and proceeds with its write ▶️✅ (or fails with SQLITE_BUSY if the timeout was expired)

and

1. Attacker opens an immediate transaction
2. Victim opens an immediate transaction and is ⏸️ suspended according to `busyMode`
3. Attacker writes ✅
4. Attacker commits ✅
5. Victim is resumed and proceeds with its write ▶️✅ (or fails with SQLITE_BUSY if the timeout was expired)

Jason-Abbott:
@groue a bit of a remedial question, perhaps: since you mention “multiple processes” can I assume this isn’t a concern, even across tasks, if there’s only ever one instance of a GRDB pool or queue?

groue:
Thank you for asking, @Jason-Abbott. You got it exactly right. An instance of `DatabaseQueue` and `DatabasePool` serializes all write transactions, and this avoids all the problems created by overlapping write transactions. I amended the description of the PR above for more clarity.

freak4pc:
> 😬 When an application creates several instances of `DatabaseQueue` and `DatabasePool` for accessing a single database... Well the recommendation is to refactor the app so that the [Rule 1](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/concurrency#Rule-1-Connect-to-any-database-file-only-once) of GRDB concurrency is honored.

Just bumped into this thread and wanted to confirm this point - 
In the context of App Extensions - each piece of code is separated into its own module-space so having "multiple instances of DatabasePool" seems inevitable? (one for each process). Am I misunderstanding this ?

groue:
Note to myself: the late Keith Medcalf [recommended](https://sqlite.org/forum/forumpost/099cb3ca36c10f2d) that all transactions that intend to write are immediate:

> Also, all transactions in which one intends to WRITE should be commenced with BEGIN IMMEDIATE. BEGIN DEFERRED should only be used for transaction which READ only. Upgrading locks is how one gets oneself in a mess. It has been known for almost a century that the way to avoid deadlocks is to ALWAYS obtain all required locks at the same time and in the same order, every time, and to always release all locks when the acquisition of one that is required fails.

Maybe GRDB should not bother users with `defaultTransactionKind` and do the right thing, since the public api distinguishes reads from writes.

PR #1503, opened 2024-03-08
Conform Database.ColumnType to Sendable

simba909:
First up, thank you for making this! 🙏🏻 I have used GRDB in quite a few projects throughout my career now and it's a go-to tool for me. Making my first contribution to it a fairly minor improvement :)

This silences a (couple of, due to the static lets) warning when building GRDB with strict concurrency checking enabled. `Database.ColumnType` is `RawRepresentable` as a `String` (which also is `Sendable`) and its `rawValue` cannot change, making this a safe conformance.

Please let me know if there are things I have missed or if this can be done in a better way!

### Pull Request Checklist

- [x] CONTRIBUTING: You have read https://github.com/groue/GRDB.swift/blob/master/CONTRIBUTING.md
- [x] BRANCH: This pull request is submitted against the `development` branch.
- [ ] DOCUMENTATION: Inline documentation has been updated.
- [ ] DOCUMENTATION: README.md or another dedicated guide has been updated.
- [x] TESTS: Changes are tested.
- [x] TESTS: The `make smokeTest` terminal command runs without failure.

PR #1510, opened 2024-03-17
Add Sendable conformances and unavailabilities

groue:
This pull request adds or removes `Sendable` conformance to as many types as possible.

Some types are explicitly non-sendable. For example `Database` must be used from a database access method such as `read` or `write`, and will never become `Sendable`.

Some other types are left without any explicit annotation. Those will become sendable in GRDB 7. Some apps are able to add `@unchecked Sendable` conformance to some of them, when mandatory, **under some specific conditions that should be checked in [this discussion](https://github.com/groue/GRDB.swift/discussions/1509)**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 ## Next Release
 
 - **New**: [#1503](https://github.com/groue/GRDB.swift/pull/1503) by [@simba909](https://github.com/simba909): Conform Database.ColumnType to Sendable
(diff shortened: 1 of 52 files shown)


groue:
cc @simba909: this PR completes your #1503 as far as GRDB 6 can go. For the record, I could find the exhaustive list of public types by building the DocC documentation from the GRDB SPM package, and looking for types in the built documentation:

```
$ find ~/Library/Developer/Xcode/DerivedData/GRDB-dbuvchqinknqzidnxbiauivgvrrm/Build/Products/Debug/GRDB.doccarchive/data/documentation/grdb -type d | sed "s/^.*grdb// | sort"
/adaptedfetchrequest
/aggregatingrequest
/allcolumns
...
```

PR #1515, opened 2024-03-23
Support for the CAST SQLite function

### LINKED ISSUE #1514: How can I do floating-point division on an association count?
### What did you do?

Hi @groue. 

Thanks again for the great library.

I have another support question. 

I would like to cast `association count` to `floating-point` to do `floating-point` division.
How might I achieve the below?

```swift
struct Quest: TableRecord, ... {

    let id: UUID
    let title: String

    static let activeMission = hasMany(
        Mission.self,
        using: Mission.questForeignKey
    )
    .filter(Mission.Column.isCompleted == false)

    static let totalMission = hasMany(
        Mission.self,
        using: Mission.questForeignKey
    )

}

struct Mission: TableRecord, ... {

    let id: UUID
    let title: String
    let isCompleted: Bool 
    let questId: UUID

    static let questForeignKey = ForeignKey([Column.questId])

}

struct QuestProgress: FetchableRecord, ... {

    let quest: Quest
    let progress: Double

}

extension QuestProgress {

    static func request(questId: UUID) -> QueryInterfaceRequest<QuestProgress> {

        Quest 
            .all()
            .filter(id: ...)
            .annotated(
                with: 
                    (
                        /// ------------------------------------------------------
                        ///     This will do integer division. 
                        /// 👇  Is there a way to cast to floating point division?
                        /// ------------------------------------------------------

                        Quest.activeMission.forKey("active").count / Quest.totalMission.forKey("total").count
                    )
                    .forKey(CodingKeys.progress)
            )
            .asRequest(of: QuestProgress.self)

    }

}

```

Is there a way support this at the moment? 

Or should I annotate with the two separate counts and then use a computed property to calculate progress?

Many thanks for your help in advance!

groue:
This pull request adds built-in support for the SQLite [`CAST`](https://www.sqlite.org/lang_expr.html#castexpr) function. The `cast(_:as:)` Swift function accepts expressions, and association aggregates.

```swift
cast(Column("wins"), as: .real)
cast(Player.games.count, as: .real)
```

Fixes #1514

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,7 @@ GRDB adheres to [Semantic Versioning](https://semver.org/), with one exception: 
 - **New**: [#1503](https://github.com/groue/GRDB.swift/pull/1503) by [@simba909](https://github.com/simba909): Conform Database.ColumnType to Sendable
 - **New**: [#1510](https://github.com/groue/GRDB.swift/pull/1510) by [@groue](https://github.com/groue): Add Sendable conformances and unavailabilities
 - **New**: [#1511](https://github.com/groue/GRDB.swift/pull/1511) by [@groue](https://github.com/groue): Database schema dump
+- **New**: [#1515](https://github.com/groue/GRDB.swift/pull/1515) by [@groue](https://github.com/groue): Support for the CAST SQLite function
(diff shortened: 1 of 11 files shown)


leotumwattana:
There is a more complex case where using `cast` will not work.

Please see: https://github.com/groue/GRDB.swift/issues/1514#issuecomment-2016600794

PR #1661, opened 2024-11-12
Foundation: elide extensions on Windows to match Linux

compnerd:
Similar to Linux, elide some extensions on Windows. This is required to build this module on Windows.

PR #1662, opened 2024-11-12
Core: import `ucrt` on Windows for `strcmp`

compnerd:
This uses the deprecated `strcmp` interface on Windows, which requires the C library to be imported. Add the missing import to allow building on Windows.


groue:
Thanks for this and #1661! 👍

groue:
This PR has shipped in [v7.0.0-beta.7](https://github.com/groue/GRDB.swift/tree/v7.0.0-beta.7).

Issue #1325, opened 2023-02-06
quotedDatabaseIdentifier doesn't escape double quotes


EvanHahn-Signal:
### What did you do?

Tried to create a table with a double quote in its name.

Specifically, this fails:

```swift
try db.create(table: "\"") { table in
    table.column("color", .text)
}
```

The above example is a table name that's just one double quote (`"`). It also fails, in different ways, if the table has double quotes. For example, `"foo"` also breaks.

### What did you expect to happen?

The table would be created, because SQLite accepts it.

The following is valid in SQLite:

```sql
CREATE TABLE '"' (color TEXT);
```

It's also valid if I write raw SQL with GRDB:

```swift
try db.execute(sql: "CREATE TABLE '\"' (color TEXT)")
```

However, using GRDB's syntax fails to escape the quote correctly.

### What happened instead?

I got a SQLite syntax error because it tried to run the following query:

```sql
CREATE TABLE """ ("color" TEXT);
```

### Environment

**GRDB flavor(s):** SQLite
**GRDB version:** 6.6.1 (also happens with 5.26.0)
**Installation method:** SPM
**Xcode version:** 14.1
**Swift version:** 5.7.1
**Platform(s) running GRDB:** macOS
**macOS version running Xcode:** 13.2

### Sample code

```swift
import Foundation
import GRDB

let dbQueue = try DatabaseQueue()

try dbQueue.write { db in
    // This doesn't work:
    try db.create(table: "\"") { table in
        table.column("color", .text)
    }

    // This works instead:
    // try db.execute(sql: "CREATE TABLE '\"' (color TEXT)")

    try db.execute(sql: "INSERT INTO '\"' (color) VALUES ('Red')")
    try db.execute(sql: "INSERT INTO '\"' (color) VALUES ('Blue')")

    let names = try String.fetchAll(db, sql: "SELECT color FROM '\"'")
    print(names)
}
```

### Additional info

This may be out of scope for GRDB. Feel free to close if that's the case.

groue:
Hello @EvanHahn-Signal,

Indeed the escaping algorithm is pretty basic: it just wraps identifiers in double quotes. The goal is to help users use SQL keywords as table names: `CREATE TABLE "table"`, `"index"`, `"check"`, `"case"`, `"constraint"`, `"group"`, etc.

Identifiers such as `"` are quite not expected in a regular database schema. Or do they?

groue:
[...]

PR #1707, opened 2025-01-26
Escape quotes inside identifiers when quoting strings

barnettben:
The utility extension `quotedDatabaseIdentifier` wraps a string in double quotes, but doesn't do anything with double quotes already inside the string. This commit escapes existing quotes by repeating them:

```diff
- identifier
+ "identifier"

- ident_"with"_quotes
+ "ident_""with""_quotes"
```

It also adds a couple of simple tests.


groue:
Thank you @barnettben. Do you have a motivating use case?

groue:
Note that there exists a previous issue, which was tagged as "Won't fix": https://github.com/groue/GRDB.swift/issues/1325

barnettben:
Hi @groue. I'm sorry, I'd missed that other issue.

This is such an edge case that I'm not keen on pushing for it. I don't think it's a good practice and agree with your assessment in the linked issue that GRDB should not support it.

I made this PR because I'd come across a database that had quotes in a table name and thought it was a quick job to fix. Cheerfully withdrawn!

Oh, I also learned that this is valid 😅:
`CREATE TABLE ""("" "");`

groue:
Thanks for taking previous arguments in consideration :-)

PR #1708, opened 2025-01-26
Rework import order for SQLCipher; add test conditionals for Android

marcprux:
This is a less ambitious alternative to https://github.com/groue/GRDB.swift/pull/1701 that enables adding a SQLCipher source dependency to the Package.swift based on the `GRDBCIPHER` environment variable. Tests can be run with the dependency with a command like:

```
GRDBCIPHER="https://github.com/skiptools/swift-sqlcipher.git#1.2.1" swift test
```

Running it without the environment variable will leave things exactly as they are. The majority of changes in this PR are simply re-ordering the import header checks for `GRDBCIPHER` and `SWIFT_PACKAGE`, since we would need the former to take precedence of the latter.

An advantage of doing it this way is that it facilitates creating a `GRDB-SQLCipher` fork that can follow the releases of the upstream. The only change in such a fork would be to hardwire the `GRDBCIPHER` setting in Package.swift to some SQLCipher source repository (either mine, or one that you maintain). When a client package wants to switch between GRDB and GRDB-SQLCipher, they would just need to swap their dependency URL.

This PR also adds in support building and testing against Android, mostly by stubbing out the unit tests that can't compile due to missing `NSFileCoordinator`, `Combine`, and the like. Android tests can be run with [skip](https://skip.tools/docs/native/):

```
GRDBCIPHER="https://github.com/skiptools/swift-sqlcipher.git#1.2.1" skip android test
```

Most of the tests pass, but there are a few that need to be investigated. I'll look into them next.

Closes https://github.com/groue/GRDB.swift/pull/1701

### DIFF
--- a/GRDB/Core/StatementAuthorizer.swift
+++ b/GRDB/Core/StatementAuthorizer.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
@@ -11,6 +11,8 @@ import SQLite3
 import string_h
 #elseif os(Linux)
 import Glibc
+#elseif canImport(Android)
+import Android
 #elseif os(macOS) || os(iOS) || os(watchOS) || os(tvOS) || os(visionOS)
 import Darwin
 #elseif os(Windows)
--- a/GRDB/Dump/Database+Dump.swift
+++ b/GRDB/Dump/Database+Dump.swift
@@ -1,4 +1,7 @@
 import Foundation
+#if GRDBCIPHER
+import SQLCipher
+#endif
 
 // MARK: - Dump
 
--- a/Package.swift
+++ b/Package.swift
@@ -27,6 +27,24 @@ if ProcessInfo.processInfo.environment["SPI_BUILDER"] == "1" {
     dependencies.append(.package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"))
 }
 
+var targetDependencies: [Target.Dependency] = ["GRDBSQLite"]
+
+
+var GRDBCIPHER = ProcessInfo.processInfo.environment["GRDBCIPHER"]
+// e.g.:
+//GRDBCIPHER="https://github.com/skiptools/swift-sqlcipher.git#1.2.1"
+if let SQLCipherRepo = GRDBCIPHER?.split(separator: "#").first,
+    let SQLCipherVersion = GRDBCIPHER?.split(separator: "#").last,
+    let SQLCipherRepoURL = URL(string: SQLCipherRepo.description) {
+    swiftSettings.append(.define("GRDBCIPHER"))
+    targetDependencies = [.product(name: "SQLCipher", package: SQLCipherRepoURL.deletingPathExtension().lastPathComponent)]
+    if let version = Version(SQLCipherVersion.description) { // numeric version
+        dependencies.append(.package(url: SQLCipherRepoURL.absoluteString, from: version))
+    } else { // branch
+        dependencies.append(.package(url: SQLCipherRepoURL.absoluteString, branch: SQLCipherVersion.description))
+    }
+}
+
 let package = Package(
     name: "GRDB",
     defaultLocalization: "en", // for tests
@@ -48,7 +66,7 @@ let package = Package(
             providers: [.apt(["libsqlite3-dev"])]),
         .target(
             name: "GRDB",
-            dependencies: ["GRDBSQLite"],
+            dependencies: targetDependencies,
             path: "GRDB",
             resources: [.copy("PrivacyInfo.xcprivacy")],
             cSettings: cSettings,
--- a/Tests/GRDBTests/DatabaseQueueTests.swift
+++ b/Tests/GRDBTests/DatabaseQueueTests.swift
@@ -2,6 +2,7 @@ import XCTest
 import Dispatch
 import GRDB
 
+#if canImport(Darwin) // needed for __dispatch_get_global_queue
 class DatabaseQueueTests: GRDBTestCase {
     func testJournalModeConfiguration() throws {
         do {
@@ -473,3 +474,4 @@ class DatabaseQueueTests: GRDBTestCase {
         dbQueue.releaseMemory()
     }
 }
+#endif
--- a/Tests/GRDBTests/FTS5TableBuilderTests.swift
+++ b/Tests/GRDBTests/FTS5TableBuilderTests.swift
@@ -1,6 +1,9 @@
 #if SQLITE_ENABLE_FTS5
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
 
 class FTS5TableBuilderTests: GRDBTestCase {
     func testWithoutBody() throws {
--- a/Tests/GRDBTests/FailureTestCase.swift
+++ b/Tests/GRDBTests/FailureTestCase.swift
@@ -1,6 +1,7 @@
 // Inspired by https://github.com/groue/CombineExpectations
 import XCTest
 
+#if canImport(Darwin) // needed for XCTIssue
 /// A XCTestCase subclass that can test its own failures.
 class FailureTestCase: XCTestCase {
     private struct Failure: Hashable {
@@ -207,3 +208,4 @@ class FailureTestCaseTests: FailureTestCase {
         }
     }
 }
+#endif
--- a/Tests/GRDBTests/FoundationNSDecimalNumberTests.swift
+++ b/Tests/GRDBTests/FoundationNSDecimalNumberTests.swift
@@ -1,6 +1,7 @@
 import XCTest
 import GRDB
 
+#if canImport(Darwin) // needed for NSDecimalNumber
 class FoundationNSDecimalNumberTests: GRDBTestCase {
 
     func testNSDecimalNumberPreservesIntegerValues() {
@@ -157,3 +158,4 @@ class FoundationNSDecimalNumberTests: GRDBTestCase {
         try test("18446744073709551615", isDecodedAs: NSDecimalNumber(string: "18446744073709551615"))
     }
 }
+#endif
--- a/Tests/GRDBTests/FoundationNSNumberTests.swift
+++ b/Tests/GRDBTests/FoundationNSNumberTests.swift
@@ -1,6 +1,7 @@
 import XCTest
 import GRDB
 
+#if canImport(Darwin) // needed for NSDecimalNumber
 class FoundationNSNumberTests: GRDBTestCase {
     
     func testNSNumberDatabaseValueToSwiftType() {
@@ -192,3 +193,4 @@ class FoundationNSNumberTests: GRDBTestCase {
         try test("18446744073709551615", isDecodedAs: NSDecimalNumber(value: UInt64(18446744073709551615)))
     }
 }
+#endif
--- a/Tests/GRDBTests/FoundationNSURLTests.swift
+++ b/Tests/GRDBTests/FoundationNSURLTests.swift
@@ -1,6 +1,7 @@
 import XCTest
 import GRDB
 
+#if canImport(Darwin) // needed for NSURL
 class FoundationNSURLTests: GRDBTestCase {
     
     func testNSURLDatabaseRoundTrip() throws {
@@ -46,3 +47,4 @@ class FoundationNSURLTests: GRDBTestCase {
     }
     
 }
+#endif
--- a/Tests/GRDBTests/JSONColumnTests.swift
+++ b/Tests/GRDBTests/JSONColumnTests.swift
@@ -1,5 +1,8 @@
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
 
 final class JSONColumnTests: GRDBTestCase {
     func test_JSONColumn_derived_from_CodingKey() throws {
--- a/Tests/GRDBTests/JSONExpressionsTests.swift
+++ b/Tests/GRDBTests/JSONExpressionsTests.swift
@@ -1,5 +1,8 @@
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
 
 final class JSONExpressionsTests: GRDBTestCase {
     func test_Database_json() throws {
--- a/Tests/GRDBTests/MutablePersistableRecordTests.swift
+++ b/Tests/GRDBTests/MutablePersistableRecordTests.swift
@@ -1,5 +1,8 @@
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
 
 private struct MutablePersistableRecordPerson : MutablePersistableRecord {
     var id: Int64?
--- a/Tests/GRDBTests/PersistableRecordTests.swift
+++ b/Tests/GRDBTests/PersistableRecordTests.swift
@@ -1,5 +1,8 @@
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
 
 private struct PersistableRecordPerson : PersistableRecord {
     var name: String?
--- a/Tests/GRDBTests/ValueObservationRecorderTests.swift
+++ b/Tests/GRDBTests/ValueObservationRecorderTests.swift
@@ -1,6 +1,7 @@
 import Dispatch
 import XCTest
 
+#if canImport(Darwin) // needed for FailureTestCase
 class ValueObservationRecorderTests: FailureTestCase {
     // MARK: - NextOne
     
@@ -781,3 +782,4 @@ class ValueObservationRecorderTests: FailureTestCase {
         }
     }
 }
+#endif
--- a/Tests/GRDBTests/ValueObservationTests.swift
+++ b/Tests/GRDBTests/ValueObservationTests.swift
@@ -1,7 +1,11 @@
 import XCTest
 import Dispatch
 @testable import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
 
+#if canImport(Darwin)
 class ValueObservationTests: GRDBTestCase {
     // Test passes if it compiles.
     // See <https://github.com/groue/GRDB.swift/issues/1541>
@@ -1353,3 +1357,4 @@ class ValueObservationTests: GRDBTestCase {
                 })
     }
 }
+#endif
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -73,6 +73,23 @@ jobs:
       - uses: actions/checkout@v4
       - name: ${{ matrix.name }}
         run: make test_SPM test_install_SPM
+  SPMSQLCipher:
+    name: SPM
+    runs-on: ${{ matrix.runsOn }}
+    env:
+      DEVELOPER_DIR: "/Applications/${{ matrix.xcode }}/Contents/Developer"
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - xcode: "Xcode_16.1.app"
+            runsOn: macOS-14
+            name: "Xcode 16.1"
+    steps:
+      - uses: actions/checkout@v4
+      - name: ${{ matrix.name }}
+        run: GRDBCIPHER="https://github.com/skiptools/swift-sqlcipher.git#1.2.1" swift test
   SQLCipher3:
     name: SQLCipher3
     runs-on: ${{ matrix.runsOn }}
--- a/GRDB/Core/Configuration.swift
+++ b/GRDB/Core/Configuration.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/Database+Schema.swift
+++ b/GRDB/Core/Database+Schema.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/Database+Statements.swift
+++ b/GRDB/Core/Database+Statements.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/Database.swift
+++ b/GRDB/Core/Database.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/DatabaseCollation.swift
+++ b/GRDB/Core/DatabaseCollation.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/DatabaseError.swift
+++ b/GRDB/Core/DatabaseError.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/DatabaseFunction.swift
+++ b/GRDB/Core/DatabaseFunction.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/DatabaseSnapshotPool.swift
+++ b/GRDB/Core/DatabaseSnapshotPool.swift
@@ -1,9 +1,9 @@
 #if SQLITE_ENABLE_SNAPSHOT || (!GRDBCUSTOMSQLITE && !GRDBCIPHER)
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/DatabaseValue.swift
+++ b/GRDB/Core/DatabaseValue.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/Row.swift
+++ b/GRDB/Core/Row.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
--- a/GRDB/Core/RowDecodingError.swift
+++ b/GRDB/Core/RowDecodingError.swift
@@ -1,8 +1,8 @@
 // Import C SQLite functions
-#if SWIFT_PACKAGE
-import GRDBSQLite
-#elseif GRDBCIPHER
+#if GRDBCIPHER
 import SQLCipher
+#elseif SWIFT_PACKAGE
+import GRDBSQLite
 #elseif !GRDBCUSTOMSQLITE && !GRDBCIPHER
 import SQLite3
 #endif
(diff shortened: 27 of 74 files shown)


marcprux:
> @marcprux the Android NDK doesn't seem to bundle sqlite3.h, whereas GRDB and your PR do rely on that being available. How are you providing that right now when building and testing GRDB?

It is included with the SQLCipher source module: https://github.com/skiptools/swift-sqlcipher/tree/main/Sources/SQLCipher/sqlite

Tests/GRDBTests/DatabaseQueueTests.swift
```diff
@@ -2,6 +2,7 @@ import XCTest
 import Dispatch
 import GRDB
 
+#if canImport(Darwin) // needed for __dispatch_get_global_queue
```
groue, line 5: I understand that this pull request is a draft, and you were not aiming at polishing it.

I'm thus sure you agree that the tests that do not use `__dispatch_get_global_queue` should keep on running.

Usually, I keep all tests, and disable them with `XCTSkip`, in a way or another:

```swift
func test_foo() throws {
#if condition
    // the test
#else
    throw XCTSkip("XXX is not available")
#endif
}
```

This will be needed before we merge.

- DatabaseMigratorTests
- DatabasePoolConcurrencyTests
- DatabaseQueueTests
- DatabaseRegionObservationTests
- DatabaseSnapshotTests

I spare you the tests that are not intended to run and are only testing if some code is accepted by the compiler ("Test passes if it compiles" in DatabaseRegionObservationTests et al.)

Tests/GRDBTests/FTS5TableBuilderTests.swift
```diff
@@ -1,6 +1,9 @@
 #if SQLITE_ENABLE_FTS5
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
```
groue, line 6: To be removed if we replace `sqlite3_libversion_number()` with `Database.sqliteLibVersionNumber`.

Tests/GRDBTests/FailureTestCase.swift
```diff
@@ -1,6 +1,7 @@
 // Inspired by https://github.com/groue/CombineExpectations
 import XCTest
 
+#if canImport(Darwin) // needed for XCTIssue
```
groue, line 4: `FailureTestCase` is the base class of `ValueObservationRecorderTests`, which tests `ValueObservationRecorder`. `ValueObservationRecorder`, on its turn, is used to test `ValueObservation`. And `ValueObservation` is a GRDB piece that must remain super reliable (i.e. tested very aggressively).

We'll need to find a way to restore `ValueObservationRecorder` tests.

For the time being, let's restore this class, and have `assertFailure` throw `XTCSkip` in environments that don't have XCTIssue (replacing `rethrows` with `throws`).

Tests/GRDBTests/JSONColumnTests.swift
```diff
@@ -1,5 +1,8 @@
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
```
groue, line 5: To be removed if we replace `sqlite3_libversion_number()` with `Database.sqliteLibVersionNumber`.

Tests/GRDBTests/JSONExpressionsTests.swift
```diff
@@ -1,5 +1,8 @@
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
```
groue, line 5: To be removed if we replace `sqlite3_libversion_number()` with `Database.sqliteLibVersionNumber`.

Tests/GRDBTests/MutablePersistableRecordTests.swift
```diff
@@ -1,5 +1,8 @@
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
```
groue, line 5: To be removed if we replace `sqlite3_libversion_number()` with `Database.sqliteLibVersionNumber`.

Tests/GRDBTests/PersistableRecordTests.swift
```diff
@@ -1,5 +1,8 @@
 import XCTest
 import GRDB
+#if GRDBCIPHER
+import SQLCipher
+#endif
```
groue, line 5: To be removed if we replace `sqlite3_libversion_number()` with `Database.sqliteLibVersionNumber`.

Tests/GRDBTests/ValueObservationRecorderTests.swift
```diff
@@ -1,6 +1,7 @@
 import Dispatch
 import XCTest
 
+#if canImport(Darwin) // needed for FailureTestCase
```
groue, line 4: To be removed once `FailureTestCase` is restored (see above).

groue (commented):
Hello @marcprux,

I like that very much. This pull request is a great gift 🎁! I'll shortly play with the new targets.

I made a first review and preliminary comments. I hope we agree on the broad outlines!

groue:
> The results over a dozen runs are a fairly consistent slowdown of around 10% when using SQLCipher. It'd be interesting to tweak the SQLite build flags in [swift-sqlcipher](https://github.com/skiptools/swift-sqlcipher/blob/main/Package.swift) to see what might increase/decrease the performance.

Isn't it just the cipher?

SQLCipher can also be _extremely slow_ when it opens a connection. Some people have reported up to [0.5 seconds](https://github.com/groue/GRDB.swift/issues/1256#issuecomment-1470992333) (😱). To the point I had to ship #1350.

GRDB/Dump/Database+Dump.swift
```diff
@@ -1,4 +1,7 @@
 import Foundation
+#if GRDBCIPHER
+import SQLCipher
+#endif
```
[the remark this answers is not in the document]
groue: I'll see if I can make this in another pull request.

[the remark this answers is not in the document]
groue: > I'll see if I can make this in another pull request.

I merged #1713.

groue:
I ran the same performance tests, but this time in release configuration:

```
echo "With GRDBCIPHER"; rm -rf .build; GRDB_PERFORMANCE_TESTS=1 GRDBCIPHER="https://github.com/skiptools/swift-sqlcipher.git#1.2.1" swift test --configuration release --filter GRDBPerformanceTests 2>&1 | grep --after-context=2 "Test Suite 'GRDBPackageTests.xctest' passed"
echo "Without GRDBCIPHER"; rm -rf .build; GRDB_PERFORMANCE_TESTS=1 swift test --configuration release --filter GRDBPerformanceTests 2>&1 | grep --after-context=2 "Test Suite 'GRDBPackageTests.xctest' passed"

With GRDBCIPHER
Test Suite 'GRDBPackageTests.xctest' passed at 2025-02-01 13:53:37.700.
	 Executed 23 tests, with 0 failures (0 unexpected) in 72.258 (72.260) seconds
Test Suite 'Selected tests' passed at 2025-02-01 13:53:37.700.
Without GRDBCIPHER
Test Suite 'GRDBPackageTests.xctest' passed at 2025-02-01 13:57:26.264.
	 Executed 23 tests, with 0 failures (0 unexpected) in 80.139 (80.141) seconds
Test Suite 'Selected tests' passed at 2025-02-01 13:57:26.264.
```

This time, SQLCipher tests are faster.

GRDB/Dump/Database+Dump.swift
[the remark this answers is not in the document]
groue: I pushed a rebased version of your branch in the [marcprux-sqlcipher-rebased](https://github.com/groue/GRDB.swift/tree/marcprux-sqlcipher-rebased) branch, along with two minor tweaks (avoid warnings in the GRDBPerformanceTests target, and run the ability to test this target from the Makefile). You can reset your branch to this one, or cherry pick the extra commits if you prefer.

marcprux:
> SQLCipher can also be extremely slow when it opens a connection

Agreed, but the performance hit should only occur when encryption is actually enabled on the database, which shouldn't affect any of the perf tests.

> This time, SQLCipher tests are faster.

Of course! Foolish of me to not do that. Otherwise, it is comparing the debug build of the sqlcipher/sqlite source build against the release build of the vendored sqlite.

It is heartening to see that superior performance can be squeezed out of a source build. I wonder how much more might be managed with strategic performance-related flags (e.g., https://github.com/sbooth/CSQLite/pull/59#issuecomment-2592754630).

GRDB/Dump/Database+Dump.swift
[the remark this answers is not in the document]
marcprux: I've gotten rid of the now-unnecessary imports in [65db92f](https://github.com/groue/GRDB.swift/pull/1708/commits/65db92f512c1b077b8958f226158ed1a3ae3a4b3)

Tests/GRDBTests/FailureTestCase.swift
marcprux: Done in [15e3089](https://github.com/groue/GRDB.swift/pull/1708/commits/15e308950c1f47adcedf4ad80a303b95e022c0f9)

Tests/GRDBTests/DatabaseQueueTests.swift
marcprux: OK, I've switched to throwing `XCTSkip` rather than wholesale #if'ing out the test cases when a required framework is unavailable in [289f9fd].

marcprux:
Apologies for the long delay in addressing the feedback (I was traveling).

I believe I've handled each of the requested changes. Let me know if I can do anything else to help get this PR over the finish line – it has grown quite large, so I'm nervous that rebasing is going to be difficult if it starts to become stale.

Joannis:
We've been using this branch successfully.

marcprux:
I'll follow up on the list of needs at https://mastodon.social/@groue@hachyderm.io/113990450491505387 here, just to keep it in the PR record.

- [x] Provides built-in SQLCipher support that works for 80+% of people.

This should work out of the box for anyone that just changes:

```swift
    dependencies: [
        .package(url: "https://github.com/groue/GRDB.swift.git", from: "7.0.0"),
    ]
```

to the (theoretical) locally-maintained fork:

```swift
    dependencies: [
        .package(url: "https://github.com/groue/GRDB-SQLCipher.swift.git", from: "7.0.0"),
    ]
```

- [x] No extra dependency (we must control the SQLCipher compilation options).

You are welcome to relocate https://github.com/skiptools/swift-sqlcipher.git into the `groue/` org and use that – then you would have control over releases and flags.

- [x] No binaries (they rot).

`swift-sqlcipher.git` is a pure source build. There are no binaries.

- [ ] A clear documentation that tells users what to do to get GRDB+SQLCipher with SPM, written for people who are not build experts.

This isn't part of this PR, but it would only be a couple of lines in the README.

- [x] Testable in GitHub CI.

This is part of the PR. The new `SPMSQLCipher` job just does:

```
GRDBCIPHER="https://github.com/skiptools/swift-sqlcipher.git#1.2.1" swift test
```

- [ ] Runs the full GRDB test suite with unencrypted databases, AND with encrypted databases (we already do that for SQLCipher with CocoaPods).

The tests do run against SQLCipher, but against an unencrypted database. I didn't see how this is being done with SQLCipher for CocoaPods, but I suppose we could just do the same setup here (perhaps just based on checking the `GRDBCIPHER` environment variable)

- [x] Requires minimum efforts for debugging (one can debug tests from Xcode).

This shouldn't be an issue. Just activate (i.e., un-comment) the `GRDBCIPHER` line in the `Package.swift` and run the test cases in Xcode.

- [x] Requires minimum efforts for maintenance (i.e. bumping SQLCipher version and adjusting compile options when needed)

For the SQLCipher dependency, you just need to make a new amalgamated sqlite.c from sqlcipher, commit it, and then tag a release. This is done with the script: https://github.com/skiptools/swift-sqlcipher/blob/main/scripts/build_sqlcipher.sh

An example of this sort of update commit is the bump to sqlite 3.46.1 and sqlcipher 4.6.1: https://github.com/skiptools/swift-sqlcipher/commit/5ea2f8b05e6943bcf021e54f8a25741c691f7db4

For GRDB itself, my proposal is that you maintain a fork of `GRDB.swift.git` at `https://github.com/groue/GRDB-SQLCipher.swift.git`. The fork is pristine, with the one exception that the `GRDBCIPHER` line in the `Package.swift` is un-commented. Whenever you make a release of `GRDB.swift.git`, you just update the `GRDB-SQLCipher.swift.git` fork and sync the tags.

It is a little bit of overhead, but it could be added as another step in any scripts you might already be using to maintain releases.

- [ ] A technique thats enables SQLCipher support in companion packages as well (GRDBQuery, GRDBSnapshotTesting, RxGRDB).

This is certainly a shortcoming of this method. I unfortunately don't see any solution other than to also have a `GRDBQuery-SQLCipher` that depends on the `GRDB-SQLCipher.swift.git` fork.

This last point is indeed painful. However, I think this PR gets us a lot, with minimal intrusion on the existing GRDB structure. A more "correct" solution would be along the lines of https://github.com/groue/GRDB.swift/pull/1701, which requires a large refactoring and had performance implications, and so was rejected.

We can have a working GRDB-SQLCipher implementation (plus Android support!) _today_ with this PR, while still continuing to design a more ideal future implementation.

Alex293:
This PR saved me today while integrating a closed source framework which links SQLCipher while we were already extensively using grdb.

marcprux:
I've updated this PR to use the new [package traits](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md) in Swift 6.1, which seem designed for a scenario exactly like this. Now, rather than having to create and maintain a custom GRDB-SQLCipher fork that enables GRDBCIPHER, a consumer of this package can simply enable the `GRDBCIPHER` trait when adding a dependency on GRDB:

```
    dependencies: [
        .package(url: "https://github.com/groue/GRDB.swift", from: "7.5.0", traits: ["GRDBCIPHER"])
    ]
```

Adding the trait to the dependency will activate the `GRDBCIPHER` conditionals as well as adding a conditional dependency on the `swift-sqlcipher` repository, whereas leaving the trait empty (the default) will result in GRDB continuing to use the vendored SQLite3.

For the purpose of testing and CI, the "GRDBCIPHER" environment variable is still checked, but only to see whether to enable the trait by default in the `Package.swift`, so you can run:

```
# test with SQLCipher
GRDBCIPHER=1 swift test 
# test with SQLite3
GRDBCIPHER=0 swift test 
```

The performance test environment can also still be used, which demonstrates that the custom SQLCipher build continues to be ~10% faster than the vendored SQLite3, at least on my machine:

```
# SQLCipher performance tests
GRDBCIPHER=1 GRDB_PERFORMANCE_TESTS=1 swift test -c release --filter GRDBPerformanceTests 
# Executed 23 tests, with 0 failures (0 unexpected) in 73.005 (73.009) seconds

# SQLite3 performance tests
GRDBCIPHER=0 GRDB_PERFORMANCE_TESTS=1 swift test -c release --filter GRDBPerformanceTests 
# Executed 23 tests, with 0 failures (0 unexpected) in 81.398 (81.401) seconds
```


The only drawback is that since traits are simple booleans, you can no longer specify a custom repository with the "GRDBCIPHER" environment variable, so you need to depend on a specific repository. But an advantage is that the `swift-sqlcipher.git` repository could now publish its own traits that will control various SQLite build flags, which means that consumers of the package could potentially have fine-grained control over the SQLite built itself.

marcprux:
@groue Hoping you can take another look at this PR with the new Package traits dependency. I think it really improves on the original idea.

marcprux:
We need this for our work, so I'll maintain GRDB with SQLCipher support in our fork at https://github.com/swift-everywhere/grdb-sqlcipher.git and track upstream releases, starting with v7.5.0.

aehlke:
@groue We'd love to see this get merged. It would be very useful for using with Skip and other native Android Swift usage. Is it waiting on you finding time for review or are there further changes needed? Thank you

groue:
Hello, sorry for being late to the party. I'm slow on those topics because I'm really not excited by build topics.

Let's discuss traits!

The [Trait unification](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md#trait-unification) chapter of the proposal says: 

> At this point, it is important to talk about the trait unification across the entire dependency graph. After dependency resolution the union of enabled traits per package is calculated. This is then used to determine both the enabled optional dependencies and the enabled traits for the compile time checks. Since the enabled traits of a dependency are specified on a per package level and not from the root of the tree, any combination of enabled traits must be supported. A consequence of this is that all traits **should** be *additive*. Enabling a trait **should not** disable functionality i.e. remove API or lead to any other **SemVer-incompatible** change.

A SQLCipher trait can not be 100% compatible with the SQLite version that ships on Apple operating systems (the default trait on Apple OS). We just have to wait for the two SQLite version to ship with a different set of compile options, or for the operating system to ship a newer SQLite version with more features.

In case the problem is not clear, we just need:

- An application that depends on Dependency1 and Dependency2
- Dependency1 depends on GRDB(standard) and uses an API X (does not need to be a GRDB API, it can be a SQLite C function, or an SQL syntax).
- Dependency2 depends, or starts depending on GRDB(+SQLCipher)

With bad luck, API X does not work with SQLCipher, and has no replacement (linker error, runtime error, whatever, SOME ERROR). The application owner complains to the Dependency1 maintainers, who say they're not responsible (they are correct). Owner complains to Dependency2, who say they require SQLCipher (perfectly legit). What are the options of the application owner? All painful. They are entitled to turn back to this repository, complaining that this is a misuse of package traits (and they would be correct: GRDB **should not** have exposed SQLCipher as a trait, to reuse the emphasized terminology of the proposal).

And I'm not even talking about the `import SQLite3` that the app and Dependency1 are entitled to use, with bad consequences whenever SQLCipher enters the picture.

Thoughts?

marcprux:
A diamond dependency pattern does indeed have the potential to introduce merged trait conflicts. But I envision that the preference for which SQLite runtime to use (vendored SQLite3 vs. custom SQLCipher) would be something generally decided at the application level, and so the trait declarations would be pushed up as high as possible. Dependency1 and Dependency2 ought not mandate any particular compile options, but instead adapt to the SQLite runtime in use, such as by introspecting on `pragma COMPILE_OPTIONS` (and potentially throwing a runtime error if there is an unsupported operation). E.g.: 

```swift
func encryptDatabase() throws {
    if try exec("pragma COMPILE_OPTIONS").contains("SQLCIPHER") == false { 
        throw RuntimeError("Unsupported database configuration…") 
    } 
}
```

I'm not sure how many middleware packages like this are out there currently, other than [sharing-grdb](https://github.com/pointfreeco/sharing-grdb). If you know of any, it would be interesting to examine them to see if they presume any particular compile options.

Also, I'll note that the potential issue of compile option mismatch _already_ exists on Linux: the `.apt(["libsqlite3-dev"])` dependency in use will likely have different compile options on different distributions, and these won't generally match the exact compile options on Apple platforms (which aren't even themselves consistent: macOS and iOS have different compile options, and I wouldn't be surprised if tvOS, visionOS, etc. also have mismatches). And lastly, these options can and do change in between releases. So, in short, I think it makes sense to advocate for defensive programming and validating any compile option assumptions with a check against the runtime `COMPILE_OPTIONS`.

> And I'm not even talking about the import SQLite3 that the app and Dependency1 are entitled to use, with bad consequences whenever SQLCipher enters the picture.

Yes, well, that would have to be discouraged. If GRDB were to re-export the SQLite3 or SQLCipher module, then a dependent package ought to just be able to pick up the symbols from there without needing to ever `import SQLite3`. This is somewhat orthogonal to the traits technique I'm suggesting; any world in which GRDB might depend on SQLCipher vs. SQLite3 — regardless of how it is implemented — will have this issue.

groue:
Thanks @marcprux for your careful consideration. This is well appreciated.

What I envision is that users wildly exploit any possible technique. That's my rule of thumb, and it's been quite efficient so far.

I do not assume that the preference will be decided at the application level. It is reasonable to imagine a third-party library that encrypts its private data and sets the SQLCipher trait. 

We do not have to wonder if conflicts can happen: they will. And runtime checks can't fix linker errors.

The correct attitude is vigilant optimism - and probably a fair amount of advice and empathy in the documentation.

---

On the topic of exposing the C SQLite APIs, which is an important GRDB feature - several people depend on it:

> If GRDB were to re-export the SQLite3 or SQLCipher module, then a dependent package ought to just be able to pick up the symbols from there without needing to ever import SQLite3. 

That's what GRDB was doing until GRDB 7. There were too many [reports of build errors](https://github.com/groue/GRDB.swift/issues?q=CSQLite), so I shipped #1600, forcing users to write explicit imports of their desired SQLite flavor. Since then, the flow of build error reports has nearly stopped. As far as I can tell, the root cause is probably an Xcode bug.

What I mean here is that I can't just re-export and call it a day. In an ideal world, this would work. But it does not, for reasons I can't address, and I'm unable to report (the error does happen, I saw it, but it is very difficult to reproduce, and it's not only a matter of cleaning caches).

---

Finally, there's the ["official" SQLCipher package](https://github.com/sqlcipher/sqlcipher/issues/371#issuecomment-3156623200), on which you have made several comments. If I understand well, the big problem is that a binary distribution would not work on all platforms, and that's why you suggest a source distribution. I agree with you.

Why did you pick https://github.com/skiptools/swift-sqlcipher? What are the other options? If there's a bias due to your focus on Android, would you please make it explicit, and describe the consequences of this choice for other platforms?

---

So… Using package traits for SQLCipher will create a new flow of issues.

And this is not a subject I know enough about. I do not want open issues to pile up, especially when the issue lies elsewhere, and I'm unable to prove it. I could move all those issues to discussions (some repos do that in order to keep the number of issues small), but that does not address anything. Finally I'm slow to address those topics, as everybody can see.

In the interest of all, I wish someone would volunteer for the position of “build support” in GRDB.

marcprux:
> Why did you pick https://github.com/skiptools/swift-sqlcipher? What are the other options?

I picked it simply because that is a package that I had already put together to work on Android. Any buildable SQLCipher amalgamation would work, and as I've mentioned before, you could simply pull it into the GRDB package as a C module in order to eliminate the need for an external dependency (which I think would be the ideal solution). I would be happy to expand this PR to provide that option, but it would introduce the potential maintenance burden of keeping up with subsequent SQLCipher releases, which I sense you aren't interested in.

> If there's a bias due to your focus on Android, would you please make it explicit, and describe the consequences of this choice for other platforms?

You would be correct in thinking that the two issues (SQLCipher support and Android support) are orthogonal. They just happen to both be solvable through this PR.

My ultimate goal is to bring this support to all platforms, including Android. GRDB currently doesn't work _at all_ on Android due to there not being any linkable sqlite library on the platform (unlike Linux), so a source build is currently the only solution for that platform (irrespective of whether the source build is sqlcipher or just a stock sqlite package like https://github.com/sbooth/CSQLite or https://github.com/swiftlang/swift-toolchain-sqlite).

As to your other concerns, I feel like they can all be mitigated by simply stating that this support is experimental and optional (just like the current Linux support). Nothing this PR does changes how GRDB functions on Apple platforms by default. It is only when opting-into the SQLCipher trait that the new build behavior would be activated.

I'm happy to continue iterating on the PR or provide alternate suggestions/solutions, but I also understand if you would prefer GRDB to stick to narrow path of supporting only the vendored-SQLite on Apple platforms.

groue:
All right, @marcprux, we are converging. I agree that we can just flag SQLCipher support as experimental. I'll have a closer look at the PR.

groue:
(EDIT: in response to a comment that was deleted by its author)

I'm not sure the meaning of "experimental" was correctly understood. Expressions such as "huge deal" and "intolerable" can only be supported by actual contributions.

_Il n’y a pas d’amour, il n’y a que des preuves d’amour_ ([the proof is in the pudding](https://english.stackexchange.com/a/518549)).

---

EDIT: Given the complete lack of mention of GRDB in the [SQLiteData](https://github.com/pointfreeco/sqlite-data/tree/1.0.0) README, I would advise commenters to choose their arguments with more tact and discernement than some of their favorite developers.

mbrandonw:
> EDIT: Given the complete lack of mention of GRDB in the [SQLiteData](https://github.com/pointfreeco/sqlite-data/tree/1.0.0) README, I would advise commenters to choose their arguments with more tact and discernement than some of their favorite developers.

This was an honest mistake on our part and we saw your comment on Hacker News and immediately corrected it. We took for granted the obviousness of GRDB’s presence in the library in the name alone, and so forgot to update the readme with the 1.0 release. We are very sorry for that.

Tests/GRDBTests/DatabaseErrorTests.swift
```diff
@@ -216,6 +216,9 @@ class DatabaseErrorTests: GRDBTestCase {
     }
     
     func testNSErrorBridging() throws {
+        #if !canImport(ObjectiveC)
+        throw XCTSkip("NSError bridging works differently on non-Darwin platforms")
```
groue, line 220: Same question as above

Tests/GRDBTests/FoundationNSDecimalNumberTests.swift
```diff
@@ -3,8 +3,10 @@ import GRDB
 
 class FoundationNSDecimalNumberTests: GRDBTestCase {
 
-    func testNSDecimalNumberPreservesIntegerValues() {
-        
+    func testNSDecimalNumberPreservesIntegerValues() throws {
+        #if !canImport(ObjectiveC)
+        throw XCTSkip("NSDecimalNumber unavailable")
```
groue, line 8: NSDecimalNumber is part of the [GRDB support for decimals](https://github.com/groue/GRDB.swift/blob/master/README.md#nsnumber-nsdecimalnumber-and-decimal).

I find mentions of NSDecimalNumber in https://github.com/swiftlang/swift-corelibs-foundation and https://github.com/swiftlang/swift-foundation.

Is this lack of NSDecimalNumber temporary, or Android specific? Shouldn't we use `#if os(android)` instead of `#if !canImport(ObjectiveC)`?

Finally, shouldn't we disable the whole test class, since it is named Foundation**NSDecimalNumber**Tests?

Tests/GRDBTests/FoundationNSNumberTests.swift
```diff
@@ -3,7 +3,10 @@ import GRDB
 
 class FoundationNSNumberTests: GRDBTestCase {
     
-    func testNSNumberDatabaseValueToSwiftType() {
+    func testNSNumberDatabaseValueToSwiftType() throws {
+        #if !canImport(ObjectiveC)
+        throw XCTSkip("NSNumber unavailable")
```
groue, line 8: Same question as above about NSDecimalNumber.

Tests/GRDBTests/FoundationNSURLTests.swift
```diff
@@ -4,6 +4,9 @@ import GRDB
 class FoundationNSURLTests: GRDBTestCase {
     
     func testNSURLDatabaseRoundTrip() throws {
+        #if !canImport(ObjectiveC)
+        throw XCTSkip("NSURL unavailable")
```
groue, line 8: `NSURL` is less useful than `NSNumber`/`NSDecimalNumber`, so here I'll just recommend that the whole test class is disabled on platforms that do not support NSURL.

Tests/GRDBTests/FoundationURLTests.swift
```diff
@@ -42,7 +42,7 @@ class FoundationURLTests: GRDBTestCase {
         XCTAssertNil(URL.fromDatabaseValue(databaseValue_Null))
         XCTAssertNil(URL.fromDatabaseValue(databaseValue_Int64))
         XCTAssertNil(URL.fromDatabaseValue(databaseValue_Double))
-        XCTAssertEqual(URL.fromDatabaseValue(databaseValue_Blob)!.absoluteString, "bar")
+        XCTAssertEqual(URL.fromDatabaseValue(databaseValue_Blob)?.absoluteString, "bar")
```
groue, line 45: It looks like a change that is not strictly necessary.

Tests/GRDBTests/ValueObservationTests.swift
```diff
@@ -890,7 +890,8 @@ class ValueObservationTests: GRDBTestCase {
         try test(makeDatabaseQueue())
         try test(makeDatabasePool())
     }
-    
+
+    #if canImport(Darwin) // @MainActor test cases don't compile non-Darwin platforms: "call to main actor-isolated instance method 'test_mainActor_observation()' in a synchronous nonisolated(unsafe) context"
```
groue, line 894: This needs a link to at least on issue in the Swift repository.

GRDB/Core/StatementAuthorizer.swift
```diff
@@ -11,6 +11,8 @@ import SQLite3
 import string_h
 #elseif os(Linux)
 import Glibc
+#elseif canImport(Android)
+import Android
```
groue, line 15: I have no issue here, just a curiosity. Why is this import necessary?

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:6.0
+// swift-tools-version:6.1
```
groue, line 1: Not required in this pull request, but some cleanup is needed:

- `git grep -i 'version.*\b6\.0'`
- `git grep 'compiler('`
- `README.md`
- `GRDB.swift.podspec`

Package.swift
```diff
@@ -41,14 +47,18 @@ let package = Package(
         .library(name: "GRDB", targets: ["GRDB"]),
         .library(name: "GRDB-dynamic", type: .dynamic, targets: ["GRDB"]),
     ],
+    traits: [
+        .trait(name: "GRDBCIPHER", description: "Use the SQLCipher library rather than the vendored SQLite"),
```
groue, line 51: After reading the naming concerns expressed in https://github.com/groue/GRDB.swift/issues/1816, and the trait names in [SE-0450](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md), I suggest using the name `SkipSQLCipher`, and a description that mentions the repository URL https://github.com/skiptools/swift-sqlcipher

groue: We're about to have multiple SQLCipher versions: the one in this pull request, the one from #1816, and the CocoaPods one (even if it's high time I completely removed it).

To simplify the compiler flags dance, I start working on replacing most uses of `#if GRDBCIPHER` with `#if SQLITE_HAS_CODEC`. That last flag ought to be set by all GRDB+SQLCipher flavors.

cc @R4N

Tests/GRDBTests/DatabaseErrorTests.swift
thinkpractice: Again, I'm having a similar issue on Linux. The `DatabaseError` in this test is thrown correctly but casting to an `NSError` in the `catch` clause fails. I guess this is either a bug in Swift on non-Darwin platforms or not supported. Why would we need `NSError` in this case instead of the Swift `Error` protocol? To preserve the error information in `DatabaseError`?

groue:
@marcprux, I found a pretty important blocker while validating the pull request for the official SQLCipher trait: https://github.com/groue/GRDB.swift/pull/1827#issuecomment-3418155348. This PR will suffer from the same problem.

Basically, the fact that Xcode lacks support for traits is likely to block all progress. Maybe this can be fixed, but this is beyond my skills.

orj:
⚠️ 🚨 @groue @marcprux how is this _at all_ safe to do?

I'm working on a project where we're encountering `dyld` binding issues with SQLite & GRDB due to a 3rd party vendor linking a version of SQLCipher into their dynamic framework on iOS.

This vendor has linked an old version of SQLite/SQLCipher into their framework.  This breaks GRDB because `dyld` seems to randomly select which versions of `_sqlite3_*` C calls to call from inside GRDB.

I see from the `.swiftinterface` files in the [DuckDuckGo fork of GRDB](https://github.com/duckduckgo/GRDB.swift) that GRDB inlines calls to `sqlite3_*` APIs in some areas.  This would mean that depending on how `dyld` binds SQLite to frameworks (including Apple frameworks) you might end up with random incompatibilities and duplicate global state.

The way this is manifesting in our app (if we link this vendors framework) is SQL syntax errors for syntax that is very old and very correct.

The only way I can think of this working correctly is if SQLite/SQLCipher is linked into into GRDB statically as a hidden dependency (which means no inline code or exposing of sqlite symbols `private import SQLite`). I believe it might also require that the SQLite/SQLCipher C library is linked/compiled into GRDB with `-fvisibility=hidden` [option to clang](https://clang.llvm.org/docs/LTOVisibility.html).

Though this may not be possible with Swift given this old unresolved bug: https://github.com/swiftlang/swift/issues/43633

mackoj:
> @marcprux, I found a pretty important blocker while validating the pull request for the official SQLCipher trait: [#1827 (comment)](https://github.com/groue/GRDB.swift/pull/1827#issuecomment-3418155348). This PR will suffer from the same problem.
> 
> Basically, the fact that Xcode lacks support for traits is likely to block all progress. Maybe this can be fixed, but this is beyond my skills.

Hello @groue, I haven’t read all the messages, so maybe someone has already suggested this, but there’s a pretty straightforward way to work around this limitation: you can import a local package into the project to enable the trait. 

This is the approach we use in our project using one local package as a proxy that depends on the package where we need the traits and enabling them there. This way, we have access to it. So, I don’t think the lack of trait support in Xcode should block this. People who need Linux/Android support can work around the Xcode limitation, and those who don’t need it won’t even be affected which I think applies to most users.

I’ve also shared an example project that demonstrates this approach. [testTrait.zip](https://github.com/user-attachments/files/23050625/testTrait.zip)

groue:
Thank you for your hint, @mackoj :-) The use case that is currently broken is the plain Xcode app that has a plain dependency on GRDB (via the Xcode UI), without any wrapper package. That's what most apps that use GRDB today do. The technique of the wrapper package is not a satisfying solution for those apps, because it would be a painful breaking change. Now, I hope that the use case that is currently broken can be restored while keeping the ability to use SQLCipher through SPM. That's the current discussion in #1827.

groue:
This pull request is obsolete, so I'm closing it. Thank you nevertheless @marcprux, your support is very precious! 🙏 

The Android situation should be in better shape since I merged the Linux compatibility changes of #1825 into the development branch. #1825 is still open, but it is now re-focused on the CI.

The SQLCipher situation is also in better shape: #1845 should be merged shortly.

From now on, future progress on the SQLCipher or Android topics should happen independently, in dedicated pull requests.

PR #1713, opened 2025-02-01
Expose `sqlite3_libversion_number` as `Database.sqliteLibVersionNumber`

groue:
This makes it easier to write tests that depend on SQLite version, because we do not need to import the SQLite C apis.


groue:
Oops, I merged to master by mistake. Fortunately this has no consequence.

Issue #1715, opened 2025-02-04
CancellationError during async write leaves database in readonly mode


rcarver:
When using the async version of `write` and a `CancellationError` is thrown, the database can be left in readonly mode.

Tested on 287c4be82d3a462a9bfda3822dfafc473cdb7513 and v7.0.0

```swift
try await database.write { ... }
```

As [documented here](https://github.com/groue/GRDB.swift/blob/287c4be82d3a462a9bfda3822dfafc473cdb7513/Documentation/GRDB7MigrationGuide.md#cancellable-async-database-accesses), ROLLBACK is specifically allowed within a cancelled task.

> In GRDB 7, asynchronous database accesses respect Task cancellation. If a Task is cancelled, reads and writes throw a CancellationError, pending transactions are rolled back, and the database is not modified. The only SQL statement that can execute in a cancelled database access is ROLLBACK.

I believe that returning the database to writable mode should also be allowed, for example this change: 

```diff
diff --git a/GRDB/Core/Database.swift b/GRDB/Core/Database.swift
index 3d7419ce5..df136e337 100644
--- a/GRDB/Core/Database.swift
+++ b/GRDB/Core/Database.swift
@@ -1287,7 +1287,11 @@ public final class Database: CustomStringConvertible, CustomDebugStringConvertib
         if statement.transactionEffect == .rollbackTransaction {
             return
         }
-
+        // Always allow the database to return to writable mode.
+        if statement.description == "PRAGMA query_only = 0" {
+            return
+        }
+
         // How should we interrupt the statement?
         enum Interrupt {
             case abort  // Rollback and throw SQLITE_ABORT

```

Here is a log showing where readonly mode is left on. 

```
# delete a record that is currently being observed
PRAGMA query_only = 0
BEGIN IMMEDIATE TRANSACTION
DELETE FROM "projects" WHERE "id" = ?
COMMIT TRANSACTION

# parent observation runs
PRAGMA query_only = 1
BEGIN DEFERRED TRANSACTION
SELECT "id" FROM "projects"
COMMIT TRANSACTION
PRAGMA query_only = 0

# self observation begins, but is cancelled because the containing object is destroyed
PRAGMA query_only = 1
BEGIN DEFERRED TRANSACTION
SELECT * FROM "projects" WHERE "id" = ?
COMMIT TRANSACTION
Caught error: CancellationError()

Attempt any write throws error:
BEGIN IMMEDIATE TRANSACTION

    DatabaseError(
[...]

PR #1716, opened 2025-02-04
Fix: DatabaseQueue restores its read/write abilities when an async read-only database access is cancelled.

groue:
This pull request fixes the bug reported in #1715. Thank you @rcarver!

### DIFF
--- a/GRDB/Core/Database.swift
+++ b/GRDB/Core/Database.swift
@@ -1288,6 +1288,12 @@ public final class Database: CustomStringConvertible, CustomDebugStringConvertib
             return
         }
         
+        // Suspension should not prevent adjusting the read-only mode.
+        // See <https://github.com/groue/GRDB.swift/issues/1715>.
+        if statement.isQueryOnlyPragma {
+            return
+        }
+        
         // How should we interrupt the statement?
         enum Interrupt {
             case abort  // Rollback and throw SQLITE_ABORT
--- a/GRDB/Core/Statement.swift
+++ b/GRDB/Core/Statement.swift
@@ -92,6 +92,9 @@ public final class Statement {
     /// The effects on the database (reported by `sqlite3_set_authorizer`).
     private(set) var authorizerEventKinds: [DatabaseEventKind] = []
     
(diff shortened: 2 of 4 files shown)


rcarver:
Confirming this works in the app, thank you @groue!

groue:
> Confirming this works in the app, thank you @groue!

Thanks for the confirmation 👍 

May I ask how you have noticed the bug in the first place, @rcarver? I'm very glad you did, but the bug is only present with DatabaseQueue, I was wondering why an app would use it instead of a DatabasePool. If you have a few minutes, would you please explain your setup and how the bug was triggered?

rcarver:
@groue of course! Lucky combination of factors I guess :)

I'm building a new app using the experimental GRDB in [Sharing](https://github.com/pointfreeco/swift-sharing/tree/main/Examples/GRDBDemo) which uses `ValueObservation` to keep state synced. The app is using [TCA](https://github.com/pointfreeco/swift-composable-architecture) which makes use of cancellation when dismissing child features. Finally, since the app is brand new I haven't yet explored DB Pool vs Queue, though it does seem like it will move to Pool when I get to it.

Attached is a simple app that recreates the problem, which I believe is:

* A child feature that observes a single record is added to a list
* The child starts deletes itself
* The parent observes the change and updates state
* By removing the child, its in-flight effects are cancelled, which throws the CancellationError
* The child deletion receives CancellationError and leaves the db in readonly mode

[GRDB Cancellation.zip](https://github.com/user-attachments/files/18663191/GRDB.Cancellation.zip)

groue:
Oh, thank you. I'll just remind that `DatabaseQueue` serializes all database accesses, when `DatabasePool` can run parallel reads and writes. With a pool, the read-only parts of an app can run without blocking or any delay, even if some other component of the app performs a (slow) background write. I don't have the slightest idea whether Sharing and TCA can support such parallelization.

Anyway, yes we've been lucky :-)

Happy GRDB!

groue:
Since I discovered an SQLite bug on the way: I reported it here: [Bug report: PRAGMA query_only has side effect when compiled, not when executed](https://sqlite.org/forum/forumpost/d1eebbd331).

PR #1759, opened 2025-04-12
Build request with closures

groue:
This pull request brings a new way to write query interface requests:

```swift
struct Player: Codable, Identifiable {
    var id: String
    var name: String
    var score: Int
    var isRegistered: Bool
}

extension Player: FetchableRecord, PersistableRecord {
    enum Columns {
        static let id = Column(CodingKeys.id)
        static let name = Column(CodingKeys.name)
        static let score = Column(CodingKeys.score)
        static let isRegistered = Column(CodingKeys.isRegistered)
    }
}

try dbQueue.read { db in
    // Filter, order, select, group, having now accept a closure.
    // In Swift 6.1, the closure can be written as a key path.
    let registeredPlayers = try Player
        .filter(\.isRegistered)     // NEW (Swift 6.1)
        .filter { $0.isRegistered } // NEW (Swift 6.0)
        .fetchAll(db)
    
    let highEndPlayers = try Player
        .filter { $0.score > 1000 && $0.isRegistered } // NEW
        .fetchAll(db)
    
    let bestPlayers = try Player
        .order(\.score.desc)     // NEW (Swift 6.1+)
        .order { $0.score.desc } // NEW (Swift 6.0)
        .limit(10)
        .fetchAll(db)

     let playersByScore = try Player
        .order { [ // NEW
            $0.score,
            $0.name.collating(.localizedStandardCompare),
        ] }
        .fetchAll(db)
    
    let names = try Player
        .select(\.name, as: String.self)      // NEW (Swift 6.1+)
        .select({ $0.name }, as: String.self) // NEW (Swift 6.0)
        .fetchSet(db)

    // Batch updates
    try Player.updateAll(db) { [$0.score += 10] }
    
    // Association aggregates can be defined with a closure.
    // In Swift 6.1, the closure can be written as a key path.
    let teams = try Team
        .annotated(with: Team.players.max(\.score))     // NEW (Swift 6.1+)
        .annotated(with: Team.players.max { $0.score }) // NEW (Swift 6.0)
}
```

TODO:

- [x] `enum Columns: String, ColumnExpression { ... }`: those do not support the key path syntax, because [SE-0438](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0438-metatype-keypath.md) did not ship with support for enums. We now discourage those enums, because we don't know when the language will improve.
- [x] Update documentation
- [x] Update DocC
- [x] Improve table aliases, so that `alias[Foo.Columns.name]` can be replaced with `alias.name`


freak4pc:
Love the syntax 👏

PR #1810, opened 2025-09-08
Persistable Database Views

groue:
This pull request enhances support for record types that target a database view. It is now possible to instruct GRDB about the primary key of a view, and this unlocks both reads and writes that used to throw a database error:

```swift
// Now possible, when the Captain record targets the "captain" database view:
var bob = try Captain.find(db, id: "bob")
try captain.updateChanges(db) {
  $0.name = "Bobby"
  $0.teamId = "red"
}
```


Tests/GRDBTests/PrimaryKeyInfoTests.swift
```diff
@@ -242,4 +250,121 @@ class PrimaryKeyInfoTests: GRDBTestCase {
...
+            }
+            
+            do {
+                _ = try db.primaryKey("playerView")
+                XCTFail("Expected Error")
+            } catch let error as DatabaseError {
+                XCTAssertEqual(error.resultCode, .SQLITE_ERROR)
+                XCTAssertEqual(error.message, """
+                    The database view 'playerView' has no primary key, \
+                    according to Configuration.schemaSource.
```
groue, line 286: Fix this error message, which is no longer accurate since we introduced `withSchemaSource`. Check other error messages.

stevensJourney:
Thanks for the awesome work @groue! I've tested this for PowerSync on our end with some basic queries. We're still ironing out some unrelated kinks in our PowerSync integration, but so far this works perfectly for us.

groue:
Thank you for the good news, @stevensJourney! 👍 Unless you tell me otherwise, I'll merge this PR shortly!

groue:
⛵ Shipped in v7.7.0!

PR #1818, opened 2025-09-26
Merged Migrations

groue:
The new `registerMigration(_:foreignKeyChecks:merging:migrate)` method registers a migration that **merges and replaces** a set of migrations defined in a previous version of the application. For example, to merge the migrations "v1", "v2" and "v3", redefine the "v3" migration so that it merges "v1" and "v2", as in the example below.

The second argument of the migration closure is the set of merged migrations that has been applied when the merged migration runs.

```swift
// Old code
migrator.registerMigration("v1") { db in
    // Apply schema version 1
}
migrator.registerMigration("v2") { db in
    // Apply schema version 2
}
migrator.registerMigration("v3") { db in
    // Apply schema version 3
}

// New code:
// - Migrations v1 and v2 are deleted.
// - Migration v3 is redefined and merges v1 and v2:
migrator.registerMigration("v3", merging: ["v1", "v2"]) { db, appliedIDs in
    if !appliedIDs.contains("v1") {
        // Apply schema version 1
    }
    if !appliedIDs.contains("v2") {
        // Apply schema version 2
    }
    // Apply schema version 3
}
```

See https://github.com/groue/GRDB.swift/discussions/1817 for a longer discussion.

### DIFF
--- a/GRDB/Migration/DatabaseMigrator.swift
+++ b/GRDB/Migration/DatabaseMigrator.swift
@@ -18,6 +18,7 @@ import Foundation
 /// ### Registering Migrations
 ///
 /// - ``registerMigration(_:foreignKeyChecks:migrate:)``
+/// - ``registerMigration(_:foreignKeyChecks:merging:migrate:)``
 /// - ``ForeignKeyChecks``
 ///
 /// ### Configuring a DatabaseMigrator
@@ -161,9 +162,8 @@ public struct DatabaseMigrator: Sendable {
     
     /// Registers a migration.
     ///
-    /// The registered migration is appended to the list of migrations to run:
-    /// it will execute after previously registered migrations, and before
-    /// migrations that are registered later.
+    /// The registered migration is appended to the list of migrations. It
+    /// will execute after previously registered migrations.
     ///
     /// For example:
     ///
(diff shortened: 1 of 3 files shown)


groue:
@bok- I'm about to merge this pull request, because I think it covers all use cases I could foresee. Apps will have all the information they need in order to implement the logic relevant to their use case.

[Your own use case](https://github.com/groue/GRDB.swift/discussions/1817) is slightly specific because:

> anyone who hasn't updated the app in more than 12 months would get a slightly worse experience when we delete and re-create the database

Maybe you'll write something like:

```swift
struct NeedsEraseDatabaseError: Error { }

migrator.registerMigration("v3", merging: ["v1", "v2"]) { db, appliedIDs in
    if appliedIDs.isEmpty == false {
        // The merged migrations were v1 and/or v2 are already applied.
        // We do not support this scenario because we can only
        // apply the v3 schema in its entirety: so let's just reset
        // the database.
        throw NeedsEraseDatabaseError()
    }
    // Apply full schema v3
}

...

do {
    try migrator.migrate(writer)
} catch is NeedsEraseDatabaseError {
    try writer.erase()
    try migrator.migrate(writer)
}
```

groue:
⛵ Shipped in v7.8.0

Issue #1816, opened 2025-09-22
Proposal for Official SQLCipher.swift Pull Request


sjlombardo:
On behalf of the SQLCipher team, I wanted to open a preliminary discussion about a planned pull request that would add support for the official [SQLCipher.swift](https://github.com/sqlcipher/SQLCipher.swift) package. With the deprecation of CocoaPods looming, the SQLCipher Pod subspec will no longer be an option for officially supported integration.

Since we recently released the [SQLCipher.swift](https://github.com/sqlcipher/SQLCipher.swift) Swift Package, we would like to submit a PR to support it. The PR would use a trait that developers could enable to bring in the dependency on the SQLCipher.swift package (i.e., it would not affect the default behavior in any way).

That said, we also noticed the ongoing discussion in PR #1708 related to the third party swift-sqlcipher integration. To be upfront, SQLCipher.swift is currently Apple ecosystem only and does use binaryTargets. However, it is the only package currently maintained by the SQLCipher team and provides support for the core library on Apple platforms.

We believe both approaches could technically coexist via separate traits. Our only preference would be for traits specifically called "SQLCipher" to be reserved for official packages to avoid naming or trademark confusion. In other words, we don't intend to discourage PR #1708 in any way. We simply want to ensure users have clear and well supported paths from both the SQLCipher project and third party contributors.

Please let me know what you think about this approach, or if you have any specific concerns you'd like us to address in a PR.

groue:
Hello, @sjlombardo. I'm glad to see you here :-)

Please let me invite @marcprux to this conversation, because he was quite involved in previous conversations in our repositories.

> We believe both approaches could technically coexist via separate traits. Our only preference would be for traits specifically called "SQLCipher" to be reserved for official packages to avoid naming or trademark confusion. In other words, we don't intend to discourage PR https://github.com/groue/GRDB.swift/pull/1708 in any way. We simply want to ensure users have clear and well supported paths from both the SQLCipher project and third party contributors.

Thank you for bringing this topic so clearly. This is very sensible, and I welcome those considerations. I do not want to confuse our users, so be assured the plain "SQLCipher" name will be reserved for sqlcipher/SQLCipher.swift.

> Please let me know [...] if you have any specific concerns you'd like us to address in a PR.

[...]

PR #1819, opened 2025-09-28
Accept multiple SQLCipher libraries

groue:
This pull request aims at easing the job of #1708 and #1816, by checking the `SQLITE_HAS_CODEC` compiler flag for enabling SQLCipher features.

The `GRDBCIPHER` compiler flag is now only used for detecting the SQLCipher CocoaPods:

```swift
#if GRDBCIPHER // CocoaPods (SQLCipher subspec)
import SQLCipher
#else ...
#endif

#if SQLITE_HAS_CODEC
// GRDB is linked against some flavor of SQLCipher:
// sqlite3_key, PRAGMA cipher_version, etc. are available
#endif
```

@marcprux, @R4N, I'll merge if you both agree that this pull request is actually useful.

### DIFF
--- a/Tests/GRDBTests/FTS3TableBuilderTests.swift
+++ b/Tests/GRDBTests/FTS3TableBuilderTests.swift
@@ -88,6 +88,9 @@ class FTS3TableBuilderTests: GRDBTestCase {
 
     #if GRDBCUSTOMSQLITE || SQLITE_HAS_CODEC
     func testUnicode61TokenizerDiacriticsRemove() throws {
+        guard Database.sqliteLibVersionNumber >= 3027000 else {
+            throw XCTSkip("remove_diacritics=2 is not available")
+        }
         let dbQueue = try makeDatabaseQueue()
         try dbQueue.inDatabase { db in
             try db.create(virtualTable: "documents", using: FTS3()) { t in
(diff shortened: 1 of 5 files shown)


groue:
Also, @marcprux, I believe this new interpretation of compiler flags removes the need to rework the order of imports, as in https://github.com/groue/GRDB.swift/pull/1708/commits/b26c42c83a8bb94f1a236004704c0c8f8a025353. Am I correct?

R4N:
@groue This looks good to us as well. I believe we'll still need to adjust the order so that the `SQLCIPHER`/`SKIP_SQLCIPHER` import checks are prior to `SWIFT_PACKAGE` import check in our PR's since SWIFT_PACKAGE will be true for both:

```
// Import C SQLite functions
#if SQLCIPHER
import SQLCipher
// #elseif SKIP_SQLCIPHER
// import for SKIP_SQLCIPHER
#elseif SWIFT_PACKAGE
import GRDBSQLite
#elseif GRDBCIPHER
import SQLCipher
#elseif !GRDBCUSTOMSQLITE && !SQLITE_HAS_CODEC
import SQLite3
#endif
```

we'll sync up our fork once this is merged in and base our PR off of it.

groue:
The import section was refactored in #1826. This pull request has been rebased on top of it, and does no longer modify the imports.

Issue #1820, opened 2025-09-28
update to GRDB 7.7.0 causes test failures when using the `saved()` method on `MutablePersistableRecord`


yqiang:
After upgrading to GRDB 7.7.0 from GRDB 7.6.1 I noticed that several of my unit tests started to fail. Specifically, it seems like something about the `.saved()` method seems to have changed. I narrowed down the failure to a code snippet like this, where we start with an empty, in-memory database:

```
try await appDatabase.writer.write { db in
  _ = try food1.food.saved(db)
  _ = try food2.food.saved(db)
  _ = try serving1.saved(db)
  _ = try serving2.saved(db)
  _ = try section.saved(db)
}
```

In this instance `serving1` has a foreign key dependency on on `food` and I get the following error: 

```
Test "Test frequently paired foods" recorded an issue at FrequentlyPairedFoodsTests.swift:14:4: 
Caught error: SQLite error 19: FOREIGN KEY constraint failed - while executing `INSERT INTO "serving" ("id", "description", "nutritionMultiplier", "amount", "measure", "metricUnit", "metricValue", "sortOrder", "isPublic", "derivedFrom", "nutritionPanelId", "foodId") VALUES (?,?,?,?,?,?,?,?,?,?,?,?)` with arguments ["eb46ddbf-e15b-4be2-8daa-8ab83c657953", "1 piece", 1.0, NULL, NULL, NULL, NULL, 0, 0, NULL, NULL, "food2"]
```

If I replace the `saved` calls here with `inserted`, the tests pass again. Hopefully this is enough info to go on, but if not, I'm happy to try to create a sample project. 

Environment
- Xcode 26.0 
- iOS 26.0

groue:
Thanks for the report @yqiang, I'll check quickly.

groue:
The error says "SQLite error 19: FOREIGN KEY constraint failed". The application is trying to insert a `serving` which references a non-existing relationship (probably the `foodId` with id "food2").

If you think the error is related to the GRDB upgrade, please provide a reproducing sample code.

yqiang:
Here's a sample repro. It was a bit tricky to track down but it seems to have something to do with having an FTS table associated one of the tables. 

https://github.com/yqiang/GRDBBug

Here's the repro and the log output in case it's more convenient than cloning a repo:

```
import SwiftUI
import GRDB

struct Serving: Codable, FetchableRecord, PersistableRecord {
  let id: UUID
  var description: String
  var foodId: String
[...]

PR #1822, opened 2025-09-29
Fix the save() method for record types with attached triggers

groue:
This pull request fixes a regression introduced by #1810 "Persistable Database Views".

The `save()` method performs an update, and then an insert if the update has not updated any database row. #1810 broke the check for the existence of updated rows in the case of database tables with attached triggers (FTS5 tables are an example). In this pull request we restore the correct behavior for database tables.

Thank you @yqiang for reporting the bug in #1820 🙏


groue:
:boat: Shipped in v7.7.1

Issue #1838, opened 2025-12-05
App crashes when canceling a task that contains an asynchronous GRDB write operation


sepw:
I start a task that writes to a database queue asynchronously using:

```swift
let task = Task {
  try await queue.write {
    ...
    try Task.checkCancellation()
    ...
  }
}
```

I'm attempting to cancel the task using

```swift
task.cancel()
```

and the app crashes with the following error:

```
GRDB/SerializedDatabase.swift:261: Fatal error: A transaction has been left opened at the end of a database access
```

I only get this fatal error from GRDB when the task is cancelled. The write transaction is rolled back correctly when I throw an application-defined error from the closure passed to the database writer's `write(_:)` function:

```swift
let task = Task {
  try await queue.write {
    ...
    throw AppError.oops
    ...
  }
}
```

According to the GRDB documentation for `DatabaseWriter`'s `func write(_:) async throws`, cancellation of a task performing a write is legal and supported:

```swift
extension DatabaseWriter {

    /// ...
    /// - throws: Any ``DatabaseError`` that happens while establishing the
    ///   database access, or the error thrown by `updates`, or
    ///   `CancellationError` if the task is cancelled.
    public func write<T: Sendable>(
        _ updates: @Sendable (Database) throws -> T
    ) async throws -> T
}
```

Is there something I should be doing to handle cancellation errors in asynchronous database read or write closures?

GRDB version: 7.8.0.

groue:
Hello @sepw,

Thanks for the report, it definitely looks like a bug. Do you happen to have a reproducing scenario? This would definitely help.

sepw:
Hi @groue 

Thanks for the quick response.

> Thanks for the report, it definitely looks like a bug. Do you happen to have a reproducing scenario? This would definitely help.

I have a fair amount of infrastructure built on top of GRDB, so I can't easily provide a simple reproducing scenario. I will try and strip my code back to produce a simpler example that triggers the fatal error. In the meantime:

I am experiencing the fatal error in a unit test that 

1) creates an in-memory database queue;
2) creates the database schema;
3) starts a task that reads and writes to/from the DB;
4) cancels the task after a second and waits for the canceled task to complete; and
5) tears down in-memory database queue.

The call stack of the fatal error is:

```
#4	GRDBPrecondition(_:_:file:line:) at GRDB.swift/GRDB/Utils/Utils.swift:51
[...]

PR #1839, opened 2025-12-08
Fix cancellation of async tasks that use the FTS5 full-text engine.

groue:
The issue #1838 has revealed several bugs that are fixed in this pull request:

- A GRDB bug where a database transaction might not properly rollback in case of Task cancellation.
- A GRDB bug where a read-only access might not properly leave the read-only mode in case of Task cancellation.
- An [SQLite bug](https://sqlite.org/forum/forumpost/95413eb410) where some prepared statements leaked by the FTS5 full-text engine prevent a cancelled database access from completing correctly.


groue:
⛵ @sepw, the fix has shipped in v7.9.0.

PR #1849, opened 2026-02-03
Add Sendable conformance to DispatchQueueExecutor for Android and Windows

marcprux:
As with Linux, Android and Windows need DispatchQueueExecutor to add a retroactive conformance to Sendable.


groue:
I'm happy that we can finally make progress on the Android side 👏 

Thank you very much @marcprux!

Issue #1863, opened 2026-05-04
Allow installing update hooks for statements that don't seem to write?


simolus3:
GRDB only installs update hooks on the underlying SQLite connection when running a statement that appears to affect a table an observer is interested in (based on authorizer callbacks). I assume it does that for performance to avoid unecessary callbacks, but we have a use case where looking at authorizer results alone misses writes.

In our case (for PowerSync), we have a SQLite extension registering user-defined functions and virtual tables that will internally run write statements. So when we try to run a statement like `SELECT powersync_clear()` (which internally deletes from tables) via high-level GRDB APIs, the authorizer determines that the statement can't possibly write and GRDB thus doesn't install an update hook. The same thing can happen for virtual tables, where a read on that table might be implemented internally by a statement writing to the database.

Currently, our solution to this issue is to obtain the raw SQLite pointer from GRDB and to install our own update/pre-update hook and then use direct SQLite APIs to prepare statements. We would definitely prefer being able to use high-level GRDB statement APIs, but these missed updates prevent us from doing that.

Would it be possible to perhaps mark a statement calling those functions / vtabs as writing in GRDB? E.g. if we had a way to declare that a `Statement` may write to any table and GRDB would install update hooks for that statement, this problem would be solved. Of course, it's kind of hacky that users have to declare this on their statements, but it might be a useful escape hatch? We would be happy to contribute something like this if it's acceptable for the project, but at least we wanted to start a discussion on this.

groue:
Hello @simolus3,

Thank you for opening this interesting issue.

Did you see the [`notifyChanges(in:)`](https://swiftpackageindex.com/groue/grdb.swift/documentation/grdb/database/notifychanges(in:)) method? Maybe it could provide a ready-made solution:

```swift
try dbQueue.write { db in
    // Changes are not detected here...
    try db.execute(sql: "SELECT powersync_clear()")
    // ... but you can help, by providing a description of the changed region:
    try db.notifyChanges(in: .fullDatabase)
}
```

simolus3:
Yes, we currently use `notifyChanges(in:)` in combination with custom update hooks to forward our low-level writes to GRDB.

[...]

PR #1864, opened 2026-05-05
Add option to disable database change filtering in TransactionObserver

simolus3:
Some statements might issue writes to the databases through a layer of indirection (e.g. via a user-defined function). For these statements, GRDB is unable to detect the writes made and doesn't call `TransactionObserver.observes(eventKind:)`, doesn't install an update listener and has no way of observing these writes.

This adds `TransactionObserver.observesAllDatabaseChanges`, which can be used to make transaction observers listen for all database changes. It returns `false` by default, but can be added to implementations to observe indirect statements as well.

Closes https://github.com/groue/GRDB.swift/issues/1863.

### Pull Request Checklist

### DIFF
--- a/GRDB/Core/TransactionObserver.swift
+++ b/GRDB/Core/TransactionObserver.swift
@@ -774,13 +774,14 @@ public protocol TransactionObserver: AnyObject {
     /// from being applied on the observed tables.
     func observes(eventsOfKind eventKind: DatabaseEventKind) -> Bool
     
-    /// Returns whether statements made by the statement should be notified to the observer.
+    /// Returns whether this observer should unconditionally be notified for all database changes, regardless of
+    /// ``observes(eventsOfKind:)``.
     ///
     /// This method can be used to observe indirect writes from statements (e.g. a statement invoking
     /// a custom SQL function which internally runs its own statements). ``observes(eventsOfKind:)``
     /// would not be called for such statements because potential writes can't be inferred from the syntax of
     /// these statements.
-    func observes(statement: Statement) -> Bool
+    var observesAllDatabaseChanges: Bool { get }
     
     /// Called when the database was modified in some unspecified way.
     ///
@@ -860,8 +861,8 @@ public protocol TransactionObserver: AnyObject {
 
 extension TransactionObserver {
     /// The default implementation does not observe statements.
-    public func observes(statement: Statement) -> Bool {
-        return false
+    public var observesAllDatabaseChanges: Bool {
+        false
     }
 
     /// The default implementation does nothing.
@@ -958,7 +959,11 @@ final class TransactionObservation {
     }
     
     func observes(eventsOfKind eventKind: DatabaseEventKind) -> Bool {
-        observer?.observes(eventsOfKind: eventKind) ?? false
+        guard let observer else {
+            return false
+        }
+        return observer.observesAllDatabaseChanges || observer
+            .observes(eventsOfKind: eventKind)
     }
     
     func observations(for statement: Statement) -> StatementObservation? {
@@ -966,7 +971,7 @@ final class TransactionObservation {
             return nil
         }
         
-        if observer.observes(statement: statement) {
+        if observer.observesAllDatabaseChanges {
             return StatementObservation(
                 transactionObservation: self,
                 trackingEvents: .all)
--- a/GRDB/Core/TransactionObserver.swift
+++ b/GRDB/Core/TransactionObserver.swift
@@ -334,65 +334,14 @@ class DatabaseObservationBroker {
             SchedulingWatchdog.current!.databaseObservationBroker = self
             
             // Fill statementObservations with observations that are interested
-            // in the kind of database events performed by the statement, as
-            // reported by `sqlite3_set_authorizer`.
-            //
+            // in the kind of database events performed by the statement.
             // Those statementObservations will be notified of individual changes
             // in databaseWillChange() and databaseDidChange().
-            let authorizerEventKinds = statement.authorizerEventKinds
-            
-            switch authorizerEventKinds.count {
-            case 0:
-                // Statement has no effect on any database table.
-                //
-                // For example: PRAGMA foreign_keys = ON
-                statementObservations = []
-            case 1:
-                // We'll execute a simple statement without any side effect.
-                // Eventual database events will thus all have the same kind. All
-                // detabase events can be notified to interested observations.
-                //
-                // For example, if one observes all deletions in the table T, then
-                // all individual deletions of DELETE FROM T are notified:
-                let eventKind = authorizerEventKinds[0]
-                statementObservations = transactionObservations.compactMap { observation in
-                    guard observation.observes(eventsOfKind: eventKind) else {
-                        // observation is not interested
-                        return nil
-                    }
-                    
-                    // Observation will be notified of all individual events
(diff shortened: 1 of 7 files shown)


groue:
Thank you @simolus3!

I'll review this pull request in the next few days. At first glance, it does exactly what we talked about, and it does it well 👍 

Please tell me if it allows PowerSync to achieve the goals you described in #1863, and in particular if later on you can call `notifyChanges(in:)` with the desired effects 🙏

simolus3:
> Please tell me if it allows PowerSync to achieve the goals you described in #1863, and in particular if later on you can call `notifyChanges(in:)` with the desired effects 🙏

Yes, I've tested these changes with PowerSync and they seem to do exactly what we need :) My approach right now is to add a catchall observer in a `writeWithoutTransaction` block running statements for PowerSync, once that is complete I forward collected writes with `notifyChanges(in:)`  in an empty transaction. This updates GRDB value observations reliably 👍

groue:
I had a closer look, and I could spot a design flaw 🙂 

Your changes closely reflect our initial intent, but I gave you a misleading direction. It won't be difficult to fix, fortunately!

---

The symptom of the design flaw is that the new test fails if the observer returns false from `observes(eventsOfKind:)`:

```diff
 private class EverythingObserver: TransactionObserver {
     func observes(statement: Statement) -> Bool {
         return true
     }
 
     func observes(eventsOfKind eventKind: GRDB.DatabaseEventKind) -> Bool {
-        return true  // test passes
+        return false // test fails
     }
 }
```

This is definitely a bug. With this modification of the test, the `observes(statement:)` method still returns true for both `DELETE FROM foo` and `SELECT custom_clear_function()`, so we ought to be notified of deletions.

The cause is that SQLite compiles the `DELETE FROM foo` statement with the [truncate optimization](https://sqlite.org/lang_delete.html#the_truncate_optimization) enabled. When run, the statement truncates the table instead of deleting rows one by one. Individual deletions are not notified to the update hook, and the transaction observer is blind.

And there lies the design flaw: the truncate optimization is enabled or disabled from the compile-time authorization callbacks, before we can call `observes(statement:)`, which means before the observer can instruct GRDB to disable the truncate optimization.

---

The intention described at https://github.com/groue/GRDB.swift/issues/1863#issuecomment-4372697394 still holds:

> [...] because we can't build an argument for `observes(eventsOfKind:)` before executing `SELECT powersync_clear()`, it is necessary to come up with a new `TransactionObserver` method that opts in for the events created by this statement.

It turns wrong in the next paragraph:

> For example: `func observes(statement: DatabaseStatement) -> Bool` that returns false by default (preserving the API contract with other transaction observers that do not implement this method).

I think our only solution is to let observers opt in for all update events, without any filtering:

```diff
 protocol TransactionObserver {
-    func observes(statement: Statement) -> Bool
+    var observesAllDatabaseChanges: Bool { get }
 }
```

---

If you lack time, I'll update this pull request accordingly. The [`TransactionObserver `](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/transactionobserver) high-level documentation needs an update, too. Probably we should start mentioning the "update hook" aka the [Data Change Notification Callbacks](https://sqlite.org/c3ref/update_hook.html) very explicitly, without beating around the bush. We also have various paragraphs titled "Dealing with undetected changes" ([1](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/transactionobserver#Dealing-with-Undetected-Changes), [2](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/valueobservation#Dealing-with-Undetected-Changes), [3](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/databaseregionobservation#Dealing-with-Undetected-Changes)) that need to include "statements ran from a custom SQL function" in the possible causes for undetected changes.

simolus3:
> I think our only solution is to let observers opt in for all update events, without any filtering:

That will definitely work for PowerSync too, I've changed the PR to be based on that 👍 

>  The [`TransactionObserver `](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/transactionobserver) high-level documentation needs an update, too

I've added a small note on `observesAllDatabaseChanges`.

> We also have various paragraphs titled "Dealing with undetected changes" ([1](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/transactionobserver#Dealing-with-Undetected-Changes), [2](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/valueobservation#Dealing-with-Undetected-Changes), [3](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/databaseregionobservation#Dealing-with-Undetected-Changes)) that need to include "statements ran from a custom SQL function" in the possible causes for undetected changes.

I suppose "Changes performed by SQLite statements that are not both compiled and executed through GRDB APIs" implicitly covers that (since the function executes the statement with direct SQLite APIs internally, not with GRDB APIs). But it's good to be explicit about this, and I've expanded that bullet point.

> Probably we should start mentioning the "update hook" aka the [Data Change Notification Callbacks](https://sqlite.org/c3ref/update_hook.html) very explicitly, without beating around the bush. 

I'm not really sure where we'd mention that, in the section about transaction observers controlling for which events they'd be invoked?

I would be happy to expand the documentation with a bit more help from your side, but if you think it's quicker if you update this PR, then that's definitely fine with me as well :)

groue:
Sounds great! OK I take charge of the remaining tidbits 👍 Thank you very much for your help 🙏

groue:
Hello @simolus3, sorry for the delay. May is a lazy month in France, and a heat wave did not help 😅 

I performed a last refactor, due to the misleading `observesAllDatabaseChanges` name. Practically speaking, transaction observers are not notified of changes that are performed in a save point that is eventually rollbacked. GRDB only notifies changes that have an actual chance of being committed to disk (see [Database Changes And Transactions](https://swiftpackageindex.com/groue/grdb.swift/v7.10.0/documentation/grdb/transactionobserver#Database-Changes-And-Transactions) for more details).

By exposing a `DatabaseEventObservationStrategy` type, we allow for a future extra flag that ignores save points, trading observation precision for a reduced memory footprint.

The universal observer you need is now written as below:

```swift
// An observer that observes all database changes.
class UniversalObserver: TransactionObserver {
    var databaseEventObservationStrategy: DatabaseEventObservationStrategy {
        var strategy = DatabaseEventObservationStrategy.default
        // Don't filter on database event kind, so that we are
        // notified of changes performed through the SQLite C API:
        strategy.requiresDatabaseEventKind = false
        return strategy
    }

    // You still have to provide an implementation for this method,
    // but it will never be called since `requiresDatabaseEventKind`
    // is false.
    func observes(eventsOfKind eventKind: DatabaseEventKind) -> Bool {
        false // ignored
    }

    func databaseDidChange(with event: DatabaseEvent) {
        // Handle the change
    }
}
```

simolus3:
Thanks for all your help with this :) I agree that `DatabaseEventObservationStrategy` is a nice way to make this extensible.

groue:
Thank you for your patience. Full test suite is still running, I'll ship shortly after, probably tomorrow.

groue:
⛵ The feature was shipped in v7.11.0 :-)

PR #1853, opened 2026-02-16
Adds SQLCipher Custom Fork documentation

R4N:
Adds SQLCipherCustomFork.md documentation. Adjusts README.md Encryption section to reference officially supported GRDB + SQLCipher fork and link SQLCipherCustomFork.md documentation.

This is a documentation update which includes instructions for creating a custom fork of GRDB w/SQLCipher to support requirements outside of the scope of the official fork.

Official Fork link has a placeholder of `<url to official fork coming soon>`

### Pull Request Checklist

- [x] CONTRIBUTING: You have read https://github.com/groue/GRDB.swift/blob/master/CONTRIBUTING.md
- [x] BRANCH: This pull request is submitted against the `development` branch.
- [x] DOCUMENTATION: Inline documentation has been updated.
- [x] DOCUMENTATION: README.md or another dedicated guide has been updated.
- [x] TESTS: Changes are tested.
- [x] TESTS: The `make smokeTest` terminal command runs without failure.

PR #1882, opened 2026-09-23
Create SQLCipher trait

GeorgeElsham:
Create two traits:
- `SQLite`: for unencrypted databases.
- `SQLCipher`: for encrypted databases.

### Pull Request Checklist


GeorgeElsham:
Hi @groue, I've made use of Swift Package Manager traits so that we no longer need to fork (or use a fork) of GRDB.swift in order to use SQLCipher.swift. Let me know what you think!

groue:
Hello @GeorgeElsham,

Thanks for your contribution, I really appreciate it. However I cannot merge this work, due to several limitations of Xcode and SPM.

For more information, see the "GRDB SQLCipher saga" #1827, and in particular my current conclusion at https://github.com/groue/GRDB.swift/pull/1827#issuecomment-3450562202.

SQLCipher maintain their own ready-made GRDB fork, and I have to process the #1853 pull request in order to make it easier for people to find it.

GeorgeElsham:
@groue Thank you for the reply. After reading your linked conclusion comment, as far as I understand, your main concern is Xcode support?

However, it looks like Xcode 26 is now the minimum version required for ASC. Even so, even without traits UI in Xcode the non-SQLCipher part works just fine, but users using old unsupported Xcode versions would however have to pin the package to a version they can fork if they choose to use SQLCipher.

I have also added traits so that SQLCipher.swift is not downloaded unnecessarily.

What are you latest reasons for still not supporting using SQLCipher via traits, nearly a year on from last time?

